PR #7, opened 2017-07-13
Add samplerExpr to samplers in the descriptor map

dneto0:
The value is a quoted string containing the sampler literal
initializer expression (with whitespace removed).  The string
expression lists the components in sorted order, for easier
machine processing.

Adding this explicit expression makes it easier to use the descriptor
map since you no longer have to decode the integer value for the sampler.
Instead, it closely matches the expression in the specified samplermap
input file.


dneto0:
Passed private review.

PR #16, opened 2017-07-14
Error for unsupported vector types

dneto0:
Two commits in this PR:

1
    Emit user error for some uses of long vectors
    
    Error out when there is a function definition which has an unsupported
    vector type for return value or any of its parameters.

2
    Error out for unsupported vector types in function bodies
    
    Now that we've added the length 8 and 16 vector types again,
    we should emit errors when they are used inside function bodies.

### DIFF
--- a/cmake/strip_banned_opencl_features.cmakescript
+++ b/cmake/strip_banned_opencl_features.cmakescript
@@ -51,4 +51,27 @@ string(REGEX REPLACE "\n[^;\n]*_sat[^;]*;" "" lines "${lines}")
 # get_work_dim gets stripped out, lets fudge it back in
 string(REPLACE "#define __cnfn __attribute__((const))" "#define __cnfn __attribute__((const))\nunsigned int __ovld __cnfn get_work_dim(void);" lines "${lines}")
 
+# Declare the invalid vector types so we can issue better error messages
+set(BANNED_VECTOR_TYPES "")
+foreach(LEN "8" "16")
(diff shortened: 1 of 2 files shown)


dneto0:
Please take a look @sheredom 
Will this allow too many things to go through?

dneto0:
Here's the kind of output we get:
```
$ cat a.cl
short8 x() { return (short8)(1);}
dneto@dneto0:/build/dneto/aglet/SPIRV-LLVM/build$ clspv a.cl
a.cl:1:1: error: vectors with more than 4 elements are not supported
short8 x() { return (short8)(1);}
^

$ cat aa.cl
float x() {
  short8 y;
  return 1.0;
}
$ clspv aa.cl
aa.cl:2:3: error: vectors with more than 4 elements are not supported
  short8 y;
  ^

$ cat aac.cl
float x() {
  (short8)(1);
  return 1.0;
}

$ clspv aac.cl   ### Note there is no error here.


$ cat aab.cl
float x() {
  typedef short8 foo;
  return 1.0;
}
$ clspv aa.cl
aa.cl:2:3: error: vectors with more than 4 elements are not supported
  short8 y;
  ^
```

sheredom:
Yup that is the error!

dneto0:
Rebased

dneto0:
Rebased.

Also, I checked the AST for the aac.cl test case.  The `(short8)(1);` statement does not even appear. ?!  So there's no opportunity to error out for it.

dneto0:
Passes on Linux in release and debug builds.

PR #17, opened 2017-07-14
Make utils/fetch_sources.py compatible with Python 3

alycm:
In addition to the current Python 2 compatibility.


alycm:
I signed it.

Issue #5, opened 2017-07-13
compiler adding StorageImageReadWithoutFormat when not using OpImageRead


dneto0:
Might be similar issue with StorageImageWriteWithoutFormat

Vulkan says:

> The Image Format of an OpTypeImage declaration must not be Unknown, for variables which are used for OpImageRead or OpImageWrite operations, except under the following conditions:

...

> For OpImageRead, if the shaderStorageImageReadWithoutFormat feature is enabled and the shader module declares the StorageImageReadWithoutFormat capability.

But clspv will add the capability even if only doing OpImageSampleExplicitLod. E.g. see tests/ImageBuiltins/read_imagef_sampler_float2.cl

It seems to be adding the capability when just mentioning the read image types.  

The simplifying assumption is also reflected in the documentation.

dneto0:
For read operations, the compiler only emits OpImageSampleExplicitLod, so it should not generate StorageImageReadWithoutFormat at all.

For write operations, the compiler emits OpImageWrite

PR #34, opened 2017-08-21
Don't emit StorageImageReadWithoutFormat

dneto0:
The only image read operation performed is OpImageSampleExplicitLod,
so there is no need to declare the StorageImageReadWithoutFormat
capability.

Fixes https://github.com/google/clspv/issues/5

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2710,18 +2710,8 @@ void SPIRVProducerPass::GenerateModuleInfo() {
                                                 spv::CapabilityFloat64)));
     } else if (auto *STy = dyn_cast<StructType>(Ty)) {
       if (STy->isOpaque()) {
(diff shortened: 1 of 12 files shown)


sheredom:
I couldn't work out from the issue - but in theory would we need this again in future if we tried to implement _all_ the OpenCL image builtins?

dneto0:
Yes, if we end up emitting an OpImageRead instruction.  But it should be conditional on the finer-grain test if that specific operation is used.
I haven't reviewed the full list of OpenCL image builtins to know.

sheredom:
Ok cool! Ship it 👍

dneto0:
Thanks for the review.

dneto0:
Pushed into master as https://github.com/google/clspv/commit/565571c385801d97f1c68a59ff6d0c58f61eaa50

Issue #38, opened 2017-08-25
Fails to build with latest SPIRV-Headers


dneto0:
The problem is a duplicate enum in a case table, due to token aliasing

```
include/clspv/spirv_c_strings.hpp: In function ‘const char* spv::getCapabilityName(spv::Capability)’:
include/clspv/spirv_c_strings.hpp:410:5: error: duplicate case value
     case spv::CapabilityShaderViewportIndexLayerNV: return "ShaderViewportIndexLayerNV"; 
     ^
include/clspv/spirv_c_strings.hpp:409:5: error: previously used here
     case spv::CapabilityShaderViewportIndexLayerEXT: return "ShaderViewportIndexLayerEXT"; 
     ^
ninja: build stopped: subcommand failed.
```

PR #39, opened 2017-08-25
Ignore ShaderViewportIndexLayerNV capability due to alias

dneto0:
This is required to be able to build against latest SPIRV-Headers

Fixes https://github.com/google/clspv/issues/38


dneto0:
Passes both release and debug build, with latest SPIRV-Headers

dneto0:
LGTM

PR #41, opened 2017-08-25
Use separate UniqueVector to track constants

dneto0:
Avoid polluting the main ValueMap until we actually
generate the SPIR-V constants.  It's less confusing
this way, and matches how we track and generate types.


dneto0:
Tested on Release and Debug build.
I like this code cleanup.
LGTM

Issue #15, opened 2017-07-14
Use of uchar4 generates second copy of OpTypeInt 32 0


dneto0:
That's invalid by the SPIR-V spec.  The validator briefly had that rule turned on, but some bad Vulkan CTS tests have duplicate scalar type decls that are hard to rewrite, so we turned that rule off.

Example input:
```
kernel void foo(global uchar4* A) {
  *A = (uchar4)(1,2,3,4);
}
```

Appears to be caused by SPIRVProducerPass::GenerateSPIRVTypes where it says "i8 is added to TypeMap as i32." 
The issue is that the Types collection (returned by getTypeList) has entries for i8 and i32.  But the i8 will be converted at the last second to an i32.  So we end up with duplicate OpTypeInt 32 0.

Here's a fragment of the generated assembly:

```
%1 =    OpTypeInt 32 0    ; the first one
%2 =    OpTypePointer StorageBuffer %1
%3 =    OpTypeRuntimeArray %1
%4 =    OpTypeStruct %3
%5 =    OpTypePointer StorageBuffer %4
%6 =    OpTypeInt 32 0 ;   oops
%7 =    OpTypeVoid
%8 =    OpTypeFunction %7
%9 =    OpTypeVector %6 3
```

PR #42, opened 2017-08-26
SPIR-V type for i8 always aliases the type for i32

dneto0:
This is incrementally better than what we have now, but I suspect
we should have remapped the types earlier in the flow, and we should
never see the i8 at this point.

Fixes https://github.com/google/clspv/issues/15


dneto0:
LGTM

Issue #36, opened 2017-08-25
Generation of 32bit value for <4 x i8> looks wrong


dneto0:
When accumulating the bits, this:
```
       IntValue = (IntValue << Idx) | Val;      
```
Should probably be:
```
       IntValue = (IntValue << 8 * Idx) | Val;      
```

dneto0:
This example shows the problem, and more:
```
kernel void foo(global uchar4* A) {
 *A = (uchar4)(7,8,9,10);
}
```
Generates code like:
```
     %uint_7 = OpConstant %uint 7
     %uint_8 = OpConstant %uint 8
     %uint_9 = OpConstant %uint 9
    %uint_10 = OpConstant %uint 10
         %17 = OpConstantComposite %uint %uint_7 %uint_8 %uint_9 %uint_10
     ...
         %24 = OpFunction %void None %8
         %25 = OpLabel
         %26 = OpAccessChain %_ptr_StorageBuffer_uint %23 %uint_0_0 %uint_0_0
               OpStore %26 %17
```
That OpConstantComposite is just wrong.

PR #43, opened 2017-08-26
Fix generation of <4 x i8> constants

dneto0:
Cover both when all components are int constants, or
when some are undef values.

Fixes https://github.com/google/clspv/issues/36


dneto0:
LGTM

Issue #44, opened 2017-08-26
Invalid OpLogicalNot when masking <4 x i8> values


dneto0:
Example:
```
kernel void foo(global uchar4* A, int n) {
 *A = (uchar4)((uchar)n,1,2,3);
}
```
Generates code like this:
```
         %27 = OpFunction %void None %10
         %28 = OpLabel
         %29 = OpAccessChain %_ptr_StorageBuffer_uint %25 %uint_0 %uint_0
         %30 = OpAccessChain %_ptr_StorageBuffer_uint_0 %26 %uint_0
         %31 = OpLoad %uint %30
         %32 = OpUConvert %uint %31
         %33 = OpShiftLeftLogical %uint %uint_255 %uint_0
         %34 = OpLogicalNot %uint %33   ; THIS IS WRONG
         %35 = OpBitwiseAnd %uint %uint_66051 %34
         %36 = OpShiftLeftLogical %uint %32 %uint_0
         %37 = OpBitwiseOr %uint %35 %36
               OpStore %29 %33
               OpReturn
```

dneto0:
Additionally, there are peephole optimization opportunities here.

PR #45, opened 2017-08-27
Fix insertion into <4 x i8>

dneto0:
- Use Not rather than LogicalNot
- Register the last SPIR-V instruction ID for
  the LLVM insertion.

Fixes https://github.com/google/clspv/issues/44


dneto0:
Tested on Release and Debug.
LGTM

Issue #48, opened 2017-08-28
Truncation from i32 to i8 should keep only the bottom 8 bits


dneto0:
Example from test/char4_insert.cl

```
kernel void foo(global uchar4* A, int n) {
 *A = (uchar4)(1,2,(uchar)n,4);
}
```
Currently produces:
```
         %27 = OpLoad %uint %26
         %28 = OpUConvert %uint %27  ;  This is a no-op.
         %29 = OpLoad %uint %25
         %30 = OpShiftLeftLogical %uint %uint_255 %uint_8
         %31 = OpNot %uint %30
         %32 = OpBitwiseAnd %uint %29 %31
         %33 = OpShiftLeftLogical %uint %28 %uint_8
         %34 = OpBitwiseOr %uint %32 %33
               OpStore %25 %34
```
The UConvert to produce %28 is a no-op.  The i8 LLVM type maps to a 32-bit SPIR-V int.   The LLVM truncation should become an "and" with 255.

PR #49, opened 2017-08-28
Fix truncation of i32 to i8

dneto0:
Fixes https://github.com/google/clspv/issues/48


dneto0:
LGTM

PR #51, opened 2017-08-30
Refresh SPIRV-Tools dependency

dneto0:
Pick up fix for SPIRV-Tools issue 782


dneto0:
Tested on Release and Debug

PR #53, opened 2017-09-01
Update SPIRV-Tools and SPIRV-Headers

dneto0:
Pick up first SPIRV-Tools v2017.1-dev commits


dneto0:
LGTM

Issue #55, opened 2017-09-06
uchar4 component extraction is broken:  Not seeing the effect of the mask with 255


dneto0:
Example input:
```
kernel void foo(const global float4* IN, global float4* OUT) {
  uchar4 in4 = *((global const uchar4*)IN);
  float4 result;
  result.x = in4.x;
  result.y = in4.y;
  result.z = in4.z;
  result.w = in4.w;
  *OUT = result;
}
```

dneto0:
Produces: 
```
         %35 = OpLoad %v4float %34
         %36 = OpCompositeExtract %float %35 0
         %37 = OpBitcast %uint %36
         %38 = OpShiftRightLogical %uint %37 %uint_0
         %39 = OpBitwiseAnd %uint %38 %uint_255
         %40 = OpConvertUToF %float %38 ;  This is wrong. Should use %39 not %38
         %41 = OpCompositeInsert %v4float %40 %16 0
         %42 = OpShiftRightLogical %uint %37 %uint_8
         %43 = OpBitwiseAnd %uint %42 %uint_255
         %44 = OpConvertUToF %float %42
         %45 = OpCompositeInsert %v4float %44 %41 1
         %46 = OpShiftRightLogical %uint %37 %uint_16
         %47 = OpBitwiseAnd %uint %46 %uint_255
         %48 = OpConvertUToF %float %46
         %49 = OpCompositeInsert %v4float %48 %45 2
         %50 = OpShiftRightLogical %uint %37 %uint_24
         %51 = OpBitwiseAnd %uint %50 %uint_255
         %52 = OpConvertUToF %float %50
         %53 = OpCompositeInsert %v4float %52 %49 3
               OpStore %33 %53
               OpReturn
               OpFunctionEnd
```

PR #56, opened 2017-09-06
Fix extraction of uchar4 components

dneto0:
Fixes https://github.com/google/clspv/issues/55


dneto0:
LGTM

PR #58, opened 2017-09-07
replace bitcasts: Fix elem indices for store into cast wider-vec to narrow-vec

dneto0:


dneto0:
Also tested some cases in hardware.
LGTM

PR #59, opened 2017-09-15
WIP: Read WorkgroupSize as whole vector from the spec constant value

dneto0:
Change the implementation of the get_local_size implementation to
avoid some driver bugs:
- One issue is trouble extracting individual elements.  It's seems to
  work better to read the entire vector at once.
- Another issue is that we have to extract directly from the spec
  composite value directly rather than from a variable initialized with
  that spec constant value.

### DIFF
--- a/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
+++ b/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
@@ -111,14 +111,22 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineMappedBuiltin(Module &M, StringRef 
   Value *Select1 =
       Builder.CreateSelect(Cond, &*F->arg_begin(), Builder.getInt32(0));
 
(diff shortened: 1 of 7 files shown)


dneto0:
Tested on release and debug builds, and in hardware.

dneto0:
LGTM

dneto0:
No, not so fast.  Doesn't work in some scenarios.  My HW case wasn't right.

sheredom:
I did see this bug on NVIDIA, but after filing it their latest drivers have fixed the issue fwiw.

dneto0:
> I did see this bug on NVIDIA, but after filing it their latest drivers have fixed the issue fwiw.

Yes, I had the same experience.

dneto0:
Rebased.
Tested in HW, and with Release and Debug builds.

I'd like to revisit this hack in the future.

dneto0:
LGTM

Issue #54, opened 2017-09-06
vector form of clz builtin produces invalid code


dneto0:
Example  test/IntegerBuiltins/int2_clz.cl  produces this (fragment):

```
         %14 = OpFunction %void None %9
         %15 = OpLabel
         %16 = OpAccessChain %_ptr_StorageBuffer_v2uint %12 %uint_0 %uint_0
         %17 = OpAccessChain %_ptr_StorageBuffer_v2uint %13 %uint_0 %uint_0
         %18 = OpLoad %v2uint %17
         %19 = OpExtInst %v2uint %1 FindUMsb %18
         %20 = OpISub %v2uint %uint_31 %19 ;   Scalar mixed with vector
               OpStore %16 %20
               OpReturn
```
The problem is the ISub operates on an scalar and a vector.

dneto0:
Hey @atgoo your recent validator upgrades are paying off. It found this bug.

PR #60, opened 2017-09-17
Fix codegen for vector clz

dneto0:
In the subtraction, use a splat vector of value 31 rather than scalar 31.

Fixes https://github.com/google/clspv/issues/54


dneto0:
LGTM

PR #64, opened 2017-09-29
Use LLVM's AtomicRMWInst for most of the atomics instructions.

sheredom:
LLVM provides an AtomicRMWInst that satisfies most of the spirv.atomic
'fake' intrinsics we used previously, so lets switch over to use that.
This also allows other languages that aren't coming from a CL
perspective to use AtomicRMWInst and have clspv output the correct
instructions.

I also fixed a bug with the mem_fence replacements - having two maps of
const char* and then comparing if the two pointers matched would compare
the two separate binary blobs of the const char* (EG. their pointers)
which might not match (depending on the linker). Instead I changed it to
a std::map where the second element is a std::tuple.

Lastly, MSVC would complain with SPIRVProducerPass so I fixed all the warnings-as-errors that were produced.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -48,6 +48,6 @@ add_dependencies(clspv_core clspv_c_strings clspv_glsl)
 
 if (MSVC)
   set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/SPIRVProducerPass.cpp"
-    PROPERTIES COMPILE_FLAGS "/Wall /WX /wd4710 /wd4820 /wd4625 /wd4626 /wd5026 /wd5027 /wd4061 /wd4711 /wd4996"
+    PROPERTIES COMPILE_FLAGS "/Wall /WX /wd4710 /wd4820 /wd4625 /wd4626 /wd5026 /wd5027 /wd4061 /wd4711 /wd4996 /wd4530 /wd4577"
   )
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,8 +22,8 @@ endif()
 
 # Check if required third-party dependencies exist.
 macro(use_component path)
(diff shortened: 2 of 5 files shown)


dneto0:
I just saw this. I'll take a look at it next week.
And thanks for the contribution!

sheredom:
@dneto0 gurgh, the lack of CI strikes me again. I'll get it building on Linux and get back to you.

sheredom:
I've pushed what I hope is the change - but this week I'm away on business and without my Linux machine (only my Win10 laptop). Happy to wait until I'm back at my Linux machine to push this PR further if that helps.

lib/CMakeLists.txt
```diff
@@ -48,6 +48,6 @@ add_dependencies(clspv_core clspv_c_strings clspv_glsl)
 
 if (MSVC)
   set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/SPIRVProducerPass.cpp"
-    PROPERTIES COMPILE_FLAGS "/Wall /WX /wd4710 /wd4820 /wd4625 /wd4626 /wd5026 /wd5027 /wd4061 /wd4711 /wd4996"
+    PROPERTIES COMPILE_FLAGS "/Wall /WX /wd4710 /wd4820 /wd4625 /wd4626 /wd5026 /wd5027 /wd4061 /wd4711 /wd4996 /wd4530 /wd4577"
   )
```
dneto0: What's 4577?  I didn't see it on msdn

https://msdn.microsoft.com/en-us/library/mt694070.aspx


The closest I could find was for HLSL : 
https://msdn.microsoft.com/en-us/library/windows/desktop/dn508414(v=vs.85).aspx

dneto0:
Rebased and pushed into master as 3967210de9e5

Issue #65, opened 2017-09-29
generates OpSelect with scalar bool condition with vector data operands


dneto0:
If operands 1 and 2 are vectors then condition operand must be a vector with the same number of elements.

The compiler is currently generating OpSelect with a scalar bool condition operand.  This is passed through from LLVM.

dneto0:
Example 
```
void kernel __attribute__((reqd_work_group_size(1, 1, 1))) foo(global float2* a, int c)
{
  *a = c ? (float2)(1.0,2.0) : (float2)(3.0,4.0);
}
```
Generates:
```
...
       %bool = OpTypeBool
     %uint_0 = OpConstant %uint 0
    %float_3 = OpConstant %float 3
    %float_4 = OpConstant %float 4
         %17 = OpConstantComposite %v2float %float_3 %float_4
    %float_1 = OpConstant %float 1
    %float_2 = OpConstant %float 2
         %20 = OpConstantComposite %v2float %float_1 %float_2
         %21 = OpVariable %_ptr_StorageBuffer__struct_5 StorageBuffer
         %22 = OpVariable %_ptr_StorageBuffer__struct_8 StorageBuffer
         %23 = OpFunction %void None %12
         %24 = OpLabel
         %25 = OpAccessChain %_ptr_StorageBuffer_v2float %21 %uint_0 %uint_0
         %26 = OpAccessChain %_ptr_StorageBuffer_uint %22 %uint_0
         %27 = OpLoad %uint %26
         %28 = OpIEqual %bool %27 %uint_0
         %29 = OpSelect %v2float %28 %17 %20
               OpStore %25 %29
               OpReturn
```
Condition %28 is scalar bool but selects between vectors %17 and %20

dneto0:
I'm claiming this.

PR #68, opened 2017-10-03
OpSelect: vectorize bool condition for vector operands

dneto0:
For an OpSelect with vector data operands, the condition
must be a bool vector with the same number of elements.
In this way SPIR-V is a slightly more restrictive than
LLVM IR.

Fixes https://github.com/google/clspv/issues/65


dneto0:
Tested on release and debug builds.
LGTM

Issue #67, opened 2017-10-02
By default, reuse same descriptor set number for all kernels in a compilation unit


dneto0:
When creating a compute shader pipeline, you can only supply a single entry point.  We can reduce the "pressure" on descriptor set counts by reusing the same descriptor set number for all kernels in a compilation unit.  (It would be 0 if no sampler map, and 1 if there is a sampler map.)

Add a command line option to keep the old behaviour.

PR #69, opened 2017-10-03
By default, reuse descriptor sets across kernels

dneto0:
Use option -distinct-kernel-descriptor-sets to retain old behaviour


dneto0:
Tested on Linux Release and Debug builds.
LGTM

dneto0:
Updated commit message to mention it fixes #67

PR #70, opened 2017-10-03
Update SPIRV-Tools and SPIRV-Headers

dneto0:
SPIRV-Tools adds:
- optimizer fixes and speed improvements
- optimization pass --eliminate-dead-functions
- validator does more type checking


dneto0:
Tested on Linux Release and Debug builds
LGTM

PR #72, opened 2017-10-05
cmake: Enable cmake install for include/lib/bin

robertfoss:
Instruct cmake to install headers, libclspv_core, and the clspv binary.

Signed-off-by: Robert Foss <robert.foss@collabora.com>


dneto0:
Rebased and pushed into master as fb1d6916.

I filed followup https://github.com/google/clspv/issues/76 to document CMake settings such as this.

PR #75, opened 2017-10-06
Change argType to argKind in descriptormap file

dneto0:
argKind is better because it doesn't have the connotation of providing OpenCL C type information.
It's a reduction of that.


dneto0:
Works on debug and release builds.
LGTM

Issue #3, opened 2017-07-13
On abnormal error, the output file can remain. Causes problems in build systems


dneto0:
Should have auto-cleanup behaviour like Clang does.

PR #78, opened 2017-10-13
Write output files only if compilaion succeeds

dneto0:
Fixes https://github.com/google/clspv/issues/3


dneto0:
The downside of this is that in case of file errors (e.g. the dir does not exist), then you only get the error after you've spent time compiling.   But this is better than leaving 0-length files around to confuse the user and the build system.

Note that Clang has a different strategy for dealing with this issu.

dneto0:
Tested manually on Release and Debug builds.
LGTM

PR #80, opened 2017-10-13
Report call to unknown function

dneto0:
The compiler doesn't error out, but will emit a call to a function
with id 0.  I don't know how to error out from within the code
generator.

This will make it easier to debug the case where we are missing
support for a builtin.


dneto0:
LGTM

PR #90, opened 2017-10-16
Support fract, with pointer-to-private variants

dneto0:
The tests have to workaround the disassembler rounding up the nearly-1
value up to 1.0


dneto0:
LGTM

PR #92, opened 2017-10-18
Ignore more warnings in vs2017 to make it build

Jasper-Bekkers:
Visual Studio 2017 complains about a few more things in the standard header files. This makes it shut up.

Alternative would be to disable warnings-as-errors entirely so it's easier to build in the future.


dneto0:
Rebased and pushed onto master as 1dc1d15182

Issue #95, opened 2017-10-18
Some drivers don't like constants constructed with OpUndef scalar components


dneto0:
Example input:
```
void kernel foo(global float4* A)
{
  float4 value;
  value.w = 1111.0f;
  *A = value;
}
```
Produces assembly including this:

```
         %13 = OpUndef %float
 %float_1111 = OpConstant %float 1111
         %15 = OpConstantComposite %v4float %13 %13 %13 %float_1111
         %16 = OpSpecConstant %uint 1
         %17 = OpSpecConstant %uint 1
         %18 = OpSpecConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %16 %17 %18
         %20 = OpVariable %_ptr_Private_v3uint Private %gl_WorkGroupSize
         %21 = OpVariable %_ptr_StorageBuffer__struct_5 StorageBuffer
         %22 = OpFunction %void None %9
         %23 = OpLabel
         %24 = OpAccessChain %_ptr_StorageBuffer_v4float %21 %uint_0 %uint_0
               OpStore %24 %15
               OpReturn
               OpFunctionEnd
```
Some drivers have trouble with the OpConstantComposite.  They happen to work if we use 
  %13 = OpConstantNull %float
instead.

dneto0:
Use -hack-undef to work around this problem.

dneto0:
This is solved well enough now.

PR #96, opened 2017-10-18
Add -hack-undef: Use OpConstantNull instead of OpUndef

dneto0:
Work around drive bug documented in https://github.com/google/clspv/issues/95


dneto0:
LGTM

Issue #94, opened 2017-10-18
Writing all zero's to a pointer generates incorrect code


Jasper-Bekkers:
Two test cases:

This one is broken and generates only one OpStore (for one float):
```
__kernel void myTest(__global float* jasper)
{
  *jasper++ = 0;
  *jasper++ = 0;
  *jasper++ = 0;
  *jasper++ = 0;
}
```

```
; SPIR-V
; Version: 1.0
; Generator: Codeplay; 0
; Bound: 22
; Schema: 0
               OpCapability Shader
               OpCapability VariablePointers
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpExtension "SPV_KHR_variable_pointers"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %19 "myTest"
               OpSource OpenCL_C 120
               OpDecorate %13 SpecId 0
               OpDecorate %14 SpecId 1
               OpDecorate %15 SpecId 2
               OpDecorate %_runtimearr_float ArrayStride 4
               OpMemberDecorate %_struct_4 0 Offset 0
               OpDecorate %_struct_4 Block
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
               OpDecorate %18 DescriptorSet 0
               OpDecorate %18 Binding 0
      %float = OpTypeFloat 32
%_ptr_StorageBuffer_float = OpTypePointer StorageBuffer %float
%_runtimearr_float = OpTypeRuntimeArray %float
  %_struct_4 = OpTypeStruct %_runtimearr_float
%_ptr_StorageBuffer__struct_4 = OpTypePointer StorageBuffer %_struct_4
       %uint = OpTypeInt 32 0
       %void = OpTypeVoid
          %8 = OpTypeFunction %void
     %v3uint = OpTypeVector %uint 3
%_ptr_Private_v3uint = OpTypePointer Private %v3uint
     %uint_0 = OpConstant %uint 0
    %float_0 = OpConstant %float 0
         %13 = OpSpecConstant %uint 1
         %14 = OpSpecConstant %uint 1
         %15 = OpSpecConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %13 %14 %15
         %17 = OpVariable %_ptr_Private_v3uint Private %gl_WorkGroupSize
         %18 = OpVariable %_ptr_StorageBuffer__struct_4 StorageBuffer
         %19 = OpFunction %void None %8
[...]

PR #97, opened 2017-10-18
Fix memset of null value spanning multiple pointee locations

dneto0:
Fixes https://github.com/google/clspv/issues/94


dneto0:
Tested on Release and Debug.
LGTM

PR #99, opened 2017-10-20
Add support for vector load/store.

derekjchow:
Here's my first attempt at adding support for vload/vstore. Currently only supports float vload4 and vstore4, but it should be trivial to extend this to more types.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -96,6 +96,8 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
   bool replaceAtomics(Module &M);
   bool replaceCross(Module &M);
   bool replaceFract(Module &M);
+  bool replaceVload(Module &M);
+  bool replaceVstore(Module &M);
 };
 }
 
@@ -133,6 +135,8 @@ bool ReplaceOpenCLBuiltinPass::runOnModule(Module &M) {
   Changed |= replaceAtomics(M);
   Changed |= replaceCross(M);
   Changed |= replaceFract(M);
+  Changed |= replaceVload(M);
+  Changed |= replaceVstore(M);
 
   return Changed;
 }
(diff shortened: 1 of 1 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -903,6 +907,136 @@ bool ReplaceOpenCLBuiltinPass::replaceMadandMad24andMul24(Module &M) {
...
+  struct VectorStoreOps {
+    const char* name;
+    int n;
+    Type* (*get_scalar_type_function)(LLVMContext&);
+  } vector_store_ops[] = {
+    // TODO(derekjchow): Expand this list.
+    { "_Z7vstore4Dv4_fjPU3AS1f", 4, Type::getFloatTy }
+  };
+
+  for (int i = 0; i < sizeof(vector_store_ops) / sizeof(*vector_store_ops);
```
dneto0: I'd prefer range-for here, for readability
```
for (const auto& vector_store_op : vector_store_ops) {
...
```

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -903,6 +907,136 @@ bool ReplaceOpenCLBuiltinPass::replaceMadandMad24andMul24(Module &M) {
   return Changed;
 }
 
+bool ReplaceOpenCLBuiltinPass::replaceVstore(Module &M) {
+  bool Changed = false;
+
+  struct VectorStoreOps {
+    const char* name;
+    int n;
+    Type* (*get_scalar_type_function)(LLVMContext&);
```
dneto0: I find this a little baroque.  You've already got M so I would have written this as just Type* in the member declaration, and invoked Type::getFloatTy(M.getContext() in the data.

But no big deal. What you did is fine.  :-)

dneto0:
Rebased and pushed into master as cfd368be1
I also replaced the iteration with a range-for and added test cases in 544fffcd

PR #100, opened 2017-10-20
-hack-undef should only replace numeric types

dneto0:
We need to allow pointer-to-image use an Undef value.
This occurs when we pass an image into a helper function
but the helper function ignores the image value.  The optimizer
will replace the call parameter with an Undef pointer-to-image.
That must remain.


dneto0:
Avoids downstream issues with the optimizer.

LGTM.

dneto0:
Rebased and pushed onto master as  c66b335

PR #101, opened 2017-10-23
Cluster pod kernel args: Inline the inner function by default

dneto0:
Use option -no-inline-pod to avoid inlining, i.e. get the old
behaviour.  But it's deprecated.


dneto0:
LGTM

PR #106, opened 2017-10-23
Reuse struct types for kernel pointer arg wrappers

dneto0:


dneto0:
I expect to follow this soon with a further update to reuse kernel arg *variables* between different kernels. Those tests will exercise this enhancement as well.

LGTM.

Issue #103, opened 2017-10-23
Reuse SSBOs for identically shaped kernel pointer-args from a different kernel


dneto0:
Consider this:
```
kernel void foo(global float* A) {
  *A = 1.5f;
}

kernel void bar(global float* B) {
  *B = 2.5f;
}
```
It produces this output:

```
               OpCapability Shader
               OpCapability VariablePointers
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpExtension "SPV_KHR_variable_pointers"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %21 "foo"
               OpEntryPoint GLCompute %24 "bar"
               OpSource OpenCL_C 120
               OpDecorate %14 SpecId 0
               OpDecorate %15 SpecId 1
               OpDecorate %16 SpecId 2
               OpDecorate %_runtimearr_float ArrayStride 4
               OpMemberDecorate %_struct_4 0 Offset 0
               OpDecorate %_struct_4 Block
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
               OpDecorate %19 DescriptorSet 0
               OpDecorate %19 Binding 0
               OpDecorate %20 DescriptorSet 0
               OpDecorate %20 Binding 0
      %float = OpTypeFloat 32
%_ptr_StorageBuffer_float = OpTypePointer StorageBuffer %float
%_runtimearr_float = OpTypeRuntimeArray %float
  %_struct_4 = OpTypeStruct %_runtimearr_float
%_ptr_StorageBuffer__struct_4 = OpTypePointer StorageBuffer %_struct_4
       %uint = OpTypeInt 32 0
       %void = OpTypeVoid
          %8 = OpTypeFunction %void
     %v3uint = OpTypeVector %uint 3
%_ptr_Private_v3uint = OpTypePointer Private %v3uint
     %uint_0 = OpConstant %uint 0
  %float_1_5 = OpConstant %float 1.5
  %float_2_5 = OpConstant %float 2.5
         %14 = OpSpecConstant %uint 1
         %15 = OpSpecConstant %uint 1
         %16 = OpSpecConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %14 %15 %16
         %18 = OpVariable %_ptr_Private_v3uint Private %gl_WorkGroupSize
         %19 = OpVariable %_ptr_StorageBuffer__struct_4 StorageBuffer
[...]

PR #108, opened 2017-10-23
Reuse kernel arg variables for StorageBuffers

dneto0:
We can reuse such a variable if:
 - it is of the same type
 - it was last used for a different function

Fixes https://github.com/google/clspv/issues/103


dneto0:
Superceded by #110

PR #105, opened 2017-10-23
Cluster-pod-args: Reuse struct types

dneto0:
Fixes https://github.com/google/clspv/issues/104

PR #109, opened 2017-10-24
Fix arg type mapping for pod args

dneto0:


dneto0:
This is a fix for the patch in #105

dneto0:
LGTM

PR #111, opened 2017-10-24
Update to latest SPIRV-Tools

dneto0:


dneto0:

LGTM

Issue #114, opened 2017-11-17
spvasm2checks.pl should have a only-changing-numeric-ID mode


antiagainst:
Symbolic names are themselves good checks.

PR #115, opened 2017-11-17
Nicer spvasm2checks

dneto0:
Adds two options:
  --numeric  :  only pattern match against numeric ids
  --re2   : assume RE2 is the matcher, so we can use \w to match ID characters.  (This is good for use with https://github.com://google/effcee )

Fixes https://github.com/google/clspv/issues/114


dneto0:
LGTM. Tested locally

antiagainst:
Mighty  Perl guru @dneto0! ;-P I gonna try out next time.

PR #116, opened 2017-11-18
Set EXCLUDE_FROM_ALL for LLVM and SPIRV-Tools

antiagainst:
This trims the number of default targets to build from 3299 to 1306 on my machine.

This is when you only care about `clspv`, not the `check-spirv` target.


antiagainst:
Building 3299 targets on my MacBook Pro takes ~23 min, while building 1306 targets takes ~10 min.

PR #117, opened 2017-11-18
Remove the cland dependency from the check-spirv target

antiagainst:
This avoids building the whole clang target hierarchy. It trims
the number of additional targets to build from ~1400 to ~70.


dneto0:
Rebased and pushed into master as 0f05a3ac

PR #118, opened 2017-11-20
Rewrite inserts

dneto0:
Rewrite insertions into structs to avoid OpCompositeInsert

By default: replace chains of OpCompositeInsert that cover all members of the struct.  Only rewrites cases where where the insertion replaces a top-level member of the struct (i.e. only has one indexing operand).  It also only recognizes the pattern where the members are inserted in index order.  This is turned on by default since this always makes the code smaller.

With option -hack-inserts, generalize rewrite *partial* chains, i.e. those that don't cover the entire struct.  This is a little more aggressive in that members don't have to be inserted in order.


dneto0:
LGTM

PR #124, opened 2018-03-15
Silence GCC 7.3 warning -Wnoexcept-type for now

dneto0:


dneto0:
Tested with Release and Debug builds.
LGTM

Issue #122, opened 2018-03-15
implement get_image_width and get_image_height


dneto0:
Requires ImageQuery capability. That's required for Vulkan 1.0 support.

dneto0:
Do it for 2D images first.

PR #125, opened 2018-03-15
Support get_image_width get_image_height on 2D

dneto0:
Supports both read_only and write_only images.

Partially addresses https://github.com/google/clspv/issues/122


dneto0:
LGTM

PR #127, opened 2018-04-01
Cluster constants

dneto0:
Add -pod-ubo : put plain-old-data arguments in uniform buffers instead of storage buffers.  This could be faster.

Add -module-constants-in-storage-buffer:  Big change enabled by new option.
  - Don't use Private variables for __constant data
  - Module-scope constants are collected into a single storage buffer
  - That storage buffer must be initialized by the host to contain the right initialization data by the time the pipeline executes
  - The initialization data is emitted to the descriptor map along with the descriptor set and binding
This is more correct, fixing variable-pointers validation for cases where pointer-to-constant was being dereferenced inside a function.

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -283,6 +283,70 @@ produces the following descriptor map:
 
 TODO(dneto): Give an example using images.
 
+#### Module scope constants
+
+By default, each module-scope variable in `__constant` address space is mapped to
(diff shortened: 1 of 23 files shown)


dneto0:
Also fixes a bug:  Before we were not decorating OpTypeArray with ArrayStride (!).  We were only doing it on OpTypeRuntimeArray

dneto0:
Checked on Release and Debug builds

dneto0:
LGTM

PR #129, opened 2018-04-10
Support pointer-to-local kernel arguments

dneto0:
- Descriptor map says it's a local, not a buffer.
- Don't allocate a binding for such variables.
- Maps to OpVariable in Workgroup storge of type OpTypeArray
- Array size is an integer specialization constant, with SpecId unique
  to that integer.
- Document pointer-to-local kernel args
- Add tests

- Refactor SPIRVOperandList into its own class
  Add a shift operator to add operands to the list.
  Add MkId and MkNum helpers to make operands.


dneto0:
Passes self-tests on Release and Debug builds.  Needs verification on actual HW.

dneto0:
LGTM

PR #130, opened 2018-04-11
Simplify code, mostly use SPIRVOperand shift operator

dneto0:
Also Remove dead variable nextSpecID


dneto0:
Verified with self-tests on Release and Debug builds

dneto0:
LGTM

PR #131, opened 2018-04-11
Cleanup: SPIRVInstruction constructor counts its own words

dneto0:
Remove instruction constructor that takes a word count


dneto0:
Self-tested on Release and Debug builds.
LGTM

PR #132, opened 2018-04-25
Add -hack-initializers to work around a driver bug

dneto0:
Use -hack-initializers to work around driver bugs that cause the
get_global_size builtin function to return the wrong value if used with
a non-constant dimension argument.

The underlying issue is that SPIR-V Private variable initializers may
not take effect.  The workaround is to start every kernel with an
explicit store to write the initializer value.  This is only required
for a compiler-generated variable that stores the workgroup size, only
used to implement the get_global_size builtin.


dneto0:
Necessary for some driver versions.
LGTM.

Issue #135, opened 2018-05-28
compiler generates OpPtrAccessChain on pointer-to-Function in a callee function


dneto0:
VariablePointers permits OpPtrAccessChain only on StorageBuffer or Workgroup.  It's not permitted on Function storage class.

Example input:
```
int helper(int arr[]) {
  arr[1] = 22;
}

kernel void foo(global int* A, int n) {
  int arr[2];
  helper(arr);
  *A = arr[1];
}
```
Produces the following:
```
; SPIR-V
; Version: 1.0
; Generator: Codeplay; 0
; Bound: 40
; Schema: 0
               OpCapability Shader
               OpCapability VariablePointers
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpExtension "SPV_KHR_variable_pointers"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %32 "foo"
               OpSource OpenCL_C 120
               OpDecorate %21 SpecId 0
               OpDecorate %22 SpecId 1
               OpDecorate %23 SpecId 2
               OpDecorate %_runtimearr_uint ArrayStride 4
               OpMemberDecorate %_struct_4 0 Offset 0
               OpDecorate %_struct_4 Block
               OpMemberDecorate %_struct_6 0 Offset 0
               OpDecorate %_struct_6 Block
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
               OpDecorate %26 DescriptorSet 0
               OpDecorate %26 Binding 0
               OpDecorate %27 DescriptorSet 0
               OpDecorate %27 Binding 1
               OpDecorate %_arr_uint_uint_2 ArrayStride 4
       %uint = OpTypeInt 32 0
%_ptr_StorageBuffer_uint = OpTypePointer StorageBuffer %uint
%_runtimearr_uint = OpTypeRuntimeArray %uint
  %_struct_4 = OpTypeStruct %_runtimearr_uint
%_ptr_StorageBuffer__struct_4 = OpTypePointer StorageBuffer %_struct_4
  %_struct_6 = OpTypeStruct %uint
%_ptr_StorageBuffer__struct_6 = OpTypePointer StorageBuffer %_struct_6
       %void = OpTypeVoid
          %9 = OpTypeFunction %void
     %uint_2 = OpConstant %uint 2
%_arr_uint_uint_2 = OpTypeArray %uint %uint_2
[...]

PR #137, opened 2018-05-29
Inline functions with ptr to Function storage class

dneto0:
Inline a function if its return value or any argument is
a pointer to the Function storage class.  This avoids many
cases of OpPtrAccessChain on Function storage class.

Fixes https://github.com/google/clspv/issues/135


dneto0:
LGTM

PR #138, opened 2018-05-29
Zero stack vars

dneto0:
Zero-initialize stack variables.

Also generalize some of the replace-pointer-bitcasts code to handle more cases.

Builds on #137


dneto0:
Tested on Release and Debug builds, and an internal test suite (not on HW).

LGTM.

PR #140, opened 2018-06-04
User error if pointer-to-void is used

dneto0:
Pointer-to-void tends to turn into pointer-to-char, which trips up
some things.


dneto0:
LGTM

PR #141, opened 2018-06-04
Print message if literal sampler without sampler map

dneto0:
It prints to stderr, then proceeds to crash.
It's bad, but better than nothing.


dneto0:

dneto0:
LGTM

PR #144, opened 2018-06-08
Add __clspv_vloada_half4, __clspv_vloada_half2

dneto0:
Works on global, local, and private pointers.


dneto0:
Tested on Release and Debug.
LGTM

Issue #143, opened 2018-06-05
Invalid SPIR-V produced


j-schulte:
The SPIR-V ([spvasm.txt](https://github.com/google/clspv/files/2072915/spvasm.txt)) that is generated for the attached kernel ([invalid.txt](https://github.com/google/clspv/files/2072900/invalid.txt)) is rejected by spirv-val with

> error: 62: VectorInsertDynamic: expected Vector type to be equal to Result Type

as the arguments for OpVectorInsertDynamic are incorrect.
Note that the kernel is not spec compliant, as it uses brackets to index a vector type, which is supported by Clang (https://clang.llvm.org/docs/LanguageExtensions.html#vectors-and-extended-vectors) but not 
allowed in the spec. So I'm not sure, whether there is something to be done here at all.

Regards
Johannes

PR #148, opened 2018-06-09
Fix generation of OpVectorInsertDynamic instructions

jrprice:
The previous implementation assumed the same operand order as
OpCompositeInsert, which is not the case. In addition, the index
operand was incorrect.

Fixes https://github.com/google/clspv/issues/143


dneto0:
Rebased and pushed into master as d26efea6e

PR #149, opened 2018-06-13
Tests: Reduce number of digits to match

dneto0:
The SPIRV-Tools disassembler recently started printing
more digits in floating point numbers, to better reproduce
values.

This change accommodates both old and new SPIRV-Tools by truncating
the specified floating point digits to match.


dneto0:
LGTM

PR #150, opened 2018-06-15
Remove trivial GEPs after pointer bitcast removal

dneto0:


dneto0:
LGTM

Issue #142, opened 2018-06-05
Hit assertion "Use still stuck ..." when compiling kernel


j-schulte:
When compiling the attached kernel ([assertion.txt](https://github.com/google/clspv/files/2072886/assertion.txt)) with Clspv, the compilation aborts with the following output:

> While deleting: i8* %
> Use still stuck around after Def is destroyed:  call void @llvm.lifetime.end.p0i8(i64 128, i8* nonnull <badref>)
> Use still stuck around after Def is destroyed:  call void @llvm.lifetime.start.p0i8(i64 128, i8* nonnull <badref>)
> clspv: ../third_party/llvm/lib/IR/Value.cpp:86: llvm::Value::~Value(): Assertion `use_empty() && "Uses remain when a value is destroyed!"' failed.
> Aborted (core dumped)`

Regards
Johannes

dneto0:
This has been fixed with an update about 2 weeks ago.

I'm attaching the generates SPIR-V assembly
[assert.spvasm.txt](https://github.com/google/clspv/files/2104162/assert.spvasm.txt)

johannesschulte:
I'm sorry, I forgot to mention that I'm running clspv with "-pod-ubo -cluster-pod-kernel-args". With these options, the error persists.

j-schulte:
I'm sorry, I forgot to mention that I'm running clspv with "-pod-ubo -cluster-pod-kernel-args". With these flags, the error persists.

dneto0:
Reproduces with just -cluster-pod-kernel-args

dneto0:
Thanks for the high quality bug report.

j-schulte:
Thanks for the quick fixes!

PR #151, opened 2018-06-15
Remove @llvm.lifetime.* intrinsics

dneto0:
The corresponding OpLifetimeStart and OpLifetimeEnd instructions
require Kernel capability, so we can't use them in Vulkan shaders
anyway.  Just remove them.

Fixes https://github.com/google/clspv/issues/142


dneto0:
LGTM.

PR #152, opened 2018-06-16
Current descriptor set index maintained in named metadata

dneto0:
It's in module-scope metadata with name "clspv.descriptor.index"

This will make it easier to allocate descriptor sets in different parts
of the compiler.


dneto0:
LGTM

PR #160, opened 2018-07-26
Allocate descriptor set and binding in an LLVM module pass

dneto0:
Do it in an LLVM pass so we can later use a smarter algorithm
to reuse resource variables.

This is a combination of 93 commits.

Pull IsImageType and IsSamplerType out to ArgKind.cpp

Simplify GetArgKindForType

Create ArgKind as an enum

Add stub for -direct-buffer-access

WIP: Discriminate kernel args into different buckets

Map kernel arg descriminants to unique indices

Create descriptor resource type and var-accessor function

Turn off ReplaceIndirectBufferAccessPass

Add stub for AllocateDescriptorsPass

Pass sampler map into AllocateDescriptorsPass

WIP: Allocate literal samplers

WIP: allocation of storage buffer descriptors

WIP: SPIRVProducer changes for clspv.resource.var

Simple compile fixes (partly)

Refactor to make FindGlobalConstVars

Small tidy-up with calle_name var

Fix typo in pass message for ReplaceIndirectBufferAccessPass

Add TODOs, fix asserts

Make types for resource vars.

Arrays of length zero map to runtime arrays.

WIP: Handle POD args

Add TODO for sampler kernel args

Descriptor allocation records arg kind

Record type info for resource vars

WIP: Remove runtime array gunk

Generate OpVariable for each resource variable

Map the resource var builtin to the generated variable

Also decorate set and binding.
TODO: emit descriptor map info

WIP: stub for sampler and images

Fix extraction of arg kind

Fix finding array types, image component types

Better message when we can't lookup a type

Fix generation of runtime array types

Fix remapping of opaque types. Defer their load.

Can't change their address space.

Add test file proving we need to preserve pointer-to-constant address space

Run verifier after allocation of descriptors.

WIP: Preserve addres space of __constant until SPIRVProducer

OpenCL builtin vars get External linkage

This makes the verifier complain less

Generate storage buffer vars after workgroup size vars

This makes it match previous code gen better, and is harmless

Generate unnamed structs only.  Saves on SPIR-V output

Generate pointer-to-elem before runtime array

Two hacks to make codegen match better

Handle POD for early generation of pointer-to-elem type

Generate NonWritable / NonReadable

Only tested with __constant

Support pointer-to-local

Tweak order of generation of sampled image float type

WIP: Emit desriptor map info

Tweak type generation ordering to pass more tests

Reorderd some type decls in tests

Test fixes

Avoid generating spurious adds with 0

Fix tests

Test fixes

Type declarations and constants can usually be in relaxed order

But respect dominance rules

Type and constants can appear in relaxed order.

SPIR-V checkline gen: types and constants appear in relaxed order

Fix zero-handling in simplify pointer bitcasts

Test fixes

Fract etc. are ReadNone, not also ReadOnly

Avoid making a broken LLVM module. Fails verification otherwise

Test fixes

test fixes

Don't privilege i32

Simplify GEP on GEP: Avoid collapsing ptr-to-constant

(set,binding) for module-scope constant SSBO

Also, don't generate the constants for the initializer of a module-scope
constant that we will get from the SSBO.

Fix typo

Generate descriptor map info for resource vars, and ptr-to-local

Descriptor map info for pointer-to-local

Module-scope constants use a high-numbered descriptor

Update program-scope-constant tests

Update tests

UndoSRet should remove the StructRet attribute

ReadNone and ReadOnly are incompatible

Fix attribute copying.  Must skip first param

Update bool tests

Generate load for image param only if it has uses

Update tests

Update tests

Update tests

Update tests

Consider avoiding descriptors for unused args

Update tests

Update tests

fix test

Fix counting of args in presence of ptr-to-local

Fix test

With cluster pod kernel args, local ptr args show up in order

Remove stray debug output

Reuse descriptors when same kernel args match across all kernels

Fix tests

Don't show allocator debug output by default

Fix descriptor map in test

Handle -distinct-kernel-desciptor-sets

Fix distinct-kernel-descriptor-sets

Avoid iterating over resource variables in use order. That can be weird.
Instead, iterate in module-order.

Avoid warning on Mac

Fix resource vars for translation units with lots of params


dneto0:
This is not ready yet.

PR #163, opened 2018-08-01
Direct Resource Access optimization

dneto0:
Where possible, convert helper function accesses to resource variables
into direct accesses of that resource variable.  This happens under
the following conditions:

- Affects kernel arguments of the following type, which are converted
  to resource variables that are associated with Vulkan descriptors:
   - pointer-to-global
   - pointer-to-constant
   - image
   - sampler
- All callees of the helper function are using the same resource
  variable. (E.g. Generally, same kernel argument index and same type.)
- The base pointer of the resource is passed in (directly, without
  modification or sub-object access chain)

Use option -no-dra to disable this optimization.

Use of -distinct-kernel-descriptor-sets disables this optimiztaion.

List of commits included
------------------------

1.
Resource var call also saves discriminant index

This is more conveneint and explicit than trying to pick
off the pointer type or get the last component of the function
name.

2.
Compute call-level ordering of reachable functions via
Kahn's topological sort algorithm.

3.
Descriptor map should only get info for kernels

4.
Update tests.
In two cases, also verify the old behaviour via -no-dra

5.
spvasm2checks: Don't emit CHECK-DAG

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -21,9 +21,9 @@ bool F16BitStorage();
 // Returns true if each kernel must use its own descriptor set for all arguments.
 bool DistinctKernelDescriptorSets();
 
-// Returns true if the compiler should try to use direct buffer accesses
+// Returns true if the compiler should try to use direct resource accesses
 // within helper functions instead of passing pointers via function arguments.
-bool DirectBufferAccess();
+bool DirectResourceAccess();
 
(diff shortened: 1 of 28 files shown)


dneto0:
Tested with:
 - Release build
 - Debug build
 - Compilation of an internal suite of kernels
 - Verification in HW of an internal test suite

dneto0:
LGTM

PR #164, opened 2018-08-01
Update known-good SPIRV-Tools

dneto0:
Includes -combine-access-chain


dneto0:
Verified along with #163 

LGTM

PR #165, opened 2018-08-01
Direct resource access: Fix gep zero counting

dneto0:
When nesting one GEP inside another, account for the fact
that the first ("element" or "slide") index will be combined
with the last element of the previous (dominating) GEP.
If we don't do this then the type accounting is wrong and we
generate an invalid GEP.


dneto0:
Fixes an assert failure on the internal shader suite found with the Debug build.

Passes Release and Debug build integrated tests.
Passes in HW on internal test suite.

LGTM

Issue #166, opened 2018-08-03
reinterpret conversions as_type is not supported


dneto0:
See 6.2.4.2 of OpenCL C 2.0

Example:

kernel void foo(global float *A, uint a) {
  *A = as_float(a);
}

yields:

$ clspv b.cl
b.cl:2:8: error: implicit declaration of function 'as_float' is invalid in OpenCL
  *A = as_float(a);
       ^
b.cl:2:8: error: function with no prototype cannot use the spir_function calling convention

PR #167, opened 2018-08-05
Use \n as delimiter for stripping OpenCL features

jrprice:
The previous regex was matching a semicolon to end the string to be removed, which was nuking many additional lines when matching macro definitions like `as_type()`.

This approach just matches entire lines that contain the target string, which I assume was (at least roughly) the intent of the original approach.

As well as removing all of the `as_type` macros as per issue #166, the old regex was also getting rid of `get_work_dim` (hence why the hack to add this back in can now be removed), and a couple of kernel attribute definitions.

Here's a diff showing how this change affects the generated header:
https://gist.github.com/jrprice/05d8485880b3a7eb42ad3857a62e0bb5

The only slightly strange side-effect is a slightly mangled comment (the end of the above diff), but I assume this isn't a big problem.


dneto0:
Thank you!   I'll test locally.

dneto0:
Rebased and pushed into master as  76542919ae

PR #170, opened 2018-08-09
Add -hack-scf: signed integer comparisons fixup

dneto0:
With -hack-scf, replace signed integer comparisons with other
equivalent instructions.


dneto0:
Tested on HW with a ClspvTest test bench

dneto0:
LGTM

PR #172, opened 2018-08-16
Rework pointer-to-local SpecId assignment and apply DRA in more cases

alan-baker:
SpecIds for pointer-to-local kernel arguments are now assigned during AllocateDescriptorsPass. The SpecIds are shared between kernels for arguments of the same types where applicable. Additionally, new functions are that replace uses of the kernel arguments.

Example:
kernel void foo(local int* a, local int* b);
kernel void bar(local int* c, local float* d);

In the above example, three SpecIds will be generated and `a` and `c` will share a SpecId, while `b` and `d` will each get their own.

The assignments are stored as module-level metadata that is read during the SPIRVProducerPass. Much of the producer logic surrounding pointer-to-local kernel arguments has been modified. The new functions that access workgroup variables do not codegen as instructions. They merely act as references to the variable id. The return type of these functions would appear to be a pointer to a runtime array of the appropriate element type. This is used to ensure the appropriately indexed getelementptr instructions are generated. Some care is taken avoid generating unnecessary types runtime array types.

DirectResourceAccess is now also applied to pointer-to-local kernel arguments and global variables. This can eliminate many instances of OpPtrAccessChain from being generated in the output SPIR-V.

I have tested this against [ClspvTest](https://github.com/ericberdahl/ClspvTest) in addition to the unit tests.


dneto0:
Rebased, squashed, and pushed into master as 202c8c7615451677

PR #174, opened 2018-08-21
Share global workgroup variables

alan-baker:
Introduces new pass to merge global workgroup variables of the same type if they are used by distinct sets of kernels. This pass runs before direct resource access because it creates more opportunities for direct resource access.

* Fixed a small bug in the debug output of DRA.
* Ran clang-format on affected files


dneto0:
Rebased, squashed, and pushed into master as fc6888e6fa

PR #175, opened 2018-08-24
New pass to inline single call site functions

alan-baker:
* Adds a pass that inlines functions with a single call site
 * Added -no-inline-single option to disable the pass
* Pass is disabled in many tests to avoid substantial updates
* Added new lit tests for the pass


dneto0:
Rebased, squashed, and pushed onto master as 0dd3fd2090d

Issue #176, opened 2018-08-28
Assertion in replace pointer bitcasts


alan-baker:
```
struct a {
  int x;
  int y;
};

kernel void foo(global struct a* struct_out, int n) {
  struct a local_a;
  if (n == 0) {
    local_a.x = 0;
  }
  *struct_out = local_a;
}
```

LLVM translates this at some point the flow to have:
```
store i64 0, i64 addrspace(1)* %0, align 4
```

ReplacePointerBitcastPass sees the bitwidths are equal and tries to bitcast the i64 to a { i32, i32} which is invalid.

```
Assertion failed: (CastInst::castIsValid(opc, C, Ty) && "Invalid constantexpr cast!"), function getCast, file ../third_party/llvm/lib/IR/Constants.cpp, line 1440.
```

dneto0:
Need to use the debug build to see the assertion.
The release build generates bad code. :-(

dneto0:
I'm experimenting with adding the ReplaceLLVMIntrinsics pass before inst-combine.  

The theory is that the bitcast is injected by SPIR 1.2 initial codegen to prepare an argument to a memcpy.    In this case, with zero-initialization of variables, the initial store to the function-scope variable ends up being scalarized to an i64 store of a zero value.  Then replace-pointer-bitcasts can't cope with the conversion between i64 pointer and pointer to a struct.

That seems to work, but causes changes in a few tests.  I'm evaluating whether it's not too annoying a difference.

dneto0:
Instead, if we replace LLVM intrinsics early, then we convert the llvm memcpy to a spirv.copy_memory that instcombine doesn't recognize:

```
*** IR Dump Before Combine redundant instructions ***
; Function Attrs: nounwind
define spir_kernel void @foo(%struct.a addrspace(1)* %struct_out, i32 %n) #0 !kernel_arg_addr_space !3 !kernel_arg_access_qual !4 !kernel_arg_type !5 !kernel_arg_base_type !5 !kernel_arg_type_qual !6 {
  %local_a = alloca %struct.a, align 4
  store %struct.a zeroinitializer, %struct.a* %local_a
  %cmp = icmp eq i32 %n, 0
  br i1 %cmp, label %1, label %2

; <label>:1:                                      ; preds = %0
  %x = getelementptr inbounds %struct.a, %struct.a* %local_a, i32 0, i32 0
  store i32 0, i32* %x, align 4
  br label %2

; <label>:2:                                      ; preds = %1, %0
[...]

PR #177, opened 2018-08-28
Extend hack undef to handle all nullable types

alan-baker:
Not all types are valid to use in OpConstantNull in SPIR-V. This PR extends -hack-undef to generate OpConstantNull for all possible cases.

Test works around #176.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -6129,7 +6129,6 @@ void SPIRVProducerPass::WriteSPIRVBinary() {
 }
 
 bool SPIRVProducerPass::IsTypeNullable(const Type* type) const {
-  outs() << "IsTypeNullable: " << *type << "\n";
   switch (type->getTypeID()) {
     case Type::HalfTyID:
     case Type::FloatTyID:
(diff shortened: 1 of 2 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -6127,3 +6127,40 @@ void SPIRVProducerPass::WriteSPIRVBinary() {
...
+      const StructType* struct_type = cast<StructType>(type);
+      // Images and samplers are not nullable.
+      if (struct_type->isOpaque()) return false;
+      for (const auto element : struct_type->elements()) {
+        if (!IsTypeNullable(element)) return false;
+      }
+      return true;
+    }
+    default:
+      return false;
```
dneto0: I suspect some compilers will not do the control flow analysis to see that all paths lead to a return of a bool.
I would code this as default: break; and then return false as last statement in the function.

But let's wait until we see that problem.

lib/SPIRVProducerPass.cpp
```diff
@@ -6127,3 +6127,40 @@ void SPIRVProducerPass::WriteSPIRVBinary() {
...
+      return true;
+    }
+    case Type::ArrayTyID:
+      return IsTypeNullable(cast<CompositeType>(type)->getTypeAtIndex(0u));
+    case Type::StructTyID: {
+      const StructType* struct_type = cast<StructType>(type);
+      // Images and samplers are not nullable.
+      if (struct_type->isOpaque()) return false;
+      for (const auto element : struct_type->elements()) {
+        if (!IsTypeNullable(element)) return false;
```
dneto0: This bottoms out because the pointer check doesn't recurse. So this is ok.

dneto0:
Pushed onto master as 9bf93fb26

PR #179, opened 2018-08-29
Extends complete insertion chain replacement to arrays

alan-baker:
* Complete insertion replacement now handle array types
 * Some annoying rewrites due to CompositeType not providing some common
 functionality.
 * new test


dneto0:
Rebased, squashed, and pushed into master as f866141

PR #180, opened 2018-08-29
Add the ability to inline all kernels exhaustively

alan-baker:
* Gated by option
* new pass exhaustively inlines all kernels
* added a test

### DIFF
--- a/lib/InlineEntryPointsPass.cpp
+++ b/lib/InlineEntryPointsPass.cpp
@@ -1,0 +1,94 @@
+// Copyright 2018 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <vector>
+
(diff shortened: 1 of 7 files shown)


lib/InlineEntryPointsPass.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  // Clean up dead functions. This done here to avoid ordering requirements on
+  // inlining.
+  std::vector<Function *> to_delete;
+  for (auto &F : M) {
+    if (F.isDeclaration() || F.getCallingConv() == CallingConv::SPIR_KERNEL)
+      continue;
+
+    if (F.user_empty())
```
dneto0: I would have put this in a convergence loop.  We might visit a leaf function before one of its callees.  In which case the leaf sticks around, incorrectly.

alan-baker: This example works and is roughly the same:
```
int func_3(local int *in, int n) { return in[n]; }
int func_2(local int *in, int n) { return func_3(in, n); }
int func_1(local int *in, int n) { return func_2(in, n); }
kernel void kernel_1(local int *in, global int *out, int n) {
  out[n] = func_1(in, n);
}
kernel void kernel_2(local int *in, global int *out, int n) {
  out[n] = func_1(in, n);
}

int foo(local int *in, int n) { return func_3(in, n); }
```

I can't explain the difference. For your example LLVM reports that both inner() and outer() have 2 uses before inlining and 1 afterwards. For my example all the uses reported make sense and the cleanup removes all function except the kernels.

dneto0: I can't explain those stray uses.  Weird.

dneto0:
Rebased, squashed, and pushed into master as f8661418a0ae

dneto0:
Whoops.  I had the wrong PR.

dneto0:
Rebased and pushed into master as af289ab156b

PR #181, opened 2018-08-30
Pass to scalarize composites

alan-baker:
* Currently only scalarizes struct type phis if an option (-hack-phis)
is enabled
 * Includes a test

The idea is that this pass can be extended as necessary to handle any scalarization cases that prove to necessary in the future.

### DIFF
--- a/lib/ScalarizePass.cpp
+++ b/lib/ScalarizePass.cpp
@@ -1,0 +1,124 @@
+// Copyright 2018 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


lib/ScalarizePass.cpp
```diff
@@ -0,0 +1,124 @@
...
+static RegisterPass<ScalarizePass>
+    X("Scalarize", "Scalarizes some instructions with composite returns");
+
+bool ScalarizePass::runOnModule(Module &M) {
+  bool Changed = false;
+  for (auto &F : M) {
+    for (auto &BB : F) {
+      for (auto &I : BB) {
+        if (auto *phi = dyn_cast<PHINode>(&I)) {
+          if (clspv::Option::HackPhis() && phi->getType()->isStructTy()) {
```
dneto0: Until this pass is generalized, you could save time by pull this option check to the outside all the loops.  The if the option is off you don't traverse anything.

PR #183, opened 2018-08-31
Generalize "Undo Truncated Switch" pass

dneto0:
Now it handles more cases, including select and phi.

Added a test.

### DIFF
--- a/lib/UndoTruncatedSwitchConditionPass.cpp
+++ b/lib/UndoTruncatedSwitchConditionPass.cpp
@@ -1,4 +1,4 @@
-// Copyright 2017 The Clspv Authors. All rights reserved.
+// Copyright 2018 The Clspv Authors. All rights reserved.
 //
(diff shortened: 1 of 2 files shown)


lib/UndoTruncatedSwitchConditionPass.cpp
```diff
@@ -66,30 +131,33 @@ bool UndoTruncatedSwitchConditionPass::runOnModule(Module &M) {
...
+      // And we replace the old value.
+      Cases.setValue(CI);
+    }
+  }
 
-      TI->eraseFromParent();
-    } else {
-      Cond->print(errs());
-      llvm_unreachable("Unhandled switch instruction condition!");
+  // Remove the zombies if we can.  We expect to.  They are ordered from
```
dneto0: This is not a perfect ordering for erasure, but it's good enough for the cases I've seen.

Example where it could fail: 
- phis that feed themselves, or mutally.
- the same narrow value feeding multiple switches.

PR #188, opened 2018-09-06
Extend RewriteInsertPass to handle complete vector chains.

alan-baker:
* Pass now handles complete chains of InsertElementInst
* updated tests

I didn't try to handle InsertElementInst in the same function as InsertValueInst since it seemed more trouble than it was worth to do the bookkeeping.

This fixes some customer issues I'm facing.

### DIFF
--- a/lib/RewriteInsertsPass.cpp
+++ b/lib/RewriteInsertsPass.cpp
@@ -154,6 +154,14 @@ class RewriteInsertsPass : public ModulePass {
    // containing the insertion instructions, in member order.
    // Otherwise returns nullptr.  Only handle insertions into vectors.
    InsertionVector *CompleteInsertionChain(InsertElementInst *ie) {
+     // Don't handle i8 vectors. Only <4 x i8> is supported and it is
+     // translated as i32.  Only handle single-index insertions.
+     if (auto vec_ty = dyn_cast<VectorType>(ie->getType())) {
+       if (vec_ty->getVectorElementType() == Type::getInt8Ty(ie->getContext())) {
+         return nullptr;
+       }
+     }
+
      // Only handle single-index insertions.
      if (ie->getNumOperands() == 3) {
        auto numElems = GetNumElements(ie->getType());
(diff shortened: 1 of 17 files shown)


lib/RewriteInsertsPass.cpp
```diff
@@ -323,7 +371,9 @@ bool RewriteInsertsPass::ReplacePartialInsertions(Module &M) {
     // Record the list of tails for the next round.
     InsertionVector NextRoundWorkList;
 
-    for (InsertValueInst *insertion : WorkList) {
+    //for (InsertValueInst *insertion : WorkList) {
```
dneto0: Delete dead commented-out code.

lib/RewriteInsertsPass.cpp
```diff
@@ -323,7 +371,9 @@ bool RewriteInsertsPass::ReplacePartialInsertions(Module &M) {
...
+    for (Instruction *inst : WorkList) {
+      InsertValueInst *insertion = cast<InsertValueInst>(inst);
```
dneto0: This cast will assert out if it's incorrect.  So why not keep the old more-specific type declaration in the range-for?

lib/RewriteInsertsPass.cpp
```diff
@@ -137,6 +149,56 @@ class RewriteInsertsPass : public ModulePass {
...
    }
 
+   // If this is the tail of a chain of InsertElementInst instructions
+   // that covers the entire vector, then return a small vector
+   // containing the insertion instructions, in member order.
+   // Otherwise returns nullptr.  Only handle insertions into vectors.
+   InsertionVector *CompleteInsertionChain(InsertElementInst *ie) {
+     // Don't handle i8 vectors. Only <4 x i8> is supported and it is
+     // translated as i32.  Only handle single-index insertions.
+     if (auto vec_ty = dyn_cast<VectorType>(ie->getType())) {
```
alan-baker: Since <4 x i8> is codegen'd as i32 in SPIRVProducerPass, I've disabled the transformation for vectors of i8 so that invalid SPIR-V is not produced.

dneto0:
Rebased, squashed, and pushed into master as abc935edb

PR #193, opened 2018-09-22
Accept -Ipath with no space between the option and path

### LINKED ISSUE #192: Passing -Ipath with no space doesn't work
clspv doesn't accept include search path options (`-Ipath`) without a space between `-I` and `path`.

kpet:
Fixes #192 


dneto0:
Rebased and pushed into master as 7e67a56d958

PR #195, opened 2018-09-22
Enable utils/fetch_sources.py to be called from anywhere

### LINKED ISSUE #194: utils/fetch_sources.py has to be called from the top directory
This is a bit of a pain and forced me to document the following instructions for clvk:

```
cd external/clspv
./utils/fetch_sources.py
cd ../..
```

I think it would be better for `utils/fetch_sources.py` to assume its location relative to the top of the directory structure.

kpet:
Fixes #194 

### DIFF
--- a/utils/fetch_sources.py
+++ b/utils/fetch_sources.py
@@ -26,9 +26,13 @@ import os.path
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


dneto0:
Rebased and pushed onto master as  a2add47e66c

PR #196, opened 2018-09-22
Support fast_ variants of the geometric built-in functions

kpet:
They pass the OpenCL conformance tests run with clvk on
Nvidia hardware.


dneto0:
Rebased and pushed into master as 7d09cec66a811

PR #197, opened 2018-09-22
Emit pod argument size to descriptor map

### LINKED ISSUE #185: Impossible to determine the size of the POD buffer from the descriptor map
When using `-cluster-pod-kernel-args`, it is not possible for an application to know how much memory to allocate for the POD buffer without prior knowledge of the kernels used. Unless I've missed something, the best an application can do without prior knowledge is to figure out the maximum offset within the buffer from the descriptor map and assume a maximum size for an argument.

kpet:
Enable users to:

- know how much memory to allocate for their pod buffer when
  using -cluster-pod-kernel-args

- check the size of the data bound to pod bindings is correct

Fixes #185 

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -162,9 +162,9 @@ Consider this example:
 It generates the following descriptor map:
 
     kernel,foo,arg,a,argOrdinal,0,descriptorSet,0,binding,0,offset,0,argKind,buffer
-    kernel,foo,arg,f,argOrdinal,1,descriptorSet,0,binding,1,offset,0,argKind,pod
+    kernel,foo,arg,f,argOrdinal,1,descriptorSet,0,binding,1,offset,0,argKind,pod,argSize,4
     kernel,foo,arg,b,argOrdinal,2,descriptorSet,0,binding,2,offset,0,argKind,buffer
-    kernel,foo,arg,c,argOrdinal,3,descriptorSet,0,binding,3,offset,0,argKind,pod
+    kernel,foo,arg,c,argOrdinal,3,descriptorSet,0,binding,3,offset,0,argKind,pod,argSize,4
 
 For kernel arguments of types pointer-to-global, pointer-to-constant, and
 plain-old-data types, the fields are:
@@ -189,6 +189,8 @@ plain-old-data types, the fields are:
   - `ro_image` - Read-only image
   - `wo_image` - Write-only image
   - `sampler` - Sampler
+- `argSize`
+- only present for plain-old-data kernel arguments.
 
 Consider this example, which uses pointer-to-local arguments:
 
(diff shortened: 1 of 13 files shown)


ericberdahl:
Is it feasible to only emit argSize annotations for pod arguments? They don't have a meaning for arguments that are not pod arguments.

ericberdahl:
I'm thinking that you are observing that pod and pod_ubo arguments are different from other argument kinds. This is directly analogous to the way that local arguments are different from other argument kinds. Thus, the argument kind field already provides a key to the consuming software for which fields may be included with the argument and which fields will not.

As a client, I would prefer that the fields included in an argument entry in the descriptor map be relevant to that argument. If a field is not relevant, it should not be present. This makes a stronger contract between the compiler and the consuming software.

kpet:
@dneto0 @alan-baker Any opinion on this change? Which way do you want to go? I really don't mind either way personally.

dneto0:
Sorry I'm late to this.  I agree with @ericberdahl that the argSize key/value pair should only be included for pod and pod_ubo fields.  I also agree with @kpet that in that case they key/value pair should appear after the argKind key/value pair.    That might help some consumers.

docs/OpenCLCOnVulkan.md
```diff
@@ -161,10 +161,10 @@ Consider this example:
 
 It generates the following descriptor map:
 
-    kernel,foo,arg,a,argOrdinal,0,descriptorSet,0,binding,0,offset,0,argKind,buffer
-    kernel,foo,arg,f,argOrdinal,1,descriptorSet,0,binding,1,offset,0,argKind,pod
-    kernel,foo,arg,b,argOrdinal,2,descriptorSet,0,binding,2,offset,0,argKind,buffer
-    kernel,foo,arg,c,argOrdinal,3,descriptorSet,0,binding,3,offset,0,argKind,pod
+    kernel,foo,arg,a,argOrdinal,0,descriptorSet,0,binding,0,offset,0,argSize,0,argKind,buffer
+    kernel,foo,arg,f,argOrdinal,1,descriptorSet,0,binding,1,offset,0,argSize,4,argKind,pod
```
dneto0: As per @ericberdahl 's commentary, I'd like this only on pod and pod_ubo.
And put the new argSize after  the argKind key/value pair.

ericberdahl:
From my perspective, the location of the argSize key/value pair relative to the argKind key/value pair is of less consequence. From a larger perspective having the argSize pair follow the argKind pair seems very reasonable (e.g. existing consumers may have existing, albeit unwise imo, implementations that presume the order of data for kernel argument lines).

kpet:
@dneto0 Thanks. I'll update the PR.

>existing consumers may have existing, albeit unwise imo, implementations that presume the order of data for kernel argument lines

@ericberdahl This is what I'm doing in clvk. There is nothing documented that states that order doesn't matter. As my previous comment said:
>This would be a bit simpler if we decided and documented that the format is just a serialisation of a set of key-value pairs, which seems to be true for all cases.

If we can all agree on that and document it, then I agree there would no longer be any good reason to expect tokens in a precise order.

docs/OpenCLCOnVulkan.md
kpet: Done.

kpet:
I've updated the PR to address review comments and added a test to cover the combination of `-pod-ubo` and `-cluster-pod-kernel-args` (there was none).

Since the preference is to only emit meaningful fields, I guess we may want to change the way `offset` is handled. I'll raise an issue to discuss that and more generally what constraints the format should obey to.

PR #198, opened 2018-09-22
Add tests for the step built-in function

kpet:
These tests cover the case where both arguments have the same
type which is the only one correctly implemented.

The OpenCL conformance tests for these cases pass on Nvidia hardware when
run through clvk.


dneto0:
Rebased and pushed into master as 397601cfc2c

PR #201, opened 2018-09-24
Remove 'local' kernel arguments from the list of restrictions

kpet:
This is definitely supported now.

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -495,9 +495,6 @@ variant of SPIR-V are restricted.
 
 OpenCL C language kernels **must not** be called from other kernels.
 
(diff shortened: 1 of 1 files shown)


ericberdahl:
I believe this is only supported if the underlying implementation supports the associated extensions. At a minimum, the feature that allows half* kernel arguments should be an opt-in feature protected by a command-line option.

kpet:
Which specific extensions did you have in mind?

>At a minimum, the feature that allows half* kernel arguments should be an opt-in feature protected by a command-line option.

It is documented as a separate restriction which I haven't touched but your comment suggest that it should probably be relaxed.

alan-baker:
Let's leave unrelated changes for separate issues, I agree pointer-to-local kernel args work.

PR #202, opened 2018-09-26
Make it possible to provide SPIRV-Tools and SPIRV-Headers externally

kpet:
The default behaviour doesn't change but a project embedding clspv
can now:
- pass SPIRV_HEADERS_SOURCE_DIR to provide SPIRV-Headers
- pass SPIRV_TOOLS_SOURCE_DIR to provide SPIRV-Tools
- pass a list of dependencies to be managed by utils/fetch_sources.py

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,8 +43,26 @@ endif()
 
 use_component(${CMAKE_CURRENT_SOURCE_DIR}/third_party/clang)
 use_component(${CMAKE_CURRENT_SOURCE_DIR}/third_party/llvm)
-use_component(${CMAKE_CURRENT_SOURCE_DIR}/third_party/SPIRV-Headers)
-use_component(${CMAKE_CURRENT_SOURCE_DIR}/third_party/SPIRV-Tools)
+
+if (NOT DEFINED SPIRV_HEADERS_SOURCE_DIR)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -43,8 +43,26 @@ endif()
...
+  use_component(${SPIRV_TOOLS_SOURCE_DIR})
+
+  # First tell SPIR-V Tools where to find SPIR-V Headers
+  set(SPIRV-Headers_SOURCE_DIR ${SPIRV_HEADERS_SOURCE_DIR})
+
+  # Bring in the SPIR-V Tools repository which we'll use for testing
+  add_subdirectory(${SPIRV_TOOLS_SOURCE_DIR} EXCLUDE_FROM_ALL)
+endif()
+
+set(SPIRV_TOOLS_BINARY_DIR "$<TARGET_FILE_DIR:spirv-dis>")
```
dneto0: I found an issue with this line.
Scenario is:  
- We specify SPIRV_TOOLS_SOURCE_DIR via a cmake -D  option, to point elsewhere
- The target doesn't exist, and Cmake errors out.

Eventually I'd like to handle this case gracefully.

I think this doesn't degrade the current case, and you a use case that is an enhancement.  So I'll merge this now.

PR #204, opened 2018-09-28
Support building out-of-tree SPIRV-{Headers,Tools}

### LINKED ISSUE #203: support out-of-tree SPIRV-Tools and SPIRV-Headers where the spirv-dis target doesn't exist yet
For example, 

With this kind of command line:
```
cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DSPIRV_TOOLS_SOURCE_DIR=/foosrc/third_party/SPIRV-Tools -DSPIRV_HEADERS_SOURCE_DIR=/foosrc/third_party/SPIRV-Headers ../clspv
```

Currently, I get an error like this:

```
CMake Error at test/CMakeLists.txt:17 (add_custom_target):
  Error evaluating generator expression:

    $<TARGET_FILE_DIR:spirv-dis>

  No target "spirv-dis"
```

In this case we then add the SPIRV-Tools project from that other location?
See comment on #202

kpet:
Fixes #203.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,16 +50,17 @@ if (NOT DEFINED SPIRV_HEADERS_SOURCE_DIR)
 endif()
 
 if (NOT DEFINED SPIRV_TOOLS_SOURCE_DIR)
-
   set(SPIRV_TOOLS_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/third_party/SPIRV-Tools)
-
   use_component(${SPIRV_TOOLS_SOURCE_DIR})
+endif()
 
+if (NOT TARGET spirv-dis)
   # First tell SPIR-V Tools where to find SPIR-V Headers
   set(SPIRV-Headers_SOURCE_DIR ${SPIRV_HEADERS_SOURCE_DIR})
 
   # Bring in the SPIR-V Tools repository which we'll use for testing
-  add_subdirectory(${SPIRV_TOOLS_SOURCE_DIR} EXCLUDE_FROM_ALL)
+  add_subdirectory(${SPIRV_TOOLS_SOURCE_DIR}
+      ${CMAKE_CURRENT_BINARY_DIR}/third_party/SPIRV-Tools EXCLUDE_FROM_ALL)
 endif()
 
 set(SPIRV_TOOLS_BINARY_DIR "$<TARGET_FILE_DIR:spirv-dis>")


dneto0:
Hey @dnovillo please take a look at this one.  I expect it's good but don't have the cycles to confirm.

dnovillo:
> Hey @dnovillo please take a look at this one. I expect it's good but don't have the cycles to confirm.

Sure.

CMakeLists.txt
```diff
@@ -50,16 +50,17 @@ if (NOT DEFINED SPIRV_HEADERS_SOURCE_DIR)
 endif()
 
 if (NOT DEFINED SPIRV_TOOLS_SOURCE_DIR)
-
   set(SPIRV_TOOLS_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/third_party/SPIRV-Tools)
-
   use_component(${SPIRV_TOOLS_SOURCE_DIR})
+endif()
 
+if (NOT TARGET spirv-dis)
```
dnovillo: This will always be true until we add the SPIRV-Tools subdirectory.  I would just take it away.

kpet: This won't be true for projects that embed clspv and have already added SPIRV-Tools as clvk does :). You may want to read #202 and #203 for the rationale. Currently supported are the following cases:
- a project embeds clspv and SPIRV-Tools and provides both `SPIRV_TOOLS_SOURCE_DIR` and the target `spirv-dis` since SPIRV-Tools has already been added. Both conditions are false.
- a project embeds clspv and doesn't provide SPIRV-Tools. clspv uses its own version. `SPIRV_TOOLS_SOURCE_DIR` isn't defined so we set the variable to the stock path in third_party/ and the target `spirv-dis` doesn't exist so we add SPIRV-Tools as a sub-directory.

The case that is currently not handled is the one that arises when the user provides `SPIRV_TOOLS_SOURCE_DIR` but wants the build to be managed by clspv's build system. In this case we just want to add SPIRV-Tools using the path provided by the user.

dnovillo: > This won't be true for projects that embed clspv and have already added SPIRV-Tools as clvk does :). You may want to read #202 and #203 for the rationale. Currently supported are the following cases:

Ah, quite right.  I had missed that.  Thanks.  In that case, it LGTM.

kpet:
@dnovillo Anything you need me to do before merging this one?

dnovillo:
> @dnovillo Anything you need me to do before merging this one?

No.  It's good as-is and I approved it a few days ago.  You can submit it now.  Or do you need me to?

kpet:
@dnovillo  Thanks!

PR #210, opened 2018-10-02
Fix codegen for logical operations on bool vectors

### LINKED ISSUE #209: Invalid SPIR-V generated for logical operations on vectors of booleans
The following kernel exposes an existing codegen issue around vectors of booleans:

```
void kernel foo(float4 a, float4 b, float4 c, float4 d, global int4 *o)
{
    int4 ab = (a <= b);
    int4 cd = (c > d);
    *o = (ab && cd);
}
```

An `OpBitwiseAnd` is emitted instead of an `OpLogicalAnd`. This was flagged by the validator (`Expected int scalar or vector type as Result Type: BitwiseAnd`).

PR coming.

kpet:
Bitwise instructions were being emitted. The tests added only cover `And` and `Or` as I couldn't find a way of producing `Xor` on boolean vectors which seems to line up with a few observations from the OpenCL specification:

- there is no xor logical operator (although `^^` has been reserved for that use)
- boolean vectors aren't explicitly specified (even though `booln` (2,3,4,8 and 16) is a reserved type)
- the result of relational operators applied to two vectors is guaranteed to be a vector of signed integers (this is the case that made me discover the issue)

Clang's omission of the sign extension to integer vectors when the result of multiple relational operators is only used by a logical operation is an optimisation. Since there is no logical xor operator it is reasonable for Clang to always sign extend the result of relational operators (according to §6.3d of the OpenCL 1.2 specification) that feed into an xor operator and perform the operation as bitwise.

An optimisation similar to that done for `And` and `Or` could be made by clspv in cases where the results of relational operators are only used by xor operators but should IMHO be discussed separately from this change.

Also tidy up a few related things:

- use `isIntOrIntVectorTy(1)` where it makes the code clearer
- tighten the check in `SPIRVProducerPass::GenerateInstruction` that special cases `Instruction::Xor` for booleans. It was allowing `OpLogicalNot` to be emitted in all cases where one of the operands was constant, which is correct only when the constrant is 1. The value of the constant is now checked.

Fixes #209 

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -793,9 +793,7 @@ void SPIRVProducerPass::GenerateLLVMIRInfo(Module &M, const DataLayout &DL) {
 
           auto OpTy = I.getOperand(0)->getType();
 
-          if (OpTy->isIntegerTy(1) ||
-              (OpTy->isVectorTy() &&
-               OpTy->getVectorElementType()->isIntegerTy(1))) {
+          if (OpTy->isIntOrIntVectorTy(1)) {
             if (I.getOpcode() == Instruction::ZExt) {
               APInt One(32, 1);
               FindConstant(Constant::getNullValue(I.getType()));
@@ -866,9 +864,7 @@ void SPIRVProducerPass::GenerateLLVMIRInfo(Module &M, const DataLayout &DL) {
 
           auto OpTy = I.getOperand(0)->getType();
 
-          if (OpTy->isIntegerTy(1) ||
-              (OpTy->isVectorTy() &&
-               OpTy->getVectorElementType()->isIntegerTy(1))) {
+          if (OpTy->isIntOrIntVectorTy(1)) {
             if (I.getOpcode() == Instruction::ZExt) {
               APInt One(32, 1);
               FindConstant(Constant::getNullValue(I.getType()));
@@ -3480,7 +3476,7 @@ spv::Op SPIRVProducerPass::GetSPIRVCastOpcode(Instruction &I) {
 }
 
 spv::Op SPIRVProducerPass::GetSPIRVBinaryOpcode(Instruction &I) {
-  if (I.getType()->isIntegerTy(1)) {
+  if (I.getType()->isIntOrIntVectorTy(1)) {
     switch (I.getOpcode()) {
     default:
       break;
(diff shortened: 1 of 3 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -1613,7 +1609,7 @@ void SPIRVProducerPass::FindConstantPerFunc(Function &F) {
         }
 
         continue;
-      } else if ((Instruction::Xor == I.getOpcode()) && I.getType()->isIntegerTy(1)) {
+      } else if ((Instruction::Xor == I.getOpcode()) && I.getType()->isIntOrIntVectorTy(1)) {
```
alan-baker: Within this if we iterate through the operands of `I`. If any operand is a constant it is a assumed to be a ConstantInt. That `cast` seems like it would be invalid in that case.

kpet:
Looking at this a bit more, I've found another problem around `Xor` and booleans. `SPIRVProducerPass::GenerateInstruction` has got a special case for `Instruction::Xor` on booleans with at least one constant operand; it always produces `OpLogicalNot` using the operand that is not constant. This only works when the constant is 1, which happens to be the only thing covered by the tests (ifelse.cl, select_between_constant_load.cl, for.cl and others) and the only thing I've managed to convince Clang to produce so far, even with `-O0`.

At this point I'm tempted to suggest that one option (and I'm happy to be called lazy for that proposal :)) is to delete all the broken special handling code for `Xor` on booleans (the code managing constants and the code that emits `OpLogicalNot`), rely on the default case that always produces `OpLogicalNotEqual` and count on consumers to optimise `OpLogicalNotEqual` with a constant operand which seems like a pretty basic optimisation. Fixing the vector case would then becomes trivial and the only question left would be that of testing `Xor` on vectors of booleans (which I still need to spend more time on).

lib/SPIRVProducerPass.cpp
kpet: I've reverted this change. This handling of constants only exists to accompany the special case for `Instruction::Xor` in `SPIRVProducerPass::GenerateInstruction`. Since that can't meaningfully be updated to deal with vectors as well, the corresponding handling of constants shouldn't be changed.

kpet:
@alan-baker  I've updated the PR. All the rationale for what I've done is in the PR description/commit message. There was no need to consider the lazy option in the end :). I believe this is ready for another round of review.

Issue #208, opened 2018-10-02
Add support to generate pointer-to-constant kernel args in the Uniform storage class


alan-baker:
This should be implemented in multiple steps.
 1. Assuming exhaustive inlining (with appropriate errors otherwise).
 1. Support within DRA (i.e. must be fully DRA'd)

Take a new argument to clspv: `-constant-args-ubo`. Any pointer-to-constant kernel arg should be generated in the Uniform storage class with the Block decoration.  If the option is enable `Constant` addresspaces should be codegen'd in the `Uniform` storage class.

Needs error checking for layout rules ([Vulkan spec 14.5.4](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#interfaces-resources)). Assumption is the author will provide an appropriate struct definition in the source. The checking is complicated by the fact that Clang will insert padding to satisfy alignment requirements. For example:
```
struct data_type {
  int x;
  int y __attribute__((aligned(16));
};
```
Will translate into LLVM as:
```
%struct.data_type = type { i32, [12 x i8], i32, [12 x i8] }
```

Layout validation should only check used members and translation should codegen unused members as `OpTypeInt 32 0` so other indexing is unchanged. Will not support loading the whole struct initially.

alan-baker:
Add tests for uchar4 since that is codegen'd so strangely.

alan-baker:
This is done enough to close this issue. Future enhancements can be filed separately.

PR #214, opened 2018-10-09
Pointer-to-constant kernel args in uniform buffers

alan-baker:
Contributes to #208 

Initial implementation of support for pointer-to-constant kernel arguments in uniform buffers.

If the new option (`-constant-args-ubo`) is enabled, pointer-to-constant kernel arguments will be generated as variables in the Uniform storage class. A new ArgKind, `BufferUBO`, has been added to indicate this change. Standard uniform buffer layout rules from Vulkan specification section 14.5.4 are checked in the AST.

Limitations
 * Full inlining must be enabled
  * this will be somewhat relaxed in the future for cases that DRA can handle
 * Extra layout restrictions
  * All types must have a size that is a multiple of their alignment
  * structures may not have implicit padding
  * Both of these restrictions should be relaxed in the future, but require significant changes to support.

I added a new command line option, -verify, which enables diagnostic verification to support testing the new AST based error checks.


alan-baker:
Closing this in favour of a more comprehensive PR.

PR #215, opened 2018-10-13
Add a script to help with creating test cases

kpet:
One can just write the OpenCL C source and let the script build
a SPIR-V module, disassemble it, postprocess the disassembly
to introduce FileCheck variables and check directives (using DAG
directives where possible), add a standard set of RUN directives,
and print the final test case file, ready to be dropped in test/.

Usage (assuming clspv is in PATH):

./utils/make_test.py source.cl > test/my_new_shiny_test_case.cl

### DIFF
--- a/utils/make_test.py
+++ b/utils/make_test.py
@@ -1,0 +1,195 @@
+#!/usr/bin/env python
+
+import argparse
+import collections
+import os
+import re
+import subprocess
+import tempfile
+
+DESCRIPTION = """This script can be used to help with writing test cases for clspv.
+
+When passed an OpenCL C source file, it will:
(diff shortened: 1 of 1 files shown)


utils/make_test.py
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env python
+
```
dneto0: Needs a copyright statement please.

dneto0:
Rebased and pushed into master as 3caca6b61dac

utils/make_test.py
kpet: Will create a PR.

PR #224, opened 2018-10-22
pointer-to-constant args in Uniform buffer

alan-baker:
This adds support for pointer-to-constant args in Uniform buffers.

Contributes to #208 

New option: `-constant-args-ubo` tells clspv to put _all_ pointer-to-constant kernel args in an Uniform buffer. Compiler errors if the argument does not satisfy std140 layouts.

Major changes:
 * New frontend checks that the input conforms to standard Uniform buffer layout rules (Vulkan 14.5.4)
  * error if non-compliant
 * New ArgKind, `BufferUBO`, to represent difference
 * New transform `UBOTypeTransformPass` fixes up LLVM struct padding to work with uniform layout rules
  * Passes type transform information forward in the flow via module scope metadata maps
  * SPIRVProducerPass now uses wrapper methods around DataLayout accessors to ensure the right type information is used
 * Tested with module scope __constant variables, pointer-to-local args, etc.

Restrictions:
 * All or nothing: can't fallback on SSBO currently
 * Requires inlining currently
  * future improvement to relax this requirement
 * non-multiple of 16 byte arrays are not supported
  * 14.5.4 rules about where the next element goes requires significant changes in the llvm representation to make this work

### DIFF
--- a/lib/UBOTypeTransformPass.cpp
+++ b/lib/UBOTypeTransformPass.cpp
@@ -33,38 +33,66 @@ using namespace llvm;
 namespace {
 
 class UBOTypeTransformPass final : public ModulePass {
- public:
+public:
   static char ID;
   UBOTypeTransformPass() : ModulePass(ID) {}
   bool runOnModule(Module &M) override;
- private:
-  Type *MapType(Type* type, Module &M);
+
+private:
+  // Returns the remapped version of |type| that satisfies UBO requirements.
+  Type *MapType(Type *type, Module &M);
+
+  // Returns the remapped version of |type| that satisfies UBO requirements.
   StructType *MapStructType(StructType *struct_ty, Module &M);
+
+  // Performs type mutation on |M|. Returns true if |M| was modified.
   bool RemapTypes(Module &M);
+
+  // Performs type mutation for functions that require it. Returns true if the
+  // module is modified.
+  //
+  // If a function requires type mutation it will be replaced by a new
+  // function. The function's basic blocks are moved into the new function and
+  // all metadata is copied.
+  bool RemapFunctions(SmallVectorImpl<Function *> *functions_to_modify,
+                      Module &M);
+
+  // Performs type mutation on |user|. Recursively fixes operands of |user|.
+  // Returns true if the module is modified.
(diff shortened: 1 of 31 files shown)


lib/UBOTypeTransformPass.cpp
```diff
@@ -0,0 +1,520 @@
...
+                                                Module &M) {
+  bool changed = false;
+  SmallVector<Type *, 8> elements;
+  SmallVector<uint64_t, 8> offsets;
+  const auto *layout = M.getDataLayout().getStructLayout(struct_ty);
+  for (unsigned i = 0; i != struct_ty->getNumElements(); ++i) {
+    Type *element = struct_ty->getElementType(i);
+    uint64_t offset = layout->getElementOffset(i);
+    const auto *array = dyn_cast<ArrayType>(element);
+    if (array && array->getElementType()->isIntegerTy(8) && offset % 16 != 0) {
```
dneto0: The 16 is a bit magic here, but I'll allow it.
It's from the UBO layout requirements.

lib/UBOTypeTransformPass.cpp
```diff
@@ -0,0 +1,520 @@
...
+    Constant *null_constant = Constant::getNullValue(remapped_ty);
+    remapped_globals_[constant] = null_constant;
+    return null_constant;
+  } else if (isa<UndefValue>(constant)) {
+    // This case should catch the padding transformations since the padding
+    // can't be initialized.
+    Constant *undef_constant = UndefValue::get(remapped_ty);
+    remapped_globals_[constant] = undef_constant;
+    return undef_constant;
+  } else if (auto *agg_constant = dyn_cast<ConstantAggregate>(constant)) {
```
dneto0: I wonder if you'll also need a ConstantDataSequential case. That only inherits from ConstantData

I guess we'll find out if you hit the llvm_unreachable

dneto0: ok for now since this is only turned on under flag.

dneto0:
Rebased, squashed, and pushed into master as fcda9485b

PR #230, opened 2018-10-28
Add support for the bitselect built-in function 

kpet:
This adds support for the bitselect built-in function as well as support for printing assembly for 64-bit integer literals that was needed for the tests.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -78,6 +78,7 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
   bool replaceIsInfAndIsNan(Module &M);
   bool replaceAllAndAny(Module &M);
   bool replaceSelect(Module &M);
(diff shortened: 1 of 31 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -888,6 +890,137 @@ bool ReplaceOpenCLBuiltinPass::replaceSelect(Module &M) {
...
+          } else {
+            // The operands have an integer type, use it directly
+            BitType = OpType;
+          }
+
+          // All the operands are now always integers
+          // implement as (c & b) | (~c & a)
+
+          // Create our negated predicate value
+          auto AllOnes = ConstantInt::getSigned(BitType->getScalarType(), -1);
```
alan-baker: These lines could be replaced with:
```
auto AllOnes = Constant::getAllOnesValue(BitType);
```

kpet: Thanks, hadn't spotted that one.

PR #233, opened 2018-10-30
Add an option for a stronger basic block ordering

alan-baker:
* Adds -hack-block-order to produce blocks in structured order to
satisfy some drivers

Did not change the default ordering because this is slower, the default ordering should be accepted by drivers and there would be too many test updates.

### DIFF
--- a/lib/ReorderBasicBlocksPass.cpp
+++ b/lib/ReorderBasicBlocksPass.cpp
@@ -40,6 +40,17 @@ struct ReorderBasicBlocksPass : public FunctionPass {
 
 private:
   // Produce structured sorting from |block| into |order|.
+  //
+  // This order has the following properties:
+  // * dominators come before all blocks they dominate
+  // * a merge block follows all blocks that are in control constructs of the
+  // associated header
+  // * blocks in a loop construct precede the blocks in the associated continue
+  // construct
+  // * no interleaving; blocks in one construct A are mixed with blocks in
+  // another construct B only if one is nested inside the other (without loss
+  // of generality, A's header dominates all blocks in B, and all B's blocks
+  // appear contiguously within A's blocks)
   void StructuredOrder(BasicBlock *block, const LoopInfo &LI,
                        std::deque<BasicBlock *> *order,
                        DenseSet<BasicBlock *> *visited);
@@ -72,8 +83,15 @@ void ReorderBasicBlocksPass::StructuredOrder(BasicBlock *block,
     merge_block = loop->getExitBlock();
 
     if (loop->isLoopLatch(block)) {
+      // The header is also the latch (i.e. a single block loop).
       continue_block = block;
     } else {
+      // The continue block satisfies the following conditions:
+      // 1. Is dominated by the header (true by construction at this point).
+      // 2. Dominates the latch block.
+      // We can assume the loop has multiple blocks since the single block loop
+      // is handled above. By construction the header always dominates the
+      // latch block, so we exclude that case specifically.
       DominatorTree &DT = getAnalysis<DominatorTreeWrapperPass>().getDomTree();
       auto *header = loop->getHeader();
       auto *latch = loop->getLoopLatch();
@@ -81,12 +99,17 @@ void ReorderBasicBlocksPass::StructuredOrder(BasicBlock *block,
         if (bb == header)
           continue;
 
+        // Several block might dominate the latch, we can pick any.
         if (DT.dominates(bb, latch))
           continue_block = bb;
       }
     }
+    // At least the latch dominates itself so we will always find a continue block.
     assert(continue_block && merge_block && "Bad loop");
   } else if (terminator->getNumSuccessors() > 1) {
+    // This is potentially a selection case, but it could also be a conditional
+    // branch with one arm back to the loop header, which would make this block
+    // the latch block.
     bool has_back_edge = false;
 
     for (unsigned i = 0; i < terminator->getNumSuccessors(); ++i) {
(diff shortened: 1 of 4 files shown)


lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+
+  // Identify the merge and continue blocks for special treatment.
+  const auto *terminator = dyn_cast<TerminatorInst>(block->getTerminator());
+  BasicBlock *continue_block = nullptr;
+  BasicBlock *merge_block = nullptr;
+  if (LI.isLoopHeader(block)) {
+    Loop *loop = LI.getLoopFor(block);
+    merge_block = loop->getExitBlock();
+
+    if (loop->isLoopLatch(block)) {
```
dneto0: Needs some comments about what's going on.

My understanding is:   the "then" case occurs when the block (which is a loop header) is its own latch.  So that's the  case of a single-block loop.  Please add a comment to that effect.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+      continue_block = block;
+    } else {
+      DominatorTree &DT = getAnalysis<DominatorTreeWrapperPass>().getDomTree();
```
dneto0: // This "else" case is for when the loop has several blocks.

// The continue block satisfies the following conditions:
// 1. Is dominated by the header. (That's true by construction at this point.)
// 2. Dominates the latch block.
// We can assume the loop has multiple blocks, since that single block loop case was handled separately.  
// By construction, the header always dominate the latch block, so we have to exclude that case specifically.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+      auto *header = loop->getHeader();
+      auto *latch = loop->getLoopLatch();
+      for (auto *bb : loop->blocks()) {
+        if (bb == header)
+          continue;
+
+        if (DT.dominates(bb, latch))
+          continue_block = bb;
```
dneto0: // Several blocks might dominate the latch block.  It doesn't matter which one we pick.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+      }
+    }
+    assert(continue_block && merge_block && "Bad loop");
```
dneto0: // In the worst case, the latch block dominates itself, so continue_block will always be set.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+  } else if (terminator->getNumSuccessors() > 1) {
+    bool has_back_edge = false;
```
dneto0: // This is potentially a selection case.  But it could also be a conditional branch with one arm pointing back to the loop header, which would make this block the latch block for the loop.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+    if (continue_block)
+      StructuredOrder(continue_block, LI, order, visited);
+  }
+  for (unsigned i = 0; i < terminator->getNumSuccessors(); ++i) {
+    auto *successor = terminator->getSuccessor(i);
+    if (successor != merge_block && successor != continue_block)
+      StructuredOrder(successor, LI, order, visited);
+  }
+
+  order->push_front(block);
```
dneto0: Hm.  A deque that is only added to the front.  _If_ this were performance-critical I'd push onto the back of a vector and reverse later.  But this is not perf-critical.

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -27,11 +33,18 @@ struct ReorderBasicBlocksPass : public FunctionPass {
...
   void getAnalysisUsage(AnalysisUsage &AU) const override {
     AU.addRequired<DominatorTreeWrapperPass>();
+    AU.addRequired<LoopInfoWrapperPass>();
   }
 
   bool runOnFunction(Function &F) override;
+
+private:
+  // Produce structured sorting from |block| into |order|.
+  void StructuredOrder(BasicBlock *block, const LoopInfo &LI,
```
dneto0: Should say what structured order is, i.e. what properties it has (not how you compute it).  And maybe why it's nice.

cfg.h from SPIRV-Tools says:


    // This order has the property that dominators come before all blocks they
    // dominate and merge blocks come after all blocks that are in the control
    // constructs of their header.

It should also add the properties of the continue construct as well.... :-)
I think that would be
    // ... and the continue construct blocks appear after the blocks in the associated loop construct.

So:
    // This order has the following properties:
    // -  dominators come before all blocks they dominate
    // - a merge blocks follows all blocks that are in the control
    //   constructs of the associated header.
    // - blocks in a loop construct precede the blocks in the associated continue construct
    // - no interleaving: blocks in one construct A are mixed with blocks in another construct B only if one is nested inside the other (without loss of generality, A's header dominates all blocks in B, and all B's blocks appear contiguously within A's blocks.)

lib/ReorderBasicBlocksPass.cpp
```diff
@@ -41,21 +54,94 @@ namespace clspv {
...
+  } else if (terminator->getNumSuccessors() > 1) {
+    bool has_back_edge = false;
+
+    for (unsigned i = 0; i < terminator->getNumSuccessors(); ++i) {
+      if (LI.isLoopHeader(terminator->getSuccessor(i)))
+        has_back_edge = true;
+    }
+
+    if (!has_back_edge) {
+      merge_block = terminator->getSuccessor(1);
```
dneto0: This is really subtle.  This is a strong dependence on the behaviour of the structurizer.
It says


    /// Control flow is expressed as a branch where the true exit goes into the
    /// "Then"/"Else" region, while the false exit skips the region

That's why you select the else branch here.  Needs a comment saying so (!)

Issue #242, opened 2018-11-08
Add C++ API to clspv


alan-baker:
I will be adding a C++ API to clspv and doing some refactoring to make this work (e.g. #236). My plan is to add functionality to satisfy an OpenCL-like API (for clvk) and an API for our testing requirements. For example, I will add a function to compile from a source string. It will populate a binary container (`vector<uint32_t>`) and descriptor map container.

The descriptor map will turn into a simple data structure that is populated by clspv. I will add an ostream operator to print it to a file for command line usage. The data structure will look something like:
```
struct DescriptorMapEntry {
  enum Kind {
    Sampler,
    Kernel,
    Constant
  } kind;

  uint32_t descriptor_set;
  uint32_t binding;

  struct SamplerData {
   uint32_t mask;
   // Removed expression. Can rebuild on output.
   //std::string expression;
  } sampler_data;

 struct KernelArgData {
   std::string kernel_name;
   std::string arg_name;
   uint32_t arg_ordinal;
   ArgKind arg_kind;

   // Pointer-to-local data.
   uint32_t local_spec_id;
   uint32_t local_elem_size;

   // POD data.
   uint32_t offset;
   uint32_t arg_size;
 } kernel_arg_data;

  struct ConstantData {
    ArgKind kind;
    // Changed definition
    //std::string hex_bytes;
    std::vector<uint8_t> hex_bytes;
  } constant_data;

   DescriptorMapEntry(SamplerData &&data, uint32_t descriptor_set, uint32_t binding);
   DescriptorMapEntry(KernelArgData &&data, uint32_t descriptor_set, uint32_t binding);
   DescriptorMapEntry(ConstantData &&data, uint32_t descriptor_set, uint32_t binding);
};
```

The logging and error reporting will also change a bit. I haven't finalized on what it will look like, but there will likely be a version of the API that returns a simple result struct that contains an exit code and any messages as a string.

FYI @kpet, @ericberdahl, @dneto0. I would appreciate your comments and feedback on this proposal. I am developing this in parallel with the necessary changes to clvk to ensure it works seamlessly.

dneto0:
The sampler mask values are defined by the OpenCL bits for the various parts (addressing mode, etc.)  Need to document where we take those numbers from. (I think it's the CLK_* enums.)

The sampler string expression is purely informational since it repeats exactly the info in the mask.  I think we also canonicalize it.

[...]

Issue #186, opened 2018-09-02
clspv doesn't exit with an error code when a kernel uses an unsupported builtin


kpet:
Instead, it generates invalid SPIR-V that is usually caught by the validator.

Example `builtin.cl` source:
```
kernel void test_kernel(global float *in, global float *out)
{
    float3 a = vload3(0, in);
    vstore3(a, 0, out);
}
```
Compiling with `clspv -o builtin.spv builtin.cl` does produce a helpful error message:
```
Can't translate function call.  Missing builtin? _Z7vstore3Dv3_fjPU3AS1f in:   tail call spir_func void @_Z7vstore3Dv3_fjPU3AS1f(<3 x float> %call, i32 0, float addrspace(1)* %4) #2
Can't translate function call.  Missing builtin? _Z6vload3jPU3AS1Kf in:   %call = tail call spir_func <3 x float> @_Z6vload3jPU3AS1Kf(i32 0, float addrspace(1)* %2) #2
```
But the exit code is zero.

dneto0:
Hm.  Seems I already visited this before, but didn't finish:

          if (CalleeID == 0) {
            errs() << "Can't translate function call.  Missing builtin? "
                   << callee_name << " in: " << *Call << "\n";
            // TODO(dneto): Can we error out?  Enabling this llvm_unreachable
            // causes an infinite loop.  Instead, go ahead and generate
            // the bad function call.  A validator will catch the 0-Id.
            // llvm_unreachable("Can't translate function call");
          }

PR #244, opened 2018-11-12
Add API to compile from source string

alan-baker:
New API analogous to clBuildProgram that compiles from source strings.

Contributes to #242

* Refactored ArgKind to expose the enum externally
 * Formalized PodUBO as an ArgKind, but plumbing is minimal still
* Added an initial version of data structure for descriptor map entry
 * ostream operator writes current desriptor map
* new api populates SPIR-V and descriptor map in memory
 * SPIRVProducerPass reworked to populate the descriptor map data structure

Tested against unit tests and clvk (see https://github.com/alan-baker/clvk/tree/use-clspv-api)


ericberdahl:
I'm becoming confused about the purpose and structure of this project. clspv self-identifies as a prototype compiler for a subset of OpenCL C to Vulkan compute shaders. That's very different than providing a library or API for applications to use to compile and build OpenCL C programs as Vulkan compute shaders. I'm concerned that the purpose, goal, and direction of clspv is being subtly changed in ways that will eventually prove to be conflicting.

The goal of providing APIs and libraries that allow applications to more easily use and consume the output artifacts of clspv's compilation is a good goal. The goal of providing APIs and libraries that allow applications to more easily use clspv in a manner similar to the way they use existing OpenCL APIs is a good goal. Neither of those goals are tightly aligned with the goal of creating a tool that compiles a subset of OpenCL C to Vulkan compute shaders.

kpet:
I've pulled this locally and tried to run the OpenCL conformance tests. Out of give or take 400 tests usually passing for me in my quick set, only 19 pass with this change and your clvk branch. Most of the test binaries just crash before reporting results. The first two I've looked at crash with the same message:

```
clspv: for the -simplifycfg-sink-common option: may only occur zero or one times!
```

This confirms my fear: crashing or terminating the process will no longer be acceptable. A quick grep shows that there are at least 75 calls to `llvm_unreachable` that will need to be taken care of (and a few more asserts). Asking for these to be turned into controlled errors in one go just isn't realistic so I suggest we keep the compilation API as a build option in clvk and work through these over time. This leads me to the next question: have you given any thought as to how you want to handle errors in passes? I've tried to play with some of the error handling features of llvm (`install_fatal_error_handler` + `report_fatal_error`) while trying to solve #186 but this becomes a bit more involved when everything is running in the same process.

alan-baker:
Ok, I've been testing with the simple tests so far. I will look into invoking a compile multiple time in a process and track down those errors.

include/clspv/Compiler.h
```diff
@@ -12,7 +12,37 @@
...
+#ifndef CLSPV_INCLUDE_CLSPV_COMPILER_H_
+#define CLSPV_INCLUDE_CLSPV_COMPILER_H_
+
+#include <string>
+#include <vector>
+
+#include "clspv/DescriptorMap.h"
+
 namespace clspv {
+struct Result {
```
kpet: Have you considered using `StringError` and friends?

include/clspv/DescriptorMap.h
```diff
@@ -0,0 +1,91 @@
...
+    uint32_t local_element_size;
+
+    // POD data.
+    uint32_t pod_offset;
+    uint32_t pod_arg_size;
+  } kernel_arg_data;
+
+  struct ConstantData {
+    ArgKind constant_kind;
+    std::vector<uint8_t> hex_bytes;
```
kpet: Probably worth renaming `hex_bytes`.

lib/Compiler.cpp
```diff
@@ -151,176 +157,192 @@ static llvm::cl::opt<bool> verify("verify", llvm::cl::init(false),
 
 // Populates |SamplerMapEntries| with data from the input sampler map. Returns 0
 // if successful.
-int ParseSamplerMap(llvm::SmallVectorImpl<std::pair<unsigned, std::string>> *SamplerMapEntries) {
-  if (!SamplerMap.empty()) {
+int ParseSamplerMap(
+    const std::string &sampler_map,
+    llvm::SmallVectorImpl<std::pair<unsigned, std::string>> *SamplerMapEntries,
+    llvm::raw_ostream *error_str) {
```
kpet: Passing by it by reference would make it easier to use.

alan-baker: [Google style guide](https://google.github.io/styleguide/cppguide.html) says all references must be const-qualified.

kpet:
@alan-baker That's made a big difference already. 140 tests passing for me. The binaries now seem to be crashing mostly in the passes.

include/clspv/Compiler.h
alan-baker: Right now the API doesn't expose LLVM headers and I'd prefer to keep it that way.

include/clspv/DescriptorMap.h
alan-baker: Good suggestion. I've renamed it to initializer.

PR #253, opened 2018-12-07
Move clspv_opencl_baked_header dependency to where it's needed

dneto0:


badleeazam:
⁣Get BlueMail for Android ​

PR #256, opened 2019-01-11
Generate UBOs as constant sized arrays

### LINKED ISSUE #243: Pointer-to-constant args in UBOs shouldn't use runtime arrays
According to Vulkan SPIR-V environment, OpTypeRuntimeArray can only be used in a StorageBuffer annotated as a Block or a Uniform annotated as a BufferBlock. Currently -constant-args-ubo uses a runtime array as a Block annotated Uniform. The same solution used for used for pointer-to-local arguments could be used here: specify a specialization constant to use as the size of the array.

alan-baker:
Fixes #243 

* UBO args are now generated as constant sized arrays instead of runtime arrays
 * The size is specifiable via the command line option `-max-ubo-size` (default is 64kB)
* Improved error messaging for UBO layout restrictions
 * Additional, note message is generated at the specific location an error is generated if that location is different than the parameter location
* New check that ArrayStride is a multiple of array alignment
 * Optionally disabled by `-relaxed-ubo-layout`
* All pointer-to-constant arguments are now validated as array types to properly match how they are generated
* Updated tests and added some new tests

Example updated messaging:
```
bad_struct.cl:12:19: error: in an UBO, structs must be aligned to their largest element alignment, rounded up to a multiple of 16 bytes
__kernel void foo(__constant struct outer* c) { }
                  ^
bad_struct.cl:9:3: note: here
  struct inner y;
  ^
```


alan-baker:
Merged as 3d9e20152fea932e7d102d8b2872a3f542caa00f

PR #259, opened 2019-02-14
Downgrade variable pointers

alan-baker:
Only add the minimal variable pointers capabilities.

Most of the changes are test updates.

This only includes the detection of necessary capabilities in SPIRVProducer.  Removing unused arguments will be a separate PR.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -6559,18 +6559,16 @@ bool SPIRVProducerPass::selectFromSameObject(Instruction *inst) {
       // Null values satisfy the constraint of selecting of selecting from the
       // same object.
       if (!value) {
-        if (auto *cst = dyn_cast<Constant>(value)) {
-          if (!cst->isNullValue() && (!hack_undef || !isa<UndefValue>(value)))
+        if (auto *cst = dyn_cast<Constant>(base)) {
+          if (!cst->isNullValue() && !(hack_undef && isa<UndefValue>(base)))
             value = base;
         } else {
           value = base;
         }
       } else if (base != value) {
         if (auto *base_cst = dyn_cast<Constant>(base)) {
-          if (base_cst->isNullValue())
+          if (base_cst->isNullValue() || (hack_undef && isa<UndefValue>(base)))
             continue;
-        } else if (hack_undef && isa<UndefValue>(base)) {
-          continue;
         }
 
         if (sameResource(value, base))
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -402,6 +402,17 @@ struct SPIRVProducerPass final : public ModulePass {
   uint64_t GetTypeStoreSize(Type *type, const DataLayout &DL);
   uint64_t GetTypeAllocSize(Type *type, const DataLayout &DL);
 
+  // Returns the base pointer of |v|.
+  Value *GetBasePointer(Value *v);
+
+  // Sets |HasVariablePointersStorageBuffer| or |HasVariablePointers| base on
+  // |address_space|.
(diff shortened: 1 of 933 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3863,6 +3913,12 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
       if (PointeeTy->isStructTy() &&
           dyn_cast<StructType>(PointeeTy)->isOpaque()) {
         Ty = PointeeTy;
+      } else {
+        // Selecting between pointers requires variable pointers.
+        setVariablePointersCapabilities(Ty->getPointerAddressSpace());
+        if (!hasVariablePointers() && !selectFromSameObject(&I)) {
```
dneto0: The !hasVariablePointers() is redundant.  But this is fine.

lib/SPIRVProducerPass.cpp
```diff
@@ -5196,7 +5278,30 @@ void SPIRVProducerPass::HandleDeferredInstruction() {
...
+              // Workgroup references require variable pointers if they are not
+              // memory object declarations.
+              if (auto *operand_call = dyn_cast<CallInst>(operand)) {
+                // Workgroup accessor represents a variable reference.
+                if (!operand_call->getCalledFunction()->getName().startswith(
+                        clspv::WorkgroupAccessorFunction()))
+                  setVariablePointers(true);
+              } else {
+                // Arguments are function parameters.
+                if (!isa<Argument>(operand))
```
dneto0: I guess if the pointer is a workgroup variable, then it shows up as a call to the workgroup accessor function.
So I see how this implements the condition.  ok

lib/SPIRVProducerPass.cpp
```diff
@@ -6376,3 +6481,107 @@ uint64_t SPIRVProducerPass::GetTypeAllocSize(Type *type, const DataLayout &DL) {
...
+void SPIRVProducerPass::setVariablePointersCapabilities(unsigned address_space) {
+  if (GetStorageClass(address_space) == spv::StorageClassStorageBuffer) {
+    setVariablePointersStorageBuffer(true);
+  } else {
+    setVariablePointers(true);
+  }
+}
+
+Value *SPIRVProducerPass::GetBasePointer(Value* v) {
+  if (auto *gep = dyn_cast<GetElementPtrInst>(v)) {
```
dneto0: Could have been a while loop.

lib/SPIRVProducerPass.cpp
```diff
@@ -6376,3 +6481,107 @@ uint64_t SPIRVProducerPass::GetTypeAllocSize(Type *type, const DataLayout &DL) {
...
+        return true;
+    }
+
+    if (auto *false_cst = dyn_cast<Constant>(false_base)) {
+      if (false_cst->isNullValue() ||
+          (hack_undef && isa<UndefValue>(false_base)))
+        return true;
+    }
+
+    if (sameResource(true_base, false_base))
```
dneto0: could do just:  return sameResouce(true_base, false_base)

lib/SPIRVProducerPass.cpp
```diff
@@ -6376,3 +6481,107 @@ uint64_t SPIRVProducerPass::GetTypeAllocSize(Type *type, const DataLayout &DL) {
...
+        } else {
+          value = base;
+        }
+      } else if (base != value) {
+        if (auto *base_cst = dyn_cast<Constant>(base)) {
+          if (base_cst->isNullValue() || (hack_undef && isa<UndefValue>(base)))
+            continue;
+        }
+
+        if (sameResource(value, base))
```
dneto0: Slightly confusing.  I would have written.
`ok = ok && sameResource(value,base)` and avoided the continue.

PR #266, opened 2019-02-24
Generalise the production of constants for OpSelect

### LINKED ISSUE #235: Incorrect code produced for sign/zero extension from boolean to 64-bit integer
`OpSelect` with constants is used but the constants are always 32-bit integers.

kpet:
...so that non-32-bit integer types are correctly handled.

Also delete a bit of nearby dead code.

Fixes #235.

### DIFF
--- a/test/opselect_constants_int_scalar.cl
+++ b/test/opselect_constants_int_scalar.cl
@@ -1,0 +1,17 @@
+// RUN: clspv  %s -S -o %t.spvasm
+// RUN: FileCheck %s < %t.spvasm
+// RUN: clspv  %s -o %t.spv
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
+// CHECK:     %[[uint:[0-9a-zA-Z_]+]] = OpTypeInt 32 0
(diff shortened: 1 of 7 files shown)


test/opselect_constants_int_scalar.cl
```diff
@@ -0,0 +1,72 @@
+// RUN: clspv  %s -S -o %t.spvasm
```
alan-baker: I recently had to update ~900 tests when making an unrelated change. Could you please cut the number of CHECK statements to make this a more targeted test? I would like to get to a state where fewer tests  need to be updated for unrelated changes. It also makes the tests easier to follow.

In this case I think the following would be sufficient:
```
CHECK: [[uint:%[a-zA-Z0-9_]+]] = OpTypeInt 32 0
CHECK-DAG: [[zero:%[a-zA-Z0-9_]+]] = OpConstant [[uint]] 0
CHECK-DAG: [[one:%[[a-zA-Z0-9_]+]] = OpConstant [[uint]] 1
CHECK: OpSelect [[uint]] {{.*}} [[zero]] [[one]]
```

Similarly other tests could be reduced in terms of number of checks.

kpet: I agree. I'm having to change a lot of tests as part of another change I have in progress and this is particularly painful and a waste of our time.

PR #269, opened 2019-03-01
Remove a debug print statement

aarongreig:


aarongreig:
I have now associated my work email with my github account, we have a company level CLA.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -925,7 +925,7 @@ bool ReplacePointerBitcastPass::runOnModule(Module &M) {
           }
         } else {
 
-          errs() << "About to crash at 820" << M << "Aboot\n\n";
+          errs() << "About to crash at 820" << M << "Abort\n\n";
```
alan-baker: You can just delete this line.

dneto0: My bad.  I inserted it a long time ago so I could track down bad behaviour.  I never intended to check it in.  :-)

PR #272, opened 2019-03-04
Coherent decoration for global variables

alan-baker:
Global variables that are read and written in kernels that have barrier/fence synchronization are now decorated as coherent.

Descriptor allocation identifies whether variables are both read and written in synchronized kernels. Coherency is a new discriminant so coherent and non-coherent variables are not shared.

The spirv producer annotates function parameters by searching backwards to see which resource variables are used with it. Any coherent variable turns a function parameter coherent.

### DIFF
--- a/lib/AllocateDescriptorsPass.cpp
+++ b/lib/AllocateDescriptorsPass.cpp
@@ -926,13 +926,13 @@ bool AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(Function *F) {
                 "__spirv_control_barrier")) {
           // barrier()
           if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(2))) {
-            uses_barrier = semantics->getZExtValue() & 0x40;
+            uses_barrier = semantics->getZExtValue() & kMemorySemanticsUniformMemory;
           }
         } else if (call->getCalledFunction()->getName().equals(
                        "__spirv_memory_barrier")) {
           // mem_fence()
           if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(1))) {
-            uses_barrier = semantics->getZExtValue() & 0x40;
+            uses_barrier = semantics->getZExtValue() & kMemorySemanticsUniformMemory;
           }
         } else if (!call->getCalledFunction()->isDeclaration()) {
           // Continue searching in the subfunction.
--- a/lib/AllocateDescriptorsPass.cpp
+++ b/lib/AllocateDescriptorsPass.cpp
@@ -81,6 +81,18 @@ private:
     return result;
   }
 
+  // Returns true if |F| or call function |F| calls contains a global barrier.
+  // Specifically, it checks that the memory semantics operand contains
+  // UniformMemory memory semantics.
+  bool CallTreeContainsGlobalBarrier(Function *F);
+
+  // Returns a pair indicating if |V| is read and/or written to.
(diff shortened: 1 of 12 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3175,10 +3189,30 @@ void SPIRVProducerPass::GenerateFuncPrologue(Function &F) {
   //
 
   if (F.getCallingConv() != CallingConv::SPIR_KERNEL) {
+
+    // Find Insert Point for OpDecorate.
+    auto DecoInsertPoint =
```
dneto0: Sigh.  This is so dodgy.  Let's make this better at some point.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -864,3 +899,106 @@ bool AllocateDescriptorsPass::AllocateLocalKernelArgSpecIds(Module &M) {
 
   return Changed;
 }
+
+bool AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(Function *F) {
```
dneto0: Is it worth saying somewhere that because we're assuming OpenCL 1.2, atomics are always Relaxed.  So they don't synchronize.  So the presence of an atomic op does not require the addition of a Coherent decoration.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -864,3 +899,106 @@ bool AllocateDescriptorsPass::AllocateLocalKernelArgSpecIds(Module &M) {
...
+      break;
+  }
+
+  barrier_map_.insert(std::make_pair(F, uses_barrier));
+  return uses_barrier;
+}
+
+std::pair<bool, bool> AllocateDescriptorsPass::HasReadsAndWrites(Value *V) {
+  auto IsInterestingUser = [](const User *user) {
+    if (isa<StoreInst>(user) || isa<LoadInst>(user) || isa<CallInst>(user) || user->getType()->isPointerTy())
```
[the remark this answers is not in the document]
alan-baker: I also added explicit handling of modf and frexp. I can only test frexp since modf is untranslated currently.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -40,6 +40,10 @@ using namespace llvm;
 
 namespace {
 
+// Constant that represents bitfield for UniformMemory Memory Semantics from
+// SPIR-V. Used to test barrier semantics.
+const uint32_t kMemorySemanticsUniformMemory = 0x40;
```
dneto0: ok, so you made this constant, ... but then didn't use it.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -864,3 +899,106 @@ bool AllocateDescriptorsPass::AllocateLocalKernelArgSpecIds(Module &M) {
...
+
+  bool uses_barrier = false;
+  for (auto &BB : *F) {
+    for (auto &I : BB) {
+      if (auto *call = dyn_cast<CallInst>(&I)) {
+        if (call->getCalledFunction()->getName().equals(
+                "__spirv_control_barrier")) {
+          // barrier()
+          if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(2))) {
+            uses_barrier = semantics->getZExtValue() & 0x40;
```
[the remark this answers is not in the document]
dneto0: You made the constant but didn't use it. :-(

PR #281, opened 2019-03-06
Update SPIRV-Tools and SPIRV-Headers deps

alan-baker:
* Validator catches some errors with unit tests now
 * no longer generate NonWritable for read_only images since we generate
 them as sampled
 * have some tests expect to fail validation

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2685,12 +2685,6 @@ void SPIRVProducerPass::GenerateResourceVars(Module &) {
                              new SPIRVInstruction(spv::OpDecorate, Ops));
       }
       break;
-    case clspv::ArgKind::ReadOnlyImage:
-      Ops.clear();
-      Ops << MkId(info->var_id) << MkNum(spv::DecorationNonWritable);
-      SPIRVInstList.insert(DecoInsertPoint,
-                           new SPIRVInstruction(spv::OpDecorate, Ops));
-      break;
     case clspv::ArgKind::WriteOnlyImage:
       Ops.clear();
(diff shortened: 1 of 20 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -2685,12 +2685,6 @@ void SPIRVProducerPass::GenerateResourceVars(Module &) {
                              new SPIRVInstruction(spv::OpDecorate, Ops));
       }
       break;
-    case clspv::ArgKind::ReadOnlyImage:
```
alan-baker: Because we generate read_only images as sampled I had to remove this to validate.

test/ImageBuiltins/get_image_height_image2d_readonly.cl
```diff
@@ -3,7 +3,8 @@
 // RUN: clspv %s -o %t.spv
 // RUN: spirv-dis -o %t2.spvasm %t.spv
 // RUN: FileCheck %s < %t2.spvasm
-// RUN: spirv-val --target-env vulkan1.0 %t.spv
+// #219: Sampled images cannot be queried in Vulkan.
+// RUN: not spirv-val --target-env vulkan1.0 %t.spv
```
alan-baker: Sampled images cannot be queried in Vulkan.

test/ProgramScopeConstants/constant_array_with_function_call.cl
```diff
@@ -3,30 +3,17 @@
...
-
-
-
-
-
-
-
-
 
+// TODO(dneto): OpPtrAccessChain on pointer to Private is not alloweed by SPV_KHR_variable_pointers
```
alan-baker: Validator is catching this now, so expect failure.

PR #285, opened 2019-03-09
Allow -cluster-pod-kernel-args on kernels with only pod arguments

kpet:
There are programs out there with code like this.

### DIFF
--- a/lib/ClusterPodKernelArgumentsPass.cpp
+++ b/lib/ClusterPodKernelArgumentsPass.cpp
@@ -265,8 +265,9 @@ bool ClusterPodKernelArgumentsPass::runOnModule(Module &M) {
       CalleeArgs.back()->setName(Arg.getName());
     }
     assert(ptrIndex + podIndex == F->arg_size());
-    assert(ptrIndex = PtrArgTys.size());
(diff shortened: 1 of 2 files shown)


lib/ClusterPodKernelArgumentsPass.cpp
```diff
@@ -265,7 +265,7 @@ bool ClusterPodKernelArgumentsPass::runOnModule(Module &M) {
       CalleeArgs.back()->setName(Arg.getName());
     }
     assert(ptrIndex + podIndex == F->arg_size());
-    assert(ptrIndex = PtrArgTys.size());
+    assert(ptrIndex == PtrArgTys.size());
     assert(podIndex = PodArgTys.size());
```
alan-baker: This is likely also a bug that is meant to be an equality test and not an assignment.

kpet: That was my initial reaction but then I looked at the logic above, realised it was a valid thing to assert for (`podIndex` should never be zero at that stage as kernels with no POD arguments are skipped earlier) and decided to keep the assert. I've now made that two separate asserts to make things more obvious/readable.

PR #289, opened 2019-03-13
Only remove trivially dead insts in UndoBoolPass

### LINKED ISSUE #288: UndoBoolPass crashes on some kernels with char support
The UndoBoolPass assumes all uses of an i1 to i8 conversion are later truncated back to i1. This is not true when using chars generally (via the -int8 option). This pass should only delete trivially dead instructions.

alan-baker:
Fixes #288

* Valid uses cases when -int8 is enabled
* In either event the IR would be invalid if the instruction had uses
that remained.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -15,6 +15,8 @@
 #ifndef CLSPV_INCLUDE_CLSPV_OPTION_H_
 #define CLSPV_INCLUDE_CLSPV_OPTION_H_
 
+#include <cstdint>
+
 namespace clspv {
 namespace Option {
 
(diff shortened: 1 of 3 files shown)


include/clspv/Option.h
```diff
@@ -15,6 +15,8 @@
 #ifndef CLSPV_INCLUDE_CLSPV_OPTION_H_
 #define CLSPV_INCLUDE_CLSPV_OPTION_H_
 
+#include <cstdint>
```
alan-baker: Noticed this was an issue while debugging and including this file in UndoBoolPass.

PR #300, opened 2019-03-25
Unify handling of most SPIR-V intrinsics in the producer pass

kpet:
The end goal is to provide a generic mechanism that makes
it easy to directly generate a specific instruction in
other passes.

This change only covers the trivial cases.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -4470,61 +4470,54 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
       break;
     }
 
-    if (Callee->getName().startswith("spirv.atomic")) {
-      spv::Op opcode = StringSwitch<spv::Op>(Callee->getName())
-                           .Case("spirv.atomic_add", spv::OpAtomicIAdd)
-                           .Case("spirv.atomic_sub", spv::OpAtomicISub)
-                           .Case("spirv.atomic_exchange", spv::OpAtomicExchange)
-                           .Case("spirv.atomic_inc", spv::OpAtomicIIncrement)
-                           .Case("spirv.atomic_dec", spv::OpAtomicIDecrement)
-                           .Case("spirv.atomic_compare_exchange",
-                                 spv::OpAtomicCompareExchange)
-                           .Case("spirv.atomic_umin", spv::OpAtomicUMin)
-                           .Case("spirv.atomic_smin", spv::OpAtomicSMin)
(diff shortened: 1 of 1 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -4470,61 +4470,56 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
...
-                           .Default(spv::OpNop);
-
-      //
-      // Generate OpAtomic*.
-      //
-      SPIRVOperandList Ops;
-
-      Ops << MkId(lookupType(I.getType()));
+    // Handle SPIR-V intrinsics
+    std::map<StringRef, spv::Op> SPVIntrToOpMap = {
```
alan-baker: I'm concerned about the name overloading. Why didn't you leave unique names and change these to .StartsWith for the overloads? I'm worried someone could query an overload and get the wrong type. Do you have tests that use multiple variants of isnan for example?

kpet: The `SPIRVProducerPass` was treating all of them in the same way which makes sense as all the required type information is conveyed by the IR types alone. FWIW, all the relevant OpenCL conformance tests pass with this change.

alan-baker: Have you run a test that includes multiple variants of the same builtin? It feels like there would be a problem creating those function in LLVM.

kpet: Oh, now I understand what you meant. It's probably broken. I'll look into it.

kpet: Yeah, it's all horribly broken. https://github.com/google/clspv/pull/294 and https://github.com/google/clspv/pull/304 suffer from the same problem. I'll edit this PR to fix `isinf` and `isnan` and fix #294 as part of introducing a generic mechanism to deal with these. #304 probably shouldn't go in until then. Thanks for spotting this :).

kpet: Actually, atomics have suffered from this issue since the first public commit so you may as well let #304 in. I'll fix all these.

alan-baker: Thanks! I've merged #304

kpet: Thanks! I've updated this PR. The updated code doesn't address any of the existing issues but should no longer add more of them.

kpet:
Will do. I'll  try to include an exhaustive list of the built-in functions affected as I don't intend to fix all of them in a single PR.

PR #303, opened 2019-03-26
Support chars in UBOs

### LINKED ISSUE #279: Support chars in UBOs
Extend #278 to allow chars in UBO (via -constant-args-ubo). The type mutation requires significant updates since it can no longer blindly replace char arrays. Likely need to perform layout checks in the transform pass itself.

alan-baker:
Fixes #279

* Type mutation now replaces any char array if char arrays are not
supported
  * char arrays are supported if -int8 and -std430-ubo-layout are
  specified
* New tests

### DIFF
--- a/lib/UBOTypeTransformPass.cpp
+++ b/lib/UBOTypeTransformPass.cpp
@@ -104,6 +104,9 @@ private:
 
   // Maps a global value to its replacement.
   DenseMap<Constant *, Constant *> remapped_globals_;
+
+  // Whether char arrays are supported in UBOs.
+  bool support_int8_array_;
 };
 
 char UBOTypeTransformPass::ID = 0;
@@ -121,6 +124,10 @@ namespace {
 bool UBOTypeTransformPass::runOnModule(Module &M) {
   if (!clspv::Option::ConstantArgsInUniformBuffer())
     return false;
+
+  // Record whether char arrays are supported.
+  support_int8_array_ = clspv::Option::Int8Support() &&
+                        clspv::Option::Std430UniformBufferLayout();
 
   bool changed = false;
   for (auto &F : M) {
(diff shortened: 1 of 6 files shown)


lib/UBOTypeTransformPass.cpp
```diff
@@ -230,8 +230,14 @@ StructType *UBOTypeTransformPass::MapStructType(StructType *struct_ty,
...
     uint64_t offset = layout->getElementOffset(i);
     const auto *array = dyn_cast<ArrayType>(element);
-    if (array && array->getElementType()->isIntegerTy(8) && offset % 16 != 0) {
+    // Unless char arrays in UBOs are supported, replace all instances with an
+    // i32.
+    const bool support_int8_array = clspv::Option::Int8Support() &&
+                                    clspv::Option::Std430UniformBufferLayout();
+    if (array && array->getElementType()->isIntegerTy(8) && !support_int8_array) {
       // This is a padding element.
+      assert((array->getNumElements() % 4 == 0) &&
```
dneto0: This is not always the case.

Also, the alignment is not necessarily updated.

Here's an example that breaks both issues:

```

typedef struct {
  char a;
  int b __attribute__((aligned(16)));
} S;

 kernel void foo(global S* out, constant S* in) {
  out->b = in->b;
}
```
Compiling with -constant-args-ubo -o s.spv -print-before-all -inline-entry-points -int8, I get the last LLVM IR with a struct like this:

        %struct.S = type { i8, [15 x i8], i32, [12 x i8] }

And the struct definition is:

               OpMemberDecorate %_struct_3 0 Offset 0
               OpMemberDecorate %_struct_3 1 Offset 1
               OpMemberDecorate %_struct_3 2 Offset 16
               OpMemberDecorate %_struct_3 3 Offset 20
            ...
           %_struct_3 = OpTypeStruct %uchar %uint %uint %uint

It also fails validation:

error: line 28: Structure id 3 decorated as Block for variable in StorageBuffer storage class must follow standard storage buffer layout rules: member 1 at offset 1 is not aligned to 4
  %_struct_3 = OpTypeStruct %uchar %uint %uint %uint

alan-baker: Good catch. If supporting chars, the padding replacement could also be a char instead of an int. That's probably the easiest way out of this.

Issue #308, opened 2019-03-28
Make sure SPIR-V intrinsics have got a unique name


kpet:
Functions and calls are used in a number of places to encode specific SPIR-V instructions. In several cases the same function name is used with different type signatures, making it impossible to use both in the same module. All uses of functions to encode SPIR-V instructions should be reviewed to ensure they don't suffer from this issue. Each time a single test that uses all possible overloads should be added.

rjodinchr:
I think this is now fixed. Closing the issue. Feel free to reopen if I am wrong here.

PR #309, opened 2019-03-28
Introduce a mechanism to call specific SPIR-V instructions

kpet:
Directly calling a specific SPIR-V instruction is a common
idiom which should be shorter to write than it currently is.
Besides, having a centralised scheme that deals with encoding
SPIR-V intrinsics removes the need to worry about issues like
#308.

This change introduces a utility function that inserts a call
to a SPIR-V instruction passed as a parameter. It automatically
inserts required functions into the module if needed and
guarantees they have a unique name with a simple mangling
scheme.

The current encoding makes the IR a bit less readable as
the functions used to encode SPIR-V instructions are named
after the opcode number rather than the instruction name.
This can be improved upon later if felt necessary.

This change also uses the new mechanism for the mul_hi/mad_hi
built-in functions, fixing the issue described in #308 for
these.

### DIFF
--- a/lib/SPIRVOp.cpp
+++ b/lib/SPIRVOp.cpp
@@ -1,0 +1,85 @@
+// Copyright 2019 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "SPIRVOp.h"
+
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Module.h"
+
+#include "Constants.h"
+
+namespace clspv {
+
+using namespace llvm;
+
+// The mangling loosely follows the Itanium convention.
+// Its purpose is solely to ensure uniqueness of names, it is not
+// meant to convey type information.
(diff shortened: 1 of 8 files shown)


lib/SPIRVOp.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // Create a function in the module
+  auto M = Insert->getModule();
+  auto Int32Ty = Type::getInt32Ty(M->getContext());
+  SmallVector<Type*, 8> ArgTypes = { Int32Ty };
+  for (auto Arg : Args) {
+    ArgTypes.push_back(Arg->getType());
+  }
+  auto NewFType = FunctionType::get(RetType, ArgTypes, false);
+  auto NewF = M->getOrInsertFunction(MangledName, NewFType);
```
dneto0: For the mad_hi case, it might help optimization to call NewF->setDoesNotAccessMemory() or equivalent.

I think the best way to do that is to add another parameter which is the list of attributes to be applied to the function.  For mad_hi it would include Attribute::ReadNone

dneto0: To be clear, I mean another parameter to the InsertSpirvOp function, could be an ArrayRef<Attribute>

kpet: Yes, I shouldn't have been able to write the tests the way I did :). I was planning to look into this later. I was wondering whether we could always add this attribute based on the absence of pointer arguments. For now, I just went with your suggestion.

I've rebased the change on top of the clang/llvm update. PTAL.

lib/SPIRVOp.h
```diff
@@ -0,0 +1,33 @@
...
+#include "llvm/IR/Type.h"
+#include "llvm/IR/Value.h"
+
+#include "spirv/1.0/spirv.hpp"
+
+namespace clspv {
+
+using namespace llvm;
+
+// Insert a call to a specific SPIR-V instruction after Insert
```
alan-baker: Please make this comment a bit more verbose. Mention unique name generation and that `Attributes` must be function attributes.

kpet: I should indeed have done a better job there. I've just pushed a commit with the requested changes. I've also added a warning about the symbol table being changed as that tripped me on the very first use.

PR #313, opened 2019-03-31
Don't assume that clspv and clspv-opt are in the same folder

kpet:
All tests that use clspv-opt will fail in a project that embeds
clspv and changes the location of only the clspv binary in the
build output.


alan-baker:
clspv-opt is not enabled on windows right now. That's where the errors are coming from.

kpet:
I've updated the change. It should work on all platforms. PTAL.

Issue #276, opened 2019-03-06
Test general half support


alan-baker:
Half support is available in OpenCL-C with the extension cl_khr_fp16. Support appears to be in place to generate the Float16 capability and handle the half type generally. Add some targeted tests to provide coverage of this functionality.

File new issues for anything that doesn't work specific to half.

alan-baker:
Likely missing builtin support compared to float.

rjodinchr:
Looking at the CTS, half support looks good. Should we close this issue?

PR #318, opened 2019-04-02
Add support for the copysign built-in function

### LINKED ISSUE #287: Support copysign
> gentype copysign (gentype x, gentype y): Returns x with its sign changed to match the sign of y

Scalar or vector of floating point type.

Sign bit for floating point is the MSB. This can be done with bit manipulation.

kpet:
Passes the relevant conformance tests.

Closes #287.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -159,6 +159,7 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
 
   bool runOnModule(Module &M) override;
   bool replaceAbs(Module &M);
+  bool replaceCopysign(Module &M);
   bool replaceRecip(Module &M);
   bool replaceDivide(Module &M);
   bool replaceExp10(Module &M);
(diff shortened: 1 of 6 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -300,6 +302,81 @@ bool ReplaceOpenCLBuiltinPass::replaceAbs(Module &M) {
   return Changed;
 }
 
+bool ReplaceOpenCLBuiltinPass::replaceCopysign(Module &M) {
+  bool Changed = false;
+
+  const char *Names[] = {
+    "_Z8copysignff",
```
alan-baker: What about double and half support?

kpet: The code should just work in both cases. clspv needs some general work for double support, I've created #321. For half, well, there aren't any conformance tests (see https://github.com/KhronosGroup/OpenCL-CTS/issues/142). I suggest looking at half support/testing this as part of #276 or creating another issue similar to #321 as support is missing in quite a few places.

PR #327, opened 2019-04-03
Fix test using -S

alan-baker:
I merged a change that needed rebased. This fixes the test.


kpet:
#309 needs rebasing as well. I'll do that.

PR #331, opened 2019-04-04
[RFC] Experimental C++ support

kpet:
This change adds experimental support for C++ via Clang's
OpenCL C++ mode.

In that mode Clang uses OpenCL's generic address space for
a number of things so I've added it to
clspv::AddressSpace to align clspv's internal convention
with what Clang does a bit better.

This currently has to be used with -inline-entry-points to
maximise the chances that all traces of generic address
space will have been removed before passes that don't
support it. I have another change that plugs LLVM's address
space inference pass and enables quite a bit more code to be
compiled (and the use of the generic address space generally)
but it requires a couple of minor changes to LLVM.

Kernel function overloads are forbidden in the FrontendPlugin
with a custom diagnostic message. Kernel function names
are demangled before the descriptor map creation and the
absence of overload guarantees name unicity.

### DIFF
--- a/include/clspv/AddressSpace.h
+++ b/include/clspv/AddressSpace.h
@@ -22,6 +22,7 @@ enum Type {
   Global,            // OpenCL global memory.
   Constant,          // OpenCL constant memory.
   Local,             // OpenCL local memory.
+  Generic,           // OpenCL generic.
   Input,             // Vulkan input memory.
   Uniform,           // Vulkan uniform memory.
   UniformConstant,   // Vulkan uniform constant memory.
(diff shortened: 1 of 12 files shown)


alan-baker:
At first glance this looks very cool!

dneto0:
> I have another change that plugs LLVM's address
space inference pass and enables quite a bit more code to be
compiled (and the use of the generic address space generally)
but it requires a couple of minor changes to LLVM.

This project is structured to only use upstream LLVM.  So I'm cautious about modifying LLVM.  But I'll withhold judgement until see more.

kpet:
>This project is structured to only use upstream LLVM. So I'm cautious about modifying LLVM. But I'll withhold judgement until see more.

I have started the process to upstream the required change to LLVM. There shouldn't be any need to do anything other than update the version used by clspv.

PR #338, opened 2019-04-08
Don't require inlining to use UBOs

### LINKED ISSUE #330: Support -constant-args-ubo without inlining
Currently -constant-args-ubo also requires specifying -inline-entry-points. This shouldn't be a necessity. The UBO mutations need to be more targeted.

alan-baker:
Fixes #330 

* changes to the type mutation to touch fewer types
* RebuildType just calls MapType with rewrite == false to be clear

Pass to specialize/inline functions with UBOs

SPIR-V does not support passing UBOs by reference. If DRA cannot
determine a singular source of a UBO, the new pass will either
specialize it or inline it. The decision is made on a per call basis.
* Refactored CallGraphOrderedFunctions to de-duplicate it
* added cleanup passes to run after specialization
* Added tests

This removes the requirement to exhaustively inline entry points. It has many opportunities for future improvement by handling more cases that it currently resorts to inlining to solve. Currently it handles chains of GEPs from a resource variable.

Future improvements:
* This pass hits similar targets as DRA. Due to the selective inlining it would be possible to further DRA a module before specializing a call
* Handle more cases than just GEPs
  * e.g. iterator increment in a loop could be transformed into an induction variable passed to the function call and used as a GEP index instead of the GEP being passed
* Each call is specialized individually, it is possible in some scenarios to reduce the number of specializations by sharing them between calls

Worth noting that this doesn't solve the underlying issue of kernels using fundamental variable pointers functionality (e.g. selection between pointers) in UBOs, but it does currently solve the reference passing aspect.

### DIFF
--- a/include/clspv/Passes.h
+++ b/include/clspv/Passes.h
@@ -368,4 +368,8 @@ llvm::ModulePass *createRemoveUnusedArgumentsPass();
 /// Demangles the name of all kernel functions.
 llvm::ModulePass *createDemangleKernelNamesPass();
 
+/// Specializes or inlines functions with UBO arguments to satisfy SPIR-V
+/// requirements.
+llvm::ModulePass *createMultiVersionUBOFunctionsPass();
+
 } // namespace clspv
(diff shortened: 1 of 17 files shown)


lib/Utils.cpp
```diff
@@ -0,0 +1,107 @@
+// Copyright 2019 The Clspv Authors. All rights reserved.
```
dneto0: I would have called this CallGraphOrderedFunctions.cpp

"util" and "misc" generally become the attic of a project, with poor cohesion.

alan-baker: Good point, I'll make this change.

Issue #322, opened 2019-04-02
clspv-opt option parsing broken


kpet:
`clspv-opt` is using substring searches (checking whether a given argument contains an option name) to determine which options to parse itself and which options it should pass to `opt`. One of the options in the search list is `debug` which means that passing a file that has `debug` in its name makes `clspv-opt` think that it's an option it should parse. It will need something more robust.

dnovillo:
This will be fixed as a side-effect of fixing #307.  It's forced me to re-implement clspv-opt from whole cloth instead of using the plugin facilities of LLVM opt.  I'm currently working on this.

dnovillo:
Fixed with a4c44fa

PR #346, opened 2019-04-10
Re-implement clspv-opt as a standalone driver

### LINKED ISSUE #307: clsvp-opt should support windows
The current implementation of clsvp-opt (#283) does not support Windows because it is implemented as a plugin to LLVM's `opt`.  This is difficult to do out of tree because of the way LLVM and Windows DLL libraries work (http://go/so/41943353).

Extend clsvp-opt so it knows how to read/write LLVM IR without having to use `opt`.

dnovillo:
This fixes #307 and #322.

It re-implements clspv-opt as a standalone driver.  The initial
implementation used the plugin facility for LLVM opt, but this is
difficult to implement in Windows (#307).

The driver only supports LLVM scalar transformations and all of clspv
transformations.  To do this, passes must be declared using LLVM's
INITIALIZE_PASS macro.  Additionally, the function
initializeClspvPasses() must include a call to the initialize*()
function generated by INITIALIZE_PASS.

This change also adds tests for all the existing passes in clspv.  It
does not check for specific transformations.  It checks that the flags
are accepted by clspv-opt.

Finally, this change reduces the build time for LLVM by building no
target backends by default.


include/clspv/Passes.h
```diff
@@ -25,6 +25,47 @@ class ModulePass;
...
 template <typename T> class ArrayRef;
+
+// Clspv pass initializers.
+class PassRegistry;
+
+// Initialize all clspv passes.
+void initializeClspvPasses(PassRegistry &);
+
+// Individual pass initializers.
+void initializeAllocateDescriptorsPassPass(PassRegistry &);
```
alan-baker: Do these individual pass initializers need to be exposed in the public header?

lib/AllocateDescriptorsPass.cpp
```diff
@@ -160,8 +160,8 @@ class AllocateDescriptorsPass final : public ModulePass {
 } // namespace
 
 char AllocateDescriptorsPass::ID = 0;
-static RegisterPass<AllocateDescriptorsPass> X("AllocateDescriptorsPass",
-                                               "Allocate resource descriptors");
+INITIALIZE_PASS(AllocateDescriptorsPass, "AllocateDescriptorsPass",
+                "Allocate resource descriptors", false, false)
```
alan-baker: One of the falses is whether this is an analysis. What is the other?

lib/Option.cpp
```diff
@@ -152,6 +155,46 @@ static llvm::cl::opt<bool>
               llvm::cl::desc("Enable experimental C++ support"));
 } // namespace
 
+namespace llvm {
+
+void initializeClspvPasses(PassRegistry &r) {
```
alan-baker: This is an odd file to contain the initialization. I'd suggest making a Passes.cpp.

include/clspv/Passes.h
dnovillo: In some header available to `clspv-opt`.

lib/AllocateDescriptorsPass.cpp
dnovillo: Whether it's a CFG-only pass.

lib/Option.cpp
dnovillo: Agreed.  I was going to suggest that and then forgot.

include/clspv/Passes.h
dnovillo: Oh, and available to the passes too.  The macro expands to code that defines the `initialize*` function, so the declaration needs to be available.  `clspv/Passes.h` seemed the best home for these.

alan-baker: Doesn't just the overall initializer need to be available to clspv-opt? Could we have a header local to lib/ that declares the individual pass initializers?

dnovillo: That should work.  Yes.

dnovillo:
Added `lib/Passes.{cpp,h}`.  PTAL.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -26,9 +26,9 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "Passes.h"
```
alan-baker: This (and other instances) no longer match the style of other local directory includes like ArgKind.h and Constants.h

dnovillo: What do you mean?

dnovillo: OK, fixed.  PTAL.

CMakeLists.txt
```diff
@@ -69,15 +69,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
-if (NOT MSVC)
-  set(ENABLE_CLSPV_OPT ON)
-else()
-  # clspv-opt is not supported on Windows, because LLVM opt does not support
-  # out-of-tree plugins on Windows (http://go/so/41943353).
-  # TODO(dnovillo): Need a different implementation that does not rely on
-  #                 LLVM opt.
-  set(ENABLE_CLSPV_OPT OFF)
-endif()
+option(ENABLE_CLSPV_OPT "Enable the clspv-opt driver." ON)
```
jgavris: I was trying an experiment to generate some code without optimizations, does this flag affect the `clspv` target at the moment? I built with this on and off, I get identical output for `clspv`.

alan-baker: This controls whether to build the standalone executable clspv-opt. It now builds on all platforms by default. The executable just lets us run individual passes on IR (for testing purposes). You can -O0 as an input to clspv to disable optimizations, but there are no guarantees it generates correct code.

PR #350, opened 2019-04-12
Add flag to emit LLVM IR after parsing.

dnovillo:
This adds the flag `--emit-ir` which can be used in conjunction with
`clspv-opt` to develop and debug clspv passes.

The flag emits LLVM IR right after the module has been parsed.  It
seemed the most natural place to dump the IR, because it gives the most
flexibility.  Once the initial IR has been generated, all the passes up
to the problematic one can be applied via `clspv-opt`.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -28,6 +28,7 @@
 #include "llvm/Support/ErrorOr.h"
 #include "llvm/Support/MathExtras.h"
 #include "llvm/Support/StringSaver.h"
+#include "llvm/Support/ToolOutputFile.h"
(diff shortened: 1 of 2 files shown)


dnovillo:
Since clspv does not expose SPIRV generation as a pass, it's not really possible to go to SPIRV from `clspv-opt`.  I'm thinking that this could be done either here or as a separate PR.  Thoughts?

alan-baker:
I was considering making a second pass around the core of SPIRVProducerPass that sets up output files based on the command line. Definitely could be a separate PR.

PR #353, opened 2019-04-15
Add an option to undefine __IMAGE_SUPPORT__

kpet:
Clang now defines it by default (was picked up in the update)
and the conformance tests check that this is defined if and only
if the runtime advertises support.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -113,6 +113,9 @@ bool Int8Support();
 // Returns true when compiling C++.
 bool CPlusPlus();
 
+// Returns true when images are supported.
+bool ImageSupport();
+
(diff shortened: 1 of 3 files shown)


kpet:
One of the bots failed and it seems access to the logs is now restricted. Is that expected?

kpet:
Yeah, I was on the fence. I settled on this because it felt more direct than `images=0` but I agree the double negation is not nice. I don't really mind either way. If you prefer the other way round I can invert the logic.

kpet:
@alan-baker Do you want me to invert the option?

alan-baker:
Sorry I lost track of this. I think, yes, let's invert the option.

PR #366, opened 2019-05-01
Normalize constant global variables

alan-baker:
Addresses some failures in customer kernels due to the way clang
codegen's constant aggregates.

* New utility to normalize constant global variables
  * called in cluster constants and spirv producer
  * rewrites global constants with constant bitcast users that are just
  simple reinterpret casts
  * added several tests

### DIFF
--- a/lib/NormalizeGlobalVariable.cpp
+++ b/lib/NormalizeGlobalVariable.cpp
@@ -1,0 +1,220 @@
+// Copyright 2019 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


lib/NormalizeGlobalVariable.cpp
```diff
@@ -0,0 +1,220 @@
...
+// arrays and structs broken down into constituent constants.
+void FlattenConstant(Constant *constant, std::vector<Constant *> *flattened) {
+  uint64_t num_elements = GetNumElements(constant->getType());
+  for (uint64_t i = 0; i != num_elements; ++i) {
+    auto *const_element = constant->getAggregateElement(i);
+    auto *element_ty = const_element->getType();
+    // Special cases for constant aggregate zero and constant data sequential
+    // to populate the right number of constant elements into |flattened|.
+    if (auto caz = dyn_cast<ConstantAggregateZero>(const_element)) {
+      for (auto i = 0; i != GetNumElements(element_ty); ++i) {
```
dneto0: I probably would have saved the num-elements computation.  But no big deal.

PR #369, opened 2019-05-07
Fix #361 - Initialize clspv passes into the registry.

### LINKED ISSUE #361: -print-{after,before}-all is broken
This was apparently broken by `a4c44fae` (Re-implement clspv-opt as a standalone driver).

dnovillo:
Passes were not registering their print-after/print-before flags.  This calls the pass initializer and adds a test for a clspv pass.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -935,6 +935,7 @@ int CompileFromSourceString(
   llvm::PassRegistry &Registry = *llvm::PassRegistry::getPassRegistry();
   llvm::initializeCore(Registry);
   llvm::initializeScalarOpts(Registry);
+  llvm::initializeClspvPasses(Registry);
 
   std::unique_ptr<llvm::Module> module(action.takeModule());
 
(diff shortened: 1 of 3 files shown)


lib/Compiler.cpp
```diff
@@ -805,6 +805,7 @@ int Compile(const int argc, const char *const argv[]) {
   llvm::PassRegistry &Registry = *llvm::PassRegistry::getPassRegistry();
   llvm::initializeCore(Registry);
   llvm::initializeScalarOpts(Registry);
+  llvm::initializeClspvPasses(Registry);
```
alan-baker: This likely needs done in `CompileFromSourceString` too.

dnovillo: Sure.  I don't see this function tested anywhere, though.  Are there unit tests I can add to?

dnovillo:
> This list of commits looks strange. Have you rebased recently?

Yeah.  I created a new branch off of an old branch.  Just squashed everything.

dnovillo:
@alan-baker Added a call from `CompileFromSourceString`.  PTAL.

lib/Compiler.cpp
alan-baker: Not without doing some legwork. That function is for the in process API.

PR #370, opened 2019-05-14
Handle more pointer bitcasts

### LINKED ISSUE #362: A build error when copying 64bit struct
Hi, I found a build error happens when copying 64bit struct.

source code
```
typedef struct __InstanceTest
{
  uint data_[2];
} InstanceTest;


// Copy a global instance to local
__kernel void testCopyInstance1(const __global InstanceTest* src)
{
  __local InstanceTest instances[16];

  const size_t index = get_global_id(0);
  if (index < 16) {
    instances[index] = src[index];
  }
}

// Copy a local instance to global
__kernel void testCopyInstance2(__global InstanceTest* dst)
{
  __local InstanceTest instances[16];

  const size_t index = get_global_id(0);
  if (index < 16) {
    dst[index] = instances[index];
  }
}
```

The clspv build result
```
% clspv struct_test.cl -o struct_test.spv
Invalid bitcast
  %5 = bitcast %struct.__InstanceTest %src_val to i64
Invalid bitcast
  %6 = bitcast i64 %5 to %struct.__InstanceTest
in function testCopyInstance1
LLVM ERROR: Broken function found, compilation aborted!
clspv version: 7094491
```
The build error didn't happen when the struct isn't 64bit (e.g. when member variable is `uint data_[3];`).
Could you take a look at it when you have time?

```
% clspv --version
LLVM (http://llvm.org/):
  LLVM version 9.0.0svn
  Optimized build.
  Default target: x86_64-unknown-linux-gnu
  Host CPU: znver1
```

Thanks.

alan-baker:
Fixes #362 

Add a more general type conversion utility to ReplacePointerBitcast. I didn't put the functions in the class because I expect they might be useful to refactor to a more general utility later.

Tests exercise hitting both loads and stores. Conversion breaks a type down into its constituent elements and then rebuilds it as a new type. The type must not be padded. This does not address the fact that llvm adds an Int64 capability requirement to the SPIR-V. That would need to be addressed through LLVM patches.


alan-baker:
The tests point out how silly the generated code is, but it does not seem worth trying to hit the special case of memcpy.

PR #374, opened 2019-05-29
Allow external LLVM and Clang folders.

dj2:
This CL updates the build files to allow setting the
CLSPV_LLVM_SOURCE_DIR and CLSPV_CLANG_SOURCE_DIR without specifying an
EXTERNAL_LLVM.

Because those directories maybe outside the clspv/ source tree we have
to provide the build directory when calling add_subdirectory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,13 +85,19 @@ if (${EXTERNAL_LLVM} EQUAL 1)
     message(FATAL_ERROR "External LLVM requires CLSPV_LLVM_BINARY_DIR to be specified")
   endif()
 else()
-  use_component(${CMAKE_CURRENT_SOURCE_DIR}/third_party/clang)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -101,9 +107,8 @@ else()
   set(LLVM_TARGETS_TO_BUILD ""
       CACHE STRING "Semicolon-separated list of targets to build, or \"all\".")
 
-
   # Then pull in LLVM for building.
-  add_subdirectory(${CLSPV_LLVM_SOURCE_DIR} EXCLUDE_FROM_ALL)
```
alan-baker: Why is this necessary? It seems strange to add a directory in the build tree.

dj2: In my case I checkout third_party/clspv, third_party/clspv-llvm and third_party/clspv-clang. You can't add a subdirectory which is outside your subtree without providing the build directory. So, to keep things the same I set the build directory to what it would have been if it was in-subtree.

PR #389, opened 2019-06-13
Use InsertSPIRVOp to create calls to barrier SPIR-V intrinsics

kpet:
Add missing test for coherent decorations.

Contributes to #308.

### DIFF
--- a/lib/AllocateDescriptorsPass.cpp
+++ b/lib/AllocateDescriptorsPass.cpp
@@ -924,14 +924,14 @@ bool AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(Function *F) {
         // and StorageBuffer storage classes) semantics are checked because
         // Workgroup variables are inherently coherent (and do not require the
         // decoration).
-        if (call->getCalledFunction()->getName().startswith("spirv.op.224")) {
+        if (call->getCalledFunction()->getName().startswith("spirv.op.224.")) {
           // barrier()
           if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(3))) {
             uses_barrier =
                 semantics->getZExtValue() & kMemorySemanticsUniformMemory;
           }
         } else if (call->getCalledFunction()->getName().startswith(
-                       "spirv.op.225")) {
+                       "spirv.op.225.")) {
           // mem_fence()
           if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(2))) {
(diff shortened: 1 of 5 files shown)


lib/AllocateDescriptorsPass.cpp
```diff
@@ -924,17 +924,16 @@ bool AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(Function *F) {
...
           // barrier()
-          if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(2))) {
+          if (auto *semantics = dyn_cast<ConstantInt>(call->getOperand(3))) {
             uses_barrier =
                 semantics->getZExtValue() & kMemorySemanticsUniformMemory;
           }
-        } else if (call->getCalledFunction()->getName().equals(
-                       "__spirv_memory_barrier")) {
+        } else if (call->getCalledFunction()->getName().startswith(
+                       "spirv.op.225")) {
```
dneto0: Need to guard against the case where an instruction number is 2250, for example.
I think this should have a "." after the 225, so: "spirv.op.225."

lib/AllocateDescriptorsPass.cpp
```diff
@@ -924,17 +924,16 @@ bool AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(Function *F) {
         // and StorageBuffer storage classes) semantics are checked because
         // Workgroup variables are inherently coherent (and do not require the
         // decoration).
-        if (call->getCalledFunction()->getName().equals(
-                "__spirv_control_barrier")) {
+        if (call->getCalledFunction()->getName().startswith("spirv.op.224")) {
```
dneto0: Should add a period after the 224.  Guards against an instruction numbered 2240, for example.

test/CommonBuiltins/no_duplicate_barrier.ll
```diff
@@ -2,7 +2,7 @@
 ; RUN: FileCheck %s < %t
 
 ; CHECK: noduplicate
-; CHECK-NEXT: __spirv_control_barrier
+; CHECK-NEXT: spirv.op.224
```
dneto0: add a period at the end.

lib/AllocateDescriptorsPass.cpp
kpet: Yes, better safe than sorry.

PR #401, opened 2019-07-12
Update llvm and clang deps

alan-baker:
This is necessary to unblock https://github.com/google/amber/pull/531


alan-baker:
I'm going to have to remove the msvc2015 bot.

alan-baker:
MSVC bots seem to have too old a version of MSVC 2017. Works on a local Windows machine for me.

PR #418, opened 2019-08-11
Prevent OpVariable from being produced with an OpUndef initializer

kpet:
This was avoided most of the time by luck because the OpUndef
wouldn't be in the value map. This change makes it a guarantee
and the added assert will catch any cases where initializers
are omitted by accident.

The new test case has been produced by reducing a kernel from
the SHOC benchmark suite.

### DIFF
--- a/test/global-variable-no-undef-initializer.cl
+++ b/test/global-variable-no-undef-initializer.cl
@@ -3,8 +3,11 @@
 // RUN: FileCheck %s < %t2.spvasm
 // RUN: spirv-val --target-env vulkan1.0 %t.spv
 
-// CHECK-DAG: %[[__original_id_29:[0-9]+]] = OpVariable {{.*}} Workgroup
-// CHECK-DAG: %[[__original_id_1:[0-9]+]] = OpVariable {{.*}} Workgroup
+// CHECK: OpUndef
+// CHECK-NOT: OpVariable {{.*}} Workgroup %
+// CHECK: OpVariable {{.*}} Workgroup
+// CHECK-NOT: OpVariable {{.*}} Workgroup %
+// CHECK: OpVariable {{.*}} Workgroup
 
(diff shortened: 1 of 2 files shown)


test/global-variable-no-undef-initializer.cl
```diff
@@ -0,0 +1,31 @@
+// RUN: clspv  %s -o %t.spv
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
+// CHECK-DAG: %[[__original_id_29:[0-9]+]] = OpVariable {{.*}} Workgroup
```
alan-baker: Can these tests be strengthened by ensuring the undef is not present? Maybe something like the following?

```
// CHECK-NOT: OpUndef
// CHECK-NOT: OpVariable {{.*}} Workgroup %
// CHECK: OpVariable {{.*}} Workgroup
// CHECK-NOT: OpVariable {{.*}} Workgroup %
// CHECK: OpVariable {{.*}} Workgroup
// CHECK-NOT: OpVariable {{.*}} Workgroup %
```

kpet: Good idea but the `OpUndef` is expected to be present. This is what the test is about: the case where an `OpUndef` of the same type as the `Workgroup` `OpVariable` happens to be present and thus the now-removed dont-emit-the-initializer-if-the-vmap-lookup-fails logic doesn't fail leading to the undesired initializer.

Looking at the IR, the reason the undef is needed is that the call to `get_local_id` is omitted on the first iteration of the loop (the sub-expression from the calculation of `last_thread` is reused). The undef is used in the phi that produces `idx` after the last iteration of the loop.

PR #421, opened 2019-08-27
Use unified1 version of spirv.h

dneto0:
Output targets SPIR-V 1.0 always.


dneto0:
> Did you look into the enum and string generation? Both of those in the cmake directory use the 1.0 specific headers.

I did not.  I'll add that.

alan-baker:
Closed in favor of #534

Issue #435, opened 2019-11-08
Ingest OpenCL-flavored SPIR-V


fodinabor:
Would it be possible to feed clspv OpenCL-flavored SPIR-V and get Vulkan-flavored SPIR-V out of it?
If not, what would have to be done?

Using the SPIR-V Cross tools it might be possible to get valid LLVM IR, that's similar to what clspv uses somewhere internally?
Or are all the Khronos SPIR-V tools mostly Vulkan-flavor focused?

kpet:
This is not something that clspv supports out-of-the-box but you can try to do it (this is not going to work for all SPIR-V modules) using the [SPIRV-LLVM Translator](https://github.com/KhronosGroup/SPIRV-LLVM-Translator):

```
llvm-spirv -r -o opencl-ir.bc opencl-spirv.spv
clspv -x ir -o vulkan-spirv.spv opencl-ir.bc
```

Unfortunately, the master branches of clspv and the translator aren't always compatible at any one time (and definitely not as I type this). Both clspv and the translator are used by [clvk](https://github.com/kpet/clvk) (disclaimer: I'm the author) where they need to be compatible so you can try to build clvk to get a compatible version of both tools.

Out of curiosity, what OpenCL SPIR-V producer are you trying to convert the output of? (Depending on what it is, there might be a better path.)

fodinabor:
Yeah, just thought it would be awesome if it was possible to ingest OpenCL SPIR-V directly.

I'm currently playing around with https://github.com/intel/llvm trying to check-out whether it would be possible to run SYCL on Vulkan ;-)  
It's possible to generate LLVM-IR from it using `-fsycl-device-only`, but that seemed incompatible with clspv. I thought SPIR-V would be more stable.

But maybe I should try again to use the intel/llvm as LLVM code base for clspv, so that those to be compatible.

kpet:
>I'm currently playing around with https://github.com/intel/llvm trying to check-out whether it would be possible to run SYCL on Vulkan ;-)

I've been looking at this as well. That's one of the reasons I added prototype support for cl_khr_il_program to clvk. The SPIR-V path is likely to be easier and more stable than trying to pass LLVM IR.

I haven't tried to get it to work end-to-end yet but all the building blocks should now be there. FWIW, I know someone who's run a few examples from another SYCL implementation on Vulkan using clvk/clspv already so it's definitely possible for basic programs.

fodinabor:
[...]

PR #436, opened 2019-11-14
Update LLVM

kpet:
...to pick up a fix that makes it possible to use add_llvm_tool
when LLVM_TARGETS_TO_BUILD is set to an empty list like clspv does.

This allows to fix the clvk issue mentioned on #435.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>


kpet:
@googlebot I signed it!

kpet:
#437 fixes the issue.

PR #437, opened 2019-11-18
Update LLVM dep

alan-baker:
* Fix compile

@kpet I'm going to update to here to fix some internal builds (if the bots pass) before updating to the latest llvm. This commit is from Nov 15.


kpet:
@alan-baker The commit you picked is after my fix so I will abandon #436 if you're getting this in.

PR #449, opened 2019-12-03
Fix generation of selection merges

### LINKED ISSUE #448: fails new SPIR-V validation when compiling test/global-variable-no-undef-initializer.cl
When updating SPIRV-Tools, I get a validation error on one of the tests, test/global-variable-no-undef-initializer.cl  :

error: line 81: Header block 55[%55] is contained in the loop construct headed by 51[%51], but it's merge block 61[%61] is not
  %55 = OpLabel

alan-baker:
Fixes #448

* Refactor structured order computation into a new file
* updates deps
* Modify SPIRVProducer pass to figure out merge and continues at the
start of the pass
  * this is based on traversing in structured order
  * do not generate selection merges if one side of the branch is a loop
  continue or merge
* add a new test

The SPIRV-Tools update also includes the new validator check.

ComputeStructuredOrder is effectively a move from ReorderBasicBlock::StructuredOrder.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -52,6 +52,7 @@
 
 #include "ArgKind.h"
 #include "Builtins.h"
+#include "ComputeStructuredOrder.h"
 #include "ConstantEmitter.h"
 #include "Constants.h"
 #include "DescriptorCounter.h"
@@ -372,6 +373,9 @@ struct SPIRVProducerPass final : public ModulePass {
   // Populate UBO remapped type maps.
   void PopulateUBOTypeMaps(Module &module);
 
+  // Populate the merge and continue block maps.
+  void PopulateStructuredCFGMaps(Module &module);
+
   // Wrapped methods of DataLayout accessors. If |type| was remapped for UBOs,
   // uses the internal map, otherwise it falls back on the data layout.
   uint64_t GetTypeSizeInBits(Type *type, const DataLayout &DL);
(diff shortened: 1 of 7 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -4768,42 +4778,7 @@ void SPIRVProducerPass::HandleDeferredInstruction() {
           getAnalysis<LoopInfoWrapperPass>(*Func).getLoopInfo();
 
       BasicBlock *BrBB = Br->getParent();
-      Loop *L = LI.getLoopFor(BrBB);
```
alan-baker: This effectively moved to PopulateStructuredCFGMaps below.

lib/SPIRVProducerPass.cpp
```diff
@@ -5915,3 +5878,87 @@ bool SPIRVProducerPass::CalledWithCoherentResource(Argument &Arg) {
...
+          BasicBlock *Header = L->getHeader();
+          BasicBlock *Latch = L->getLoopLatch();
+          for (auto *loop_block : L->blocks()) {
+            if (loop_block == Header) {
+              continue;
+            }
+
+            // Check whether block dominates block with back-edge.
+            if (DT.dominates(loop_block, Latch)) {
+              ContinueBB = loop_block;
```
dneto0: Also need the backedge block to post-dominate the continue target.
How do we know that will be satisfied as well?

dneto0: Hmm. Seems like this is a consequence of the StructurizeCFG algorithm?
Needs a comment at least.

lib/SPIRVProducerPass.cpp
```diff
@@ -4768,42 +4778,7 @@ void SPIRVProducerPass::HandleDeferredInstruction() {
           getAnalysis<LoopInfoWrapperPass>(*Func).getLoopInfo();
```
dneto0: You no longer need these three variables: Func, DT, LI

PR #450, opened 2019-12-03
Add empty script for clvk bot

jrprice:
Will populate script once bot is in place (so it can be tested properly before landing). Also waiting for one more Clvk change to land.


kokoro/clvk-linux-clang-debug/build.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Copyright 2018 The Clspv Authors. All rights reserved.
```
dneto0: May as well make this 2019

jrprice: Done.

PR #452, opened 2019-12-04
Add Clvk kokoro bot

jrprice:


jrprice:
Finally got the bot to pass, so this should be ready for review now.

Summary of components and workarounds:
- Builds Vulkan-Headers, Vulkan-Loader, SwiftShader, and then clvk (which builds Clspv)
    - Using ToT for all of these, which might be too many points of failure? Could version lock the Vulkan stuff and/or SwiftShader if that's desirable.
- Needed newer CMake for Vulkan-Loader.
- Needed newer libstdc++ for clvk (which affects everything else, but hopefully not an issue).
- Using gold as linker for SwiftShader due to issue with the provided version of GNU ld.

alan-baker:
I would consider version locking the vulkan headers and loader until we need newer functionality from them.

Also, are you planning subsequent PRs to run different tests with clvk? The simple doesn't cover too much.

jrprice:
> I would consider version locking the vulkan headers and loader until we need newer functionality from them.

Done.
 
> Also, are you planning subsequent PRs to run different tests with clvk? The simple doesn't cover too much.

Yes. clvk only has one other test executable which doesn't succeed on SwiftShader right now due to lack of `StorageImageWriteWithoutFormat`. I have plans to add more tests from external sources that give better coverage in the near future (we should sync offline about this).

Issue #451, opened 2019-12-04
Remove necessity of sampler map


alan-baker:
Currently to support literal samplers, clspv requires a sampler map to be specified. Since the intervention of the descriptor map this is no longer necessary.

Support for the sampler map will be removed in the following steps:
1. Deprecate the option
  a. Issue a warning on its use
  b. Add support for generating literal samplers without the sampler map
2. (Later) remove the option altogether

The internal representation of samplers will use the data mask instead of the index into the sampler map.

PR #453, opened 2019-12-04
Deprecate the sampler map

alan-baker:
Contributes to #451

* Add a warning when a sampler map is provided (for both online and
offline compiles)
  * the sampler is still supported and works as before if it is provided
* Without a sampler map clspv now auto-generates bindings for literal
samplers based on the sampler literal
* updated sampler map tests to also test without a sampler map
* new test to verify consistency with the descriptor map

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -23,8 +23,8 @@ The following subjects are covered:
 The sampler map is deprecated. It is no longer necessary to specify a sampler
 map for literal samplers. Refer to the bindings generated for the literal
 samplers in descriptor map. The -samplermap option is still accepted, but
-support will be remove at later date. A single sampler binding is generated for
-each unique literal sampler in the program.
+support will be removed at later date. A single sampler binding is generated
+for each unique literal sampler in the program.
 
 ## SPIR-V Features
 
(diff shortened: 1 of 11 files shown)


docs/OpenCLCOnVulkan.md
```diff
@@ -18,6 +18,14 @@ The following subjects are covered:
 - Restrictions on the OpenCL C language as is to be consumed by a Vulkan
   implementation.
 
+## Deprecation Notice
+
+The sampler map is deprecated. It is no longer necessary to specify a sampler
+map for literal samplers. Refer to the bindings generated for the literal
+samplers in descriptor map. The -samplermap option is still accepted, but
+support will be remove at later date. A single sampler binding is generated for
```
kpet: s/remove/removed/

lib/AllocateDescriptorsPass.cpp
```diff
@@ -247,55 +259,71 @@ bool AllocateDescriptorsPass::AllocateLiteralSamplerDescriptors(Module &M) {
...
-          if (!const_val) {
-            call->getArgOperand(0)->print(errs());
-            llvm_unreachable(
-                "Argument of sampler initializer was non-constant!");
-          }
+  // Now replace calls to __translate_sampler_initializer
+  if (init_fn) {
+    // Copy users, to avoid modifying the list in place.
+    SmallVector<User *, 8> users(init_fn->users());
+    if (!users.empty()) {
```
dneto0: This if is non-sensical. Please delete.

PR #457, opened 2019-12-07
Support sampled read_image{f,i,ui} with integer coordinates for 2D/3D images

kpet:
Also update the docs to reflect recent changes to image support.

AFAICT the _Z11read_imagef14ocl_image2d_ro11ocl_samplerDv4_i built-in
function I've removed doesn't exist in the OpenCL 1.2 spec.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -204,7 +204,7 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
   bool replaceVstoreHalf(Module &M);
   bool replaceVstoreHalf2(Module &M);
   bool replaceVstoreHalf4(Module &M);
-  bool replaceReadImageF(Module &M);
+  bool replaceSampledReadImageWithIntCoords(Module &M);
   bool replaceAtomics(Module &M);
   bool replaceCross(Module &M);
   bool replaceFract(Module &M);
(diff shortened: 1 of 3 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2682,14 +2682,26 @@ bool ReplaceOpenCLBuiltinPass::replaceVstoreHalf4(Module &M) {
...
   const std::map<const char *, const char *> Map = {
+      // TODO 1D, 1Darray
+      // 2D
+      {"_Z11read_imagei14ocl_image2d_ro11ocl_samplerDv2_i",
+       "_Z11read_imagei14ocl_image2d_ro11ocl_samplerDv2_f"},
+      {"_Z12read_imageui14ocl_image2d_ro11ocl_samplerDv2_i",
+       "_Z12read_imageui14ocl_image2d_ro11ocl_samplerDv2_f"},
       {"_Z11read_imagef14ocl_image2d_ro11ocl_samplerDv2_i",
        "_Z11read_imagef14ocl_image2d_ro11ocl_samplerDv2_f"},
-      {"_Z11read_imagef14ocl_image2d_ro11ocl_samplerDv4_i",
```
alan-baker: This should correspond to a 2D array image read. It's ok to leave this as a todo.

kpet: Apparently not. A quick test suggests that Clang mangles these to `ocl_image2d_array_ro` which matches my reading of https://github.com/llvm/llvm-project/blob/master/clang/lib/AST/ItaniumMangle.cpp#L2682. We'll find out soon enough if it did correspond to an existing built-in :).

PR #465, opened 2019-12-11
Unsampled reads

### LINKED ISSUE #464: Support unsampled image reads
Couple ways this could be done:
1. Generate sampled reads with a newly created literal sampler (address mode none, nearest filtering and unnormalized coords).
2. Generate OpImageRead

The drawback of the first option is that there will be an unexpected sampler in the descriptor map. This isn't too bad, but users need to be aware.

The drawbacks for the second option are:
* If an image is used with and without samplers in a kernel, clspv would need to generate multiple binding because OpTypeImage can only be either sampled or unsampled in Vulkan.
* clspv would have to generate StorageImageReadWithoutFormat
* image specialization couldn't work as is (ties back in to multiple variables)

My preference is to generate the extra sampler and treat all reads as sampled.

alan-baker:
Fixes #464

* Support unsampled image reads for 1D, 2D and 3D images
  * calls are translated to sampled reads with a literal sampler using
  CLK_ADDRESS_NONE | CLK_FILTER_NEAREST | CLK_NORMALIZED_COORDS_FALSE
* Add tests
* Refactor direct uses of string literal for sampler initializer with a function

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -29,6 +29,7 @@
 
 #include "spirv/1.0/spirv.hpp"
 
+#include "clspv/AddressSpace.h"
 #include "clspv/Option.h"
 
 #include "Constants.h"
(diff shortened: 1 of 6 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2683,6 +2687,105 @@ bool ReplaceOpenCLBuiltinPass::replaceVstoreHalf4(Module &M) {
...
+          // The coordinate.
+          auto Coord = CI->getOperand(1);
+
+          // Create the sampler translation function if necessary.
+          if (!translate_sampler) {
+            // Create the sampler type if necessary.
+            if (!sampler_type) {
+              sampler_type =
+                  StructType::create(M.getContext(), "opencl.sampler_t");
+              sampler_type = sampler_type->getPointerTo(2);
```
dneto0: meh. 2 is constant address space, right...

alan-baker: Good point. Fixed.

PR #470, opened 2019-12-18
Refactor checks about pointers in structs

### LINKED ISSUE #468: StructContainsPointer diagnostic is too restrictive
This frontend check prevents us from passing structures that contain pointers to device-side functions, whereas we should only need to apply this restriction to kernel functions. This functionality is needed for kernels in the Arm Compute Library.

https://github.com/google/clspv/blob/990e9b91a8ebfae3629b4db867c6909580ffcd51/lib/FrontendPlugin.cpp#L103-L111

Commenting out this check allows Clspv to generate correct code for (some of) the Arm Compute Library kernels, although we also need to inline such functions for this to work at present (`-inline-entry-points`).

alan-baker:
Fixes #468

* Move the check about pointers in structs to only happen for kernel
arguments
* Add a new check to prevent recursive struct definitions
* New and updated tests

### DIFF
--- a/lib/FrontendPlugin.cpp
+++ b/lib/FrontendPlugin.cpp
@@ -93,7 +93,7 @@ private:
           return true;
       }
     }
- 
+
     seen->erase(canonical.getTypePtr());
     return false;
   }
--- a/lib/FrontendPlugin.cpp
+++ b/lib/FrontendPlugin.cpp
@@ -55,6 +55,7 @@ private:
     CustomDiagnosticSSBOUnalignedStruct = 15,
     CustomDiagnosticOverloadedKernel = 16,
     CustomDiagnosticStructContainsPointer = 17,
+    CustomDiagnosticRecursiveStruct = 18,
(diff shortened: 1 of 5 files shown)


alan-baker:
@jrprice I think this should do the trick without being too lax, can you give it a try?

jrprice:
Yes, this works for me, thanks.

Maybe worth adding tests that cover the cases we want to work? I can only see tests that cover the error cases at the moment. The two that would be relevant for the Arm Compute Library are:

```
struct T {
  global int* ptr;
};

void bar(struct T* t) {
  *t->ptr = 42;
}

kernel void foo(global int* out) {
  struct T t;
  t.ptr = out;
  bar(&t);
}
```

and

```
struct T {
  global int* ptr;
};

struct T bar(global int* out) {
  struct T t;
  t.ptr = out;
  return t;
}

kernel void foo(global int* out) {
  struct T t = bar(out);
  t.ptr = 42;
}
```

PR #475, opened 2019-12-20
Use unique_ptr instead of shared_ptr

dneto0:


dneto0:
Trades one violation of the Google C++ style guide (use of shared_ptr) for another (non-const ref arg for operands list to SPIRVInstruction constructor).  To resolve the latter, I could change the parameter to pointer-to-SPIRVOperandList, but that would be a much bigger change at this point.  I wanted eyes on this one first.

Issue #455, opened 2019-12-06
Missing builtin? _Z11read_imageh


alex-xia-xia:
Can't translate function call.  Missing builtin? _Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i in:   %call38 = tail call spir_func <4 x half> @_Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i(%opencl.image2d_ro_t.float.sampled addrspace(1)* %1, %opencl.sampler_t addrspace(2)* %32, <2 x i32> %33) #6
Can't translate function call.  Missing builtin? _Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i in:   %call33 = tail call spir_func <4 x half> @_Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i(%opencl.image2d_ro_t.float.sampled addrspace(1)* %1, %opencl.sampler_t addrspace(2)* %30, <2 x i32> %31) #6
Can't translate function call.  Missing builtin? _Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i in:   %call28 = tail call spir_func <4 x half> @_Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i(%opencl.image2d_ro_t.float.sampled addrspace(1)* %1, %opencl.sampler_t addrspace(2)* %28, <2 x i32> %29) #6
Can't translate function call.  Missing builtin? _Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i in:   %call23 = tail call spir_func <4 x half> @_Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i(%opencl.image2d_ro_t.float.sampled addrspace(1)* %1, %opencl.sampler_t addrspace(2)* %26, <2 x i32> %27) #6
Can't translate function call.  Missing builtin? _Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i in:   %call18 = tail call spir_func <4 x half> @_Z11read_imageh14ocl_image2d_ro11ocl_samplerDv2_i(%opencl.image2d_ro_t.float.sampled addrspace(1)* %0, %opencl.sampler_t addrspace(2)* %24, <2 x i32> %25) #6
[...]

PR #476, opened 2019-12-20
Add support for half image builtins

alan-baker:
Contributes to #455

* Add support for sampled and unsampled reads for 1D, 2D, and 3D images
  * converts to equivalent float intrinsic and converts the result
  * this means unsampled reads are converted to sampled reads and
  integer coordinates are converted to floating-point coordinates
* Add support for half image writes for 1D, 2D and 3D images
  * data is converted to float and the corresponding float version is
  used
* Added test


alan-baker:
Conversion to half is necessary because Vulkan only allows 32-bit sampled types for images.

PR #483, opened 2020-01-05
Fix warning in fetch_sources.py

kpet:
Newer versions of python report the following:

utils/fetch_sources.py:87: SyntaxWarning: "is" with a literal. Did you mean "=="?


kpet:
@alan-baker I've fixed the issue with the clvk bot (missing SHA in my branch of the SPIRV-LLVM-Translator), sorry for that.

PR #484, opened 2020-01-05
Add support for -cl-std option

kpet:
Match what Clang does except that:
- lowercase aliases of the language names are not supported
  (they are not described in the OpenCL specification)
- the default standard is OpenCL C 1.2

Remove -c++ (replaced by -cl-std=CLC++).

Turn on address space inferencing when compiling as OpenCL C 2.0.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -407,14 +407,25 @@ int SetCompilerInstanceOptions(CompilerInstance &instance,
   }
 
   clang::LangStandard::Kind standard;
-  if (clspv::Option::CPlusPlus()) {
-    standard = clang::LangStandard::lang_openclcpp;
-  } else {
+  switch (clspv::Option::Language()) {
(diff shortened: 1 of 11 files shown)


lib/Compiler.cpp
```diff
@@ -407,15 +407,26 @@ int SetCompilerInstanceOptions(CompilerInstance &instance,
   }
 
   clang::LangStandard::Kind standard;
```
alan-baker: Could this enum be used directly in the option? On the one hand having a separate enum might be interesting for defining a clspv-specific flavor of opencl, but all those extra checks are in the frontend plugin and clang wouldn't know about it. So it might be better to reuse clang's enum if possible and change the default case to be an error for an unsupported kind.

kpet: Yes, but that would mean effectively pulling Clang as a dependency to the passes library which I didn't feel comfortable with.

PR #485, opened 2020-01-06
Fix loop merge and continue as same block

### LINKED ISSUE #482: Unhandled instruction feeding switch
Mace kernel depth_to_space.cl.

Compile command:
```
clspv -w -I . depth_to_space.cl -DDATA_TYPE=float -DCMD_DATA_TYPE=f -DDEPTH1 -DDEPTH3
```
Leads to the follows assertion:
```
Unhandled instruction feeding switch   %trunc = sub nuw i3 -4, %14
Illegal instruction: 4
```
Only occurs with -DDEPTH3 specified.

alan-baker:
Fixes #482

* New pass fixes up after structurize cfg
  * splits blocks that are both a loop merge block and a loop continue
  block

The new pass caused some cfgs to change unrelated to the pass itself (happens if pass returns false and does nothing). I rewrote some of the signed compare fixup tests to make them simpler as a result.


alan-baker:
@kpet does this address the CLBlast issues?

kpet:
@alan-baker I'll test tonight/tomorrow and comment here.

kpet:
I can confirm this fixes the issues with CLBlast.

PR #498, opened 2020-01-15
Don't add int64 types due to literal operands

### LINKED ISSUE #497: clspv unnecessarily generate Int64 capability due to literal values
Instructions like insertelement, extractelement, etc, are translated to OpCompositeExtract which uses literal values for indices. Sometimes LLVM IR uses i64 for the constant values and clspv generates an unused `OpTypeInt 64 0` as a result (which leads to `OpCapability Int64`). This should be avoided.

Found in the mace file buffer_to_image.cl.

alan-baker:
Fixes #497

* Do not map types for operands that end up as literal values in SPIR-V
* Add a test to ensure Int64 capability is not added

I believe the i64 come from some sort canonicalization LLVM performs.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -1255,6 +1255,39 @@ void SPIRVProducerPass::FindTypePerFunc(Function &F) {
         continue;
       }
 
+      // #497: InsertValue and ExtractValue map to OpCompositeExtract which
+      // takes literal values for indices. As a result don't map the type of
+      // indices.
+      if (I.getOpcode() == Instruction::ExtractValue) {
+        FindType(I.getOperand(0)->getType());
+        continue;
+      }
+      if (I.getOpcode() == Instruction::InsertValue) {
+        FindType(I.getOperand(0)->getType());
+        FindType(I.getOperand(1)->getType());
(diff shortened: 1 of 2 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -1255,6 +1255,39 @@ void SPIRVProducerPass::FindTypePerFunc(Function &F) {
         continue;
       }
 
+      // #497: InsertValue and ExtractValue map to OpCompositeExtract which
```
dneto0: map to OpCompositeInsert and OpCompositeExtract....

lib/SPIRVProducerPass.cpp
```diff
@@ -1255,6 +1255,39 @@ void SPIRVProducerPass::FindTypePerFunc(Function &F) {
...
+      // takes literal values for indices. As a result don't map the type of
+      // indices.
```
dneto0: you don't have to map the type of the indices because the indices are always small, by construction (various universal limits)

alan-baker: The indices are not ids in OpCompositeInsert and OpCompositeExtract, they are literals.

Issue #102, opened 2017-10-23
Unhandled Constant case in SPIRVProducer


dneto0:
On this input:

```
constant uint scalar = 42;

constant float arr[3] = { 1.0f, 2.0f, 3.0f };

typedef struct {
  float4 u;
  float v;
} S;

constant S structval[2] = {
    {(float4)(10.5f, 11.5f, 12.5f, 13.5f), 14.5f},
    {(float4)(20.5f, 21.5f, 22.5f, 23.5f), 24.5f},
};

// Same data as arr.  Should reuse the same underlying space as arr
constant float arr2[3] = { 1.0f, 2.0f, 3.0f };

void kernel foo(global float *A, uint n)
{
  *A = arr[n] + arr[3-n] + structval[n].u.y + structval[n].v;
}
```

I get: 

clspv: /build/dneto/aglet/SPIRV-LLVM/third_party/llvm/lib/IR/User.cpp:26: void llvm::User::replaceUsesOfWith(llvm::Value*, llvm::Value*): Assertion `(!isa<Constant>(this) || isa<GlobalValue>(this)) && "Cannot call User::replaceUsesOfWith on a constant!"' failed.

dneto0:
Occurs even if the body is only:
  A[3] = structval[n].v;

dneto0:
The problem occurs in SPIRVProducer, when changing a __constant variable address space to module-scope Private.

Trouble occurs with this instruction:
 %v = getelementptr inbounds [2 x %struct.S], [2 x %struct.S] addrspace(2)* bitcast (<{ { <4 x float>, float }, { <4 x float>, float } }> addrspace(2)* @structval to [2 x %struct.S] addrspace(2)*), i32 0, i32 %n, i32 1

The code is trying to convert the use of @structval to pointer-to-private, but you can't just replace the use in an LLVM-constant expression.

rjodinchr:
Gemini:
> Constant struct arrays containing vector fields (e.g., constant S structval[2] with float4) compile cleanly without compiler crashes and pass spirv-val.

Please reopen if this is still relevant

PR #500, opened 2020-01-20
Fix issue 102

AlexVlx:
This fixes the issue identified in #102, and starts to deal with `constexpr` bitcasts (fully addressing the generic case of `constexpr` expresions present in e.g. the initialisers of constants will require additional work). The basic idea is to use the potentially unsafe modification to replace an existing global with the new global that lives in `ModuleScopePrivate`. This is OK since even though the types are different, they are equivalent. We need special handling for bitcasts to correctly propagate the new global, otherwise it is possible to observe mismatched address spaces in a bitcast, which is ill-formed.


test/ProgramScopeConstants/issue_102.cl
```diff
@@ -0,0 +1,29 @@
+// https://github.com/google/clspv/issues/102
+
+// RUN: clspv -cl-std=CLC++ -inline-entry-points %s -o %t.spv
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
```
alan-baker: Can you please add some check statements that we generate the constants correctly?

test/ProgramScopeConstants/issue_102.cl
```diff
@@ -0,0 +1,29 @@
...
+    };
+
+    // Same data as arr.  Should reuse the same underlying space as arr
+    constant float arr2[3] = { 1.0f, 2.0f, 3.0f };
+}
+
+void kernel foo(global float *A, uint n)
+{
+  *A = arr[n] + arr[3-n] + structval[n].u.y + structval[n].v;
+}
```
alan-baker: Add a new line.

AlexVlx:
> Can you please add some check statements that we generate the constants correctly? 

Done - I believe that these cover all the interesting cases, but do let me know if you disagree.

AlexVlx:
> Add a new line. 

Done.

AlexVlx:
@alan-baker thank you for the review. I would like to point out two aspects I should follow-up on, which build on / are a consequence of this:
- we probably have to handle GEPs just like we handle BitCasts, since they'd run into the same address-space mismatch;
- this will generate quite a few BitCasts, many of which can be merged since they are identical - since we insert them late, at SPIR-V generation time, they won't get cleaned up, but I expect any optimiser worth its salt to do it, but it can be fixed subsequently.

test/ProgramScopeConstants/issue_102.cl
```diff
@@ -0,0 +1,71 @@
...
+// CHECK:  [[_uint_1:%[0-9a-zA-Z_]+]] = OpConstant [[_uint]] 1
+// CHECK:  [[_float_1:%[0-9a-zA-Z_]+]] = OpConstant [[_float]] 1
+// CHECK:  [[_float_2:%[0-9a-zA-Z_]+]] = OpConstant [[_float]] 2
+// CHECK:  [[_float_3:%[0-9a-zA-Z_]+]] = OpConstant [[_float]] 3
+// CHECK:  [[_u:%[0-9a-zA-Z_]+]] = OpConstantComposite [[__arr_float_uint_3]] [[_float_1]] [[_float_2]] [[_float_3]]
+// CHECK:  [[_pu:%[0-9a-zA-Z_]+]] = OpVariable [[_ptr_Private__arr_float_uint_3]] Private [[_u]]
+// CHECK:  [[_pt:%[0-9a-zA-Z_]+]] = OpVariable [[_ptr_Private__arr__struct_28_uint_2_0]] Private [[_t]]
+// CHECK:  [[_pf:%[0-9a-zA-Z_]+]] = OpAccessChain {{.*}} [[_pu]]
+// CHECK:  [[_f:%[0-9a-zA-Z_]+]] = OpLoad [[_float]] [[_pf]]
+// CHECK:  [[_bc:%[0-9a-zA-Z_]+]] = OpBitcast [[_ptr_private__arr__struct_20_uint_2]] [[_pt]]
```
alan-baker: Ok I guess this wasn't immediately obvious to me previously, but this is invalid SPIR-V. Shaders don't support this type of bitcast. It is unfortunate that the validator is not catching this, but I think https://github.com/KhronosGroup/SPIRV-Tools/pull/3091 should flag this as invalid.

This is a case that should be handled by ReplacePointerBitcastsPass. Instead of casting the variable, clspv would have to load 4 uints, make a vector and bitcast that vector to a float vector.

AlexVlx: Bummer. For what it's worth, the SPIR-V spec reads a bit confusingly (IMHO) here. Looking at OpBitcast from here: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.pdf does not convey any such limitation. Furthermore, there are only 5 references to OpBitcast in the entire document, neither of which makes reference to any such limitation being in place. Yes, 2.2.2 is normative in what constitutes a valid source of a variable pointer, but I'd have hoped it'd be reflected under OpBitcast too.  

Granted, it appears to be normative in 2.16.1, and that outlines a more profound problem in that "A variable pointer must point to one of the following storage classes: StorageBuffer, Workgroup [...]", which is broken orthogonally to the bitcast, I think. So, TL;DR, hoisting the bitcast replacement / insertion before running ReplacePointerBitcastsPass wouldn't be a problem, but this latter aspect appears a bit more involved (to me), since it hints that mapping `__constant` to `Private` might not be a generic solution. Should I handle that separately?

PR #508, opened 2020-02-15
Added PKGBUILD file for Arch-based distro packaging

VileLasagna:
The PKGBUILD file makes its own git clone and builds from there so it's independent of working tree changes.
This also means that once an official release is out, this file can be easily tweaked to be a base for either an official package in Arch itself (or one of the derivatives, such as Manjaro) or in the AUR

### DIFF
--- a/package/arch/PKGBUILD
+++ b/package/arch/PKGBUILD
@@ -1,0 +1,47 @@
+
+# Contributor: Jehferson Mello <vilelasagna@gmail.com>
+pkgname=clspv-git
+pkgver=r339.51952280
+pkgrel=1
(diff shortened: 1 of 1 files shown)


VileLasagna:
At this point there is not an official release for clspv. As such, it doesn't make much sense for any distro to have their own packaging at this point.
Having a package in clspv's source tree gives users who want to evaluate and start working with clspv an easy path to set it up, as well as enables clspv to, should the wish arise in the future, have its own set of packaged releases, making it not "dependent" on distro maintainer support. It's a low maintenance file (only really would be tweaked once tagged releases start happening) that adds a great deal of convenience for users

Issue #510, opened 2020-02-20
Builtin cleanup


sjw36:
FYI, I'm working on an overhaul of the builtin support, starting with the giant case statement in SPIRVProducerPass.cpp. Then on to ReplaceOpenCLBuiltinPass.cpp.

Only parsing a mangled name once, then use an unordered_map lookup.

PR #513, opened 2020-02-27
Builtins::Lookup interface

sjw36:
* Moved demangler from ReplaceOpenCLBuiltinPass.cpp
* Added enum for all cl_khr_sub_groups
* Implemented sub_group_broadcast
* Added --spirv-version=1.x for features that require later SPIRV version (ie. sub_groups)
* Links to https://github.com/google/clspv/issues/510


sjw36:
I know this is a big change but hopefully easier to maintain and extend builtin support.
Next steps are:
- replace remaining mangled name lookups
- full sub_group support
- OpenCL 2.0 feature support
And of course more tests

lib/Builtins.h
```diff
@@ -18,115 +18,160 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/IR/Function.h"
 
+#include "BuiltinsGen.h"
+
+#include <string>
```
dneto0: Please put standard includes before project includes.
(following https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

lib/Builtins.h
```diff
@@ -18,115 +18,160 @@
...
-}
-
-// Returns true if the function is an OpenCL image query.
-bool IsImageQuery(llvm::StringRef name);
-inline bool IsImageQuery(llvm::Function *f) {
-  return IsImageQuery(f->getName());
-}
+  namespace Builtins {
+
+    struct ParamTypeInfo {
```
dneto0: Let clang-format format the code.

lib/Builtins.h
```diff
@@ -18,115 +18,160 @@
...
+      bool                     sign               = false;                    // is signed
+      llvm::Type::TypeID       type               = llvm::Type::VoidTyID;     // elem type
```
dneto0: please rename this to type_id

lib/Builtins.h
```diff
@@ -18,115 +18,160 @@
...
+      int                      blen               = 0;                        // byte len
```
dneto0: Don't abbreviate as much:
 I'd like to see these renamed to  byte_len and vector_size (and then comment the latter that it's the number of elements)

lib/Builtins.h
```diff
@@ -18,115 +18,160 @@
...
+      int                      vsiz               = 0;                        // vector size
+      std::string              name;                                          // struct name
+    };
+
+    class FunctionInfo {
+      bool                        m_is_valid    = false;
```
dneto0: Please don't use the m_ convention for member names.
Either use LLVM convention or Chromium/Google convention (trailing underscore).

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
 
 #include "Builtins.h"
 
+#include <unordered_map>
+#include <stdlib.h>
+
 using namespace llvm;
+using namespace clspv;
+
+////////////////////////////////////////////////////////////////////////////////
```
dneto0: Hide these in the anonymous namespace.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+// utilities for const char* map keys
+struct cstr_hash {
+  std::size_t operator()(const char *cstr) const {
+    std::size_t hash = 5381;
```
dneto0: Where did you get this hash function?  Why this hash function?

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+    for ( ; *cstr != '\0' ; ++cstr)
+      hash = (hash * 33) + *cstr;
+    return hash;
+  }
+};
+struct cstr_eq {
+  bool operator()(const char* a, const char *b) const { return strcmp(a, b) == 0; }
```
dneto0: Add a method comment:  // Returns true if the two C-style strings are equal.  Assumes they are both null-terminated.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+#include <unordered_map>
+#include <stdlib.h>
+
 using namespace llvm;
+using namespace clspv;
+
+////////////////////////////////////////////////////////////////////////////////
+// utilities for const char* map keys
+struct cstr_hash {
+  std::size_t operator()(const char *cstr) const {
```
dneto0: Add a method comment.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+  }
+};
+struct cstr_eq {
+  bool operator()(const char* a, const char *b) const { return strcmp(a, b) == 0; }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+////  Convert Builtin function name to a Type enum
+////////////////////////////////////////////////////////////////////////////////
+static Builtins::EBuiltinType GetBuiltinType(const std::string& _name) {
```
dneto0: I prefer using an anonymous namespace block rather than scattering "static" in many places.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+  static std::unordered_map<const char*, Builtins::EBuiltinType, cstr_hash, cstr_eq> s_func_map = {
```
dneto0: Thanks for not making this initialized at program startup!  :-)

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
 }
 
-bool clspv::IsImageQuery(StringRef name) {
-  return clspv::IsGetImageHeight(name) || clspv::IsGetImageWidth(name) ||
-         clspv::IsGetImageDepth(name) || clspv::IsGetImageDim(name);
+////////////////////////////////////////////////////////////////////////////////
+////  FunctionInfo
+////////////////////////////////////////////////////////////////////////////////
+
+static bool GetUnmangledName(const std::string& str, size_t& pos, std::string& name) {
```
dneto0: Please add method comments.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+    return false;
+  }
+  pos = end - str.data();
+
+  name = str.substr(pos, nameLen);
+  pos += nameLen;
+  return true;
 }
+
+static bool GetParameterType(clspv::Builtins::ParamTypeInfo& ti, const std::string& name, size_t& pos) {
```
dneto0: Reference parameters should be const. 
(Except in very rare circumstances).  https://google.github.io/styleguide/cppguide.html#Reference_Arguments

At the call site, it's not obvious that pos is being modified.
I don't know about "ti" either, and at this point of reading the text, it's not clear.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+  if (!nameLen) {
+    return false;
+  }
+  pos = end - str.data();
+
+  name = str.substr(pos, nameLen);
+  pos += nameLen;
+  return true;
 }
+
```
dneto0: Please add a method comment.
Particularly because from the implementation it *looks* like it's decoding a mangled type name, but presumably from Itanium ABI rules, and possibly with OpenCL extensions?  I don't know enough here to tell what this is supposed to do, or the code is correct.

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+    pos--;
+    if (!GetUnmangledName(name, pos, ti.name)) {
+      return false;
+    }
+    break;
+  }
+  case '.':
+    return false;
+  default:
+#ifdef DEBUG
```
dneto0: I prefer an llvm_unreachable here rather than assert(0).
I want this to trigger in a release build as well.   Don't make it conditional on DEBUG

lib/Builtins.cpp
```diff
@@ -14,171 +14,339 @@
...
+    printf("Func: %s\n", name.c_str());
+    assert(0);
+#endif
+    return false;
+  }
+  return true;
+}
+
+bool Builtins::FunctionInfo::GetFromMangledNameCheck(const std::string& name) {
```
dneto0: Needs a method comment.

lib/BuiltinsGen.cpp
```diff
@@ -0,0 +1,1083 @@
+
+    // Async
```
dneto0: You can protect the formatting of this with comments like:

     // clang-format off

...

      // clang-format on

lib/BuiltinsGen.h
```diff
@@ -0,0 +1,318 @@
+namespace clspv {
+    namespace Builtins {
```
dneto0: Let clang-format format this enum.

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -27,117 +27,23 @@
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
-#include "spirv/1.0/spirv.hpp"
+#include "spirv/unified1/spirv.hpp"
```
dneto0: Yay. Thanks.

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -180,2214 +78,1497 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
...
 
 char ReplaceOpenCLBuiltinPass::ID = 0;
 INITIALIZE_PASS(ReplaceOpenCLBuiltinPass, "ReplaceOpenCLBuiltin",
                 "Replace OpenCL Builtins Pass", false, false)
 
 namespace clspv {
-ModulePass *createReplaceOpenCLBuiltinPass() {
-  return new ReplaceOpenCLBuiltinPass();
-}
+  ModulePass *createReplaceOpenCLBuiltinPass() {
```
dneto0: Let clang-format do its work.

dneto0:
And e.g. step 3 add subgroup builtins.

sjw36:
Thanks @dneto0 , makes sense. I will split it up.

lib/Builtins.h
dneto0: actually, I also suggest changing "sign" to is_signed.

lib/Builtins.cpp
sjw36: Random online.. I will switch to FNV-1a.

lib/Builtins.cpp
sjw36: Yes, that's the ABI. Mostly copied from ReplaceOpenCLBuiltinPass.cpp.

lib/Builtins.cpp
sjw36: Well it is permissible to fail to demangle if the name is not actually a mangled name and some names are an aggregate of strings, usually with a '.' delimiter (see above case) but is it always the case? IDK

sjw36:
superceded by PR #518

Issue #342, opened 2019-04-09
Support get_global_offset() built-in function


kpet:
A few things are needed to support this:
1. A way of passing the offsets from the runtime/application to the shaders (described using the descriptor map).
2. Code in the shaders to compute global IDs from the SPIR-V `BuiltInGlobalInvocationId`.
3. `get_global_offset` needs to return the values passed in.

There are a few possible options pass the offsets:
1. Specialisation constants could make sense for applications that don't change the values much.
2. Via a uniform buffer.
3. Using push constants.
4. Another idea?

This should probably be an optional feature.

kpet:
Commenting to avoid duplication: I've started work on this.

Issue #341, opened 2019-04-09
Support get_work_dim() built-in function


kpet:
The best way is probably via a specialisation constant. In most applications the value doesn't tend to change across kernel enqueues.

kpet:
Commenting to avoid duplication: I've started work on this.

PR #514, opened 2020-03-01
Add support for get_work_dim() and global offset

kpet:
Use push constants to pass the values from the runtime/application.
The required push constants are declared in the descriptor map.

- Introduce generic support for push constants
- Move work-item built-in function tests to a separate folder
- Introduce validation code for explicit structure layouts
- Add -scalar-block-layout option to be able to change the behaviour
  of layout checks based on the presence of the feature
- Fix a few typos in related code.

Contributes to #341, #342.


lib/DefineOpenCLWorkItemBuiltinsPass.cpp
```diff
@@ -212,16 +263,36 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineGlobalSizeBuiltin(Module &M) {
 bool DefineOpenCLWorkItemBuiltinsPass::defineGlobalOffsetBuiltin(Module &M) {
   Function *F = M.getFunction("_Z17get_global_offsetj");
 
-  // If the builtin was not used in the module, don't create it!
-  if (nullptr == F) {
+  // If global offset support is disabled and the builtin not used in the
+  // module, don't create it. Otherwise, always define it as it's used in global
+  // ID calculations.
+  if (!clspv::Option::GlobalOffset() && (nullptr == F)) {
```
alan-baker: get_global_id only uses get_global_offset if `clspv::Option::GlobalOffset()` returns true. It seems like this condition shouldn't define the function unless the option is set.

lib/Option.cpp
```diff
@@ -168,6 +168,18 @@ llvm::cl::opt<clspv::Option::SourceLanguage> cl_std(
...
+static llvm::cl::opt<bool>
+    scalar_block_layout("scalar-block-layout", llvm::cl::init(false),
+                        llvm::cl::desc("Assume VK_EXT_scalar_block_layout"));
+
+static llvm::cl::opt<bool> work_dim(
+    "work-dim", llvm::cl::init(false),
+    llvm::cl::desc("Enable support for get_work_dim() built-in function"));
+
+static llvm::cl::opt<bool>
+    global_offset("global-offset", llvm::cl::init(false),
```
alan-baker: I assume this is intended to default on at some point in the future (e.g. when support is landed in clvk)?

test/WorkItemBuiltins/multiple-push-constant.cl
```diff
@@ -0,0 +1,34 @@
+// RUN: clspv -work-dim -global-offset -descriptormap=%t.dmap %s -o %t.spv
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: FileCheck --check-prefix=DMAP %s < %t.dmap
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
+// DMAP: pushconstant,name,dimensions,offset,0,size,4
```
alan-baker: You should flip these element to generate:
```
// DMAP: pushconstant,name,global_offset,offset,0,size16
// DMAP: pushconstant,name,dimensions,offset,12,size,4
```

lib/PushConstant.h
```diff
@@ -0,0 +1,31 @@
...
+#define CLSPV_LIB_PUSH_CONSTANT_H_
+
+#include "clspv/PushConstant.h"
+
+#include "llvm/IR/BasicBlock.h"
+#include "llvm/IR/Value.h"
+
+namespace clspv {
+
+const char *GetPushConstantName(PushConstant);
```
alan-baker: Function comments please.

test/WorkItemBuiltins/get_global_id-with-global-offset.cl
```diff
@@ -0,0 +1,31 @@
+// RUN: clspv -global-offset %s -o %t.spv
```
alan-baker: Can you please add a version of this test for:
```
out[0] = get_global_id(3);
```

test/WorkItemBuiltins/get_global_offset-push-constant.cl
```diff
@@ -0,0 +1,27 @@
+// RUN: clspv -global-offset -descriptormap=%t.dmap %s -o %t.spv
```
alan-baker: Same here, a version for an out of range access.

test/WorkItemBuiltins/multiple-push-constant.cl
```diff
@@ -0,0 +1,34 @@
+// RUN: clspv -work-dim -global-offset -descriptormap=%t.dmap %s -o %t.spv
```
alan-baker: Another variant of this test please, call get_global_offset with a larger index (or non-constant index) to generate the select.

lib/SPIRVProducerPass.cpp
```diff
@@ -2850,6 +2859,230 @@ void SPIRVProducerPass::GenerateResourceVars(Module &) {
...
+
+  // TODO A row-major matrix of C columns has a base alignment equal to the base
+  // alignment of a vector of C matrix components.
+  // TODO A column-major matrix has a base alignment equal to the base alignment
+  // of the matrix column type.
+
+  llvm_unreachable("Unsupported type");
+}
+
+unsigned roundUpToMultipleOf(unsigned num, unsigned multiple) {
```
alan-baker: This is effectively MathExtras.h `alignTo` (with Skew = 0). I'd prefer reusing that to adding another definition.

lib/DeclarePushConstantsPass.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+bool DeclarePushConstantsPass::runOnModule(Module &M) {
+
+  bool changed = false;
+
+  std::vector<clspv::PushConstant> PushConstants;
+
+  auto &C = M.getContext();
+
+  if (clspv::Option::WorkDim()) {
```
alan-baker: Swap these as mentioned in the multiple push constants test.

lib/SPIRVProducerPass.cpp
kpet: Thanks! I was sure there was already a function for this somewhere.

test/WorkItemBuiltins/multiple-push-constant.cl
kpet: Doh! I should have spotted that, thanks.

lib/Option.cpp
kpet: Maybe but that would mean all the applications that use clspv have to support and pass push constants.

alan-baker: Only if those builtins are used. I don't think that's a problem since push constants are a core feature.

lib/PushConstant.h
kpet: Done.

lib/DeclarePushConstantsPass.cpp
kpet: Done.

lib/DefineOpenCLWorkItemBuiltinsPass.cpp
kpet: This is actually what the code does but reversing the logic might make this clearer.

test/WorkItemBuiltins/get_global_id-with-global-offset.cl
kpet: Done.

test/WorkItemBuiltins/get_global_offset-push-constant.cl
kpet: That is covered by the `get_global_id(3)` test above IMHO since `get_global_id` calls `get_global_offset`. The only thing that makes it into the SPIR-V is a store of a 0 constant.

lib/DefineOpenCLWorkItemBuiltinsPass.cpp
kpet: I've reworked the code a bit to avoid re-declaring `get_global_offset` when it's already used and reworded the comments a bit. Hopefully this makes things clearer.

test/WorkItemBuiltins/multiple-push-constant.cl
kpet: I've ended up doing another variant of the simple `get_global_offset` test with a non-constant index. I didn't feel having get_work_dim in the same test added anything.

lib/Option.cpp
kpet: Currently the descriptor map entries are always present but we could only emit them when the built-in is used for `get_work_dim`. For `get_global_offset` however, the entry would always be present if `get_global_id` is used. I'm not worried about optionality, just forcing applications to support something they may not need. Having said that, using push constants for POD arguments might be a good justification for requiring applications to pass push constants.

kpet:
Thanks for the review!

>We'll need an argument to pass the maxPushConstantsSize limit to the compiler.

That would be good but we can also assume the minimum size required by the Vulkan spec.

I think I've addressed the review comments and fixed the Windows build error. I can't make sense of the clvk bot failure.

PTAl.

alan-baker:
Don't worry about the clvk bot. A recent swiftshader commit seems to have broken it for that use.

I think for the argument I would just default to the minimum maxPushConstantsSize and let users (including clvk) override it as they wish. It's for the future anyways.

PR #518, opened 2020-03-03
Builtins::Lookup interface - phase1

sjw36:
* Copied demangler from ReplaceOpenCLBuiltinPass.cpp
* Should also address all comments from the previous PR
* Links to #510


sjw36:
Bad merge.. I will start over..

PR #519, opened 2020-03-05
Builtins::Lookup interface - phase1

sjw36:
* Copied demangler from ReplaceOpenCLBuiltinPass.cpp
* Should also address all comments from the previous PR
* Links to #510


sjw36:
Apologies for the run-around.. I merged the last one instead of rebasing...

lib/BuiltinsEnum.h
```diff
@@ -0,0 +1,338 @@
...
+#ifndef CLSPV_LIB_BUILTINSENUM_H_
+#define CLSPV_LIB_BUILTINSENUM_H_
+
+namespace clspv {
+namespace Builtins {
+
+enum BuiltinType {
+  kBuiltinNone = 0,
+
+  kType_Async_Start = 1,
```
alan-baker: Why assign values here? It seems just to create a maintenance headache if any builtins are ever added. If you want them sortable based on ranges, I suggest:
```
kBuiltinNone = 0,
kType_Async_Start,
kAsyncWorkGroupCopy = kType_Async_Start,
kWaitGroupEvents,
kPrefetch,
kType_Async_End = kPrefetch,
...
```

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+#ifdef DEBUG
+      printf("Func: %s\n", mangled_name.c_str());
+      llvm_unreachable("failed to demangle name");
+#endif
+      return 0;
+    }
+    break;
+
+    // element types
+  case 'l':
```
alan-baker: I'd prefer no fallthrough cases. Might be clearer to separate length from type and signedness. Also please use `byte_len` instead `blen`

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+    type_info->type_id = Type::StructTyID;
+    pos--;
+    type_info->name = GetUnmangledName(mangled_name, &pos);
+    break;
+  }
+  case '.':
+    return 0;
+  default:
+#ifdef DEBUG
+    printf("Func: %s\n", mangled_name.c_str());
```
alan-baker: Prefer to use llvm::outs() here.

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
 
 #include "Builtins.h"
 
+#include <stdlib.h>
```
alan-baker: #include <cstdlib> please

lib/Builtins.cpp
sjw36: Yes, I agree it is more complicated but slightly more efficient. For the sake of maintainability I will add another switch statement specifically for byte_len.

lib/Builtins.cpp
sjw36: I don't follow?

lib/BuiltinsEnum.h
sjw36: I agree it can lead to problems but this proposal also can be error prone if adding new enums. I prefer not to refer to other enums. Probably best to rely on incremental enum values and remove all numbers and assignments. Does that sound reasonable?

alan-baker: That's ok with me.

lib/Builtins.cpp
alan-baker: Use \<cstdlib\> instead of <stdlib.h> preferably. I think the markdown formatting lost cstdlib in my last comment, sorry.

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+    blen = 2; // double float
+  case 'f':
+    blen <<= 2; // single float
+    type_info->type_id = Type::FloatTyID;
+    type_info->is_signed = true;
+    type_info->byte_len = blen;
+    break;
+  case 'v': // void
+    break;
+  case '1':
```
dneto0: Please explain this.  And why not  '0'

For example, it's not very intuititve what's going on, even when staring at the mangling reference
https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling


I *think* what's going on is that this is a typedef'd name?  
E.g.   struct Foobar { ... }
would have  6Foobar in it, and these cases occur when you're looking at that initial digit. 

??  That's interesting, and needs explanation.

Is it always going to be a struct?  E.g. what if it's a typedef'd name for different kind of thing .like pointer-to-int-in-private-space. ?

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// FunctionInfo::GetFromMangledNameCheck
+//   - parse mangled name as far as possible. Some names are an aggregate of
+//   fields separated by '.'
+//   - extract name and parameter types, and return type for 'convert' functions
+//   - return true if the mangled name can be fully parsed
+bool Builtins::FunctionInfo::GetFromMangledNameCheck(
+    const std::string &mangled_name) {
+  size_t pos = 0;
+  if (mangled_name[pos++] != '_' || mangled_name[pos++] != 'Z') {
```
dneto0: Buffer overrun:  Protect against an input like "X", where you end up indexing too far.

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+    name_ = mangled_name;
+    return false;
+  }
+
+  name_ = GetUnmangledName(mangled_name, &pos);
```
dneto0: Check the error case, empty string.

sjw36:
I believe all comments are addressed in this update

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+    return_type_.type_id = tok == 'f' ? Type::FloatTyID : Type::IntegerTyID;
+  }
+
+  ParamTypeInfo prev_type_info;
+
+  auto mangled_name_len = mangled_name.length();
+  while (pos < mangled_name_len) {
+    ParamTypeInfo type_info;
+    if (mangled_name[pos] == 'S') {
+      // handle duplicate param_type
```
dneto0: Ah, a back-reference.

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+      if (mangled_name[pos + 1] != '_') {
+        return false;
+      }
+      params_.push_back(prev_type_info);
```
dneto0: What if this is an invalid back-reference?  E.g. this S is for the first param.

In fact, could avoid the need for prev_type_info by indexing the last current value of params.
And then you also end up catching the malicious case, where params_ is empty.

lib/Builtins.cpp
```diff
@@ -14,171 +14,335 @@
...
+using namespace clspv;
+
+namespace {
+////////////////////////////////////////////////////////////////////////////////
+////  Convert Builtin function name to a Type enum
+////////////////////////////////////////////////////////////////////////////////
+Builtins::BuiltinType LookupBuiltinType(const std::string &name) {
+
+// Build static map of builtin function names
+#include "BuiltinsMap.inc"
```
dneto0: Good idea to separate this out into its own file.  It's a big table with its own edit/review criteria.  :-)

lib/Builtins.cpp
```diff
@@ -145,21 +133,46 @@ size_t GetParameterType(const std::string &mangled_name,
...
   default:
 #ifdef DEBUG
-    printf("Func: %s\n", mangled_name.c_str());
+    llvm::outs() << "Func: " << mangled_name << "\n";
     llvm_unreachable("failed to demangle name");
 #endif
     return 0;
   }
+
+  // capture byte_len
```
dneto0: This is much nicer!  thanks

lib/Builtins.cpp
sjw36: Should short circuit if mangled_name[0] != '_' but I can reverse the logic to make it more clear that the first 2 characters must equal "_Z".

lib/Builtins.cpp
sjw36: For unqualified names: https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.unqualified-name
It gets more complicated for scoped names, but we shouldn't see that for builtins. 

This doesn't check for '0' since it is only looking for the beginning of a \<positive length number>.

sjw36: typedefs should be resolved in mangled names.

sjw36: I added abit more clarification in the comments.

lib/BuiltinsEnum.h
alan-baker: I think this is still outstanding to change.

sjw36:
Anything else needed guys? I have phase-2 almost ready if you think this is going in the right direction.

lib/BuiltinsEnum.h
sjw36: hmm, thought I updated that last week.. sending shortly

sjw36: ok, updated for real this time.

sjw36:
Thanks for approving @alan-baker! @dneto0 anything else remaining?

PR #521, opened 2020-03-10
Add support for get_enqueued_local_size()

kpet:
Start implementing support for non-uniform NDRanges, following the approach outlined in #522 

### DIFF
--- a/lib/DeclarePushConstantsPass.cpp
+++ b/lib/DeclarePushConstantsPass.cpp
@@ -62,6 +62,13 @@ bool DeclarePushConstantsPass::runOnModule(Module &M) {
     PushConstants.emplace_back(clspv::PushConstant::GlobalOffset);
   }
 
+  if ((clspv::Option::Language() ==
+       clspv::Option::SourceLanguage::OpenCL_C_20) ||
+      (clspv::Option::Language() ==
+       clspv::Option::SourceLanguage::OpenCL_CPP)) {
+    PushConstants.push_back(clspv::PushConstant::EnqueuedLocalSize);
+  }
+
   if (clspv::Option::WorkDim()) {
     PushConstants.push_back(clspv::PushConstant::Dimensions);
   }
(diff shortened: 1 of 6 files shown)


kpet:
I should have explained this in my commit message. Vulkan doesn't support non-uniform workgroup sizes at all at the moment so we will need some kind of emulation until it does. As discussed in our last F2F sync, I'm planning to add support to clvk to split non-uniform NDRanges into multiple dispatches as a first step. With that approach clspv would only have to know about/support uniform regions.

alan-baker:
Can you file an issue with a summary of how you intend to non-uniform workgroups?

kpet:
I've described the overall approach in https://github.com/google/clspv/issues/522.

lib/DeclarePushConstantsPass.cpp
```diff
@@ -62,6 +62,13 @@ bool DeclarePushConstantsPass::runOnModule(Module &M) {
     PushConstants.emplace_back(clspv::PushConstant::GlobalOffset);
   }
 
+  if ((clspv::Option::Language() ==
+       clspv::Option::SourceLanguage::OpenCL_C_20) ||
+      (clspv::Option::Language() ==
+       clspv::Option::SourceLanguage::OpenCL_CPP)) {
```
alan-baker: This should also check if get_enqueued_local_size is used. Seems unnecessary to unconditionally require this value.

kpet: Yes, agree. This is a problem for `get_work_dim()` as well and I thought I'd tackle that as a separate PR but I don't mind doing it here if that's the preference.

alan-baker: I'm ok with a separate PR.

kpet: Ok, I'll make that my next PR then.

kpet:
I've updated the commit message, PTAL.

PR #531, opened 2020-03-18
Builtins::Lookup interface - phase2

sjw36:
* Updated usage in SPIRVProducer, replaced StringSwitch
* Converted SplatArgPass, programmatically generate new function call based on FunctionInfo
* Converted all ReplaceOpenCLBuiltinPass
  * Iterate once over Module->getFunctionList(), switch on FunctionInfo type
  * Changed all replace methods to create a Replacer lambda
* Links to #510

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -27,7 +27,7 @@
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
-#include "spirv/unified1/spirv.hpp"
+#include "spirv/1.0/spirv.hpp"
 
 #include "clspv/AddressSpace.h"
 #include "clspv/DescriptorMap.h"
@@ -175,7 +175,7 @@ bool ReplaceOpenCLBuiltinPass::runOnModule(Module &M) {
   std::list<Function *> func_list;
   for (auto &F : M.getFunctionList()) {
     // process only function declarations
-    if (F.empty() && runOnFunction(F)) {
+    if (F.isDeclaration() && runOnFunction(F)) {
       func_list.push_front(&F);
     }
   }
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -27,117 +27,24 @@
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
-#include "spirv/1.0/spirv.hpp"
+#include "spirv/unified1/spirv.hpp"
 
 #include "clspv/AddressSpace.h"
 #include "clspv/DescriptorMap.h"
 #include "clspv/Option.h"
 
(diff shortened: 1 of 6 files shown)


lib/SplatArgPass.cpp
```diff
@@ -41,184 +52,142 @@ namespace clspv {
...
+}
+} // namespace
+
+// Programmatically convert mangled_name to vectorized version
+std::string
+SplatArgPass::getSplatName(const Builtins::FunctionInfo &func_info,
+                           const Builtins::ParamTypeInfo &param_info,
+                           bool three_params) {
+  const char *type_code = "f";
+  int index = log2i(param_info.byte_len);
```
alan-baker: Since byte_len must be positive, I'd suggest dropping the log2i implementation and using MathExtras.h's Log2_32.

lib/SplatArgPass.cpp
```diff
@@ -41,184 +52,142 @@ namespace clspv {
...
+  std::list<Function *> func_list;
+  for (auto &F : M) {
+    // process only function declarations
+    if (F.empty() && !F.use_empty()) {
+      auto &func_info = Builtins::Lookup(&F);
+      bool has_3_params = false;
+      switch (func_info.getType()) {
+      case Builtins::kClamp:
+      case Builtins::kMix:
+        has_3_params = true;
```
alan-baker: Please avoid these fallthroughs.

lib/SPIRVProducerPass.cpp
```diff
@@ -43,7 +43,7 @@
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
-#include "spirv/1.0/spirv.hpp"
+#include "spirv/unified1/spirv.hpp"
```
alan-baker: I'm a little leary about this change without making all of clspv use the unified headers. If it's not required, I'd prefer doing this separately (a la #421, which should really get finished).

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -235,134 +172,257 @@ ModulePass *createReplaceOpenCLBuiltinPass() {
 } // namespace clspv
 
 bool ReplaceOpenCLBuiltinPass::runOnModule(Module &M) {
-  bool Changed = false;
+  std::list<Function *> func_list;
+  for (auto &F : M.getFunctionList()) {
+    // process only function declarations
+    if (F.empty() && runOnFunction(F)) {
```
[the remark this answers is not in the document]
sjw36: The intention was to only runOnFunction if it IS a declaration, so F.isDeclaration() == F.empty() right?

lib/SPIRVProducerPass.cpp
sjw36: Yep, that is required for my next PR adding sub_groups support. I'll stash it till then.

sjw36: Will #421 be pulled in soon?

alan-baker: I will try to get it finished  by early next week.

Issue #529, opened 2020-03-18
Support for per-kernel push constants


kpet:
Basic support for push constants was introduced recently. In the current implementation, all kernels in a module share the same push constants. While simple, this is sub-optimal and will get in the way of implementing https://github.com/google/clspv/issues/528. This issue covers changing the current implementations so that each kernel in a module can use a different set of push constants. Since each entrypoint can only statically use one push constant block (see 14.5.1 in the Vulkan spec), this means that each kernel will need to have its own structure type for push constants.

alan-baker:
Here is my plan for this:
1. Add the ability to generate POD args in push constants by explicit option
  * incompatible with current push constants
  * must be clustered args
  *. will check against a specified size (default 128B)
2.  Make clustering pod args the default
3. Change current module scope push constants to be specialization constants
  * this will allow drivers to optimize the bounds checking away
  * current push constants will be translated as specialization constants with null initializers
  * new enum for identifying them (since they need to be per-component)
4. Change the selection of POD args storage type to be automatic
  * priority will be: push constant (size limited, no arrays for now), ubo (layout ok), ssbo
  * still allow overrides
  * allow each kernel to make a decision independently
  * probably refactor layout checks
5. Add more options to specify the environment and affect POD arg choices (e.g. 16-bit storage capabilities)

On the clvk side:
1. Support pod push constants, but don't use them by default (after clspv's 1st action)
2. Change current push constants to spec constants (after clspv's 3rd action)
3. Stop sending specific choices about POD args (after clspv's 4th action)

Each action is intended to be a separate PR. The key integration points with clvk will be after switching to spec constants and them after introducing auto assignment of pod args. @kpet I will plan to handle the clvk integration as this occurs.

kpet:
This sounds generally excellent but I don't think I fully agree with 3.

Using specialisation constants for `get_work_dim()` is fine as the number of possibilities is small and the performance impact of having to recompile can probably be mitigated well by pipeline caching. As previously discussed, exploring the use of specialisation constants was the reason I left https://github.com/google/clspv/issues/341 open.

When it comes to global offsets though, I'm really not convinced these should be implemented with specialisation constants. Having to recompile each time the global offset changes feels wasteful.

Also, the non-uniform region support I've been working on currently uses push constants for the region offset (+global offset). We'd still need a few recompilations because of the multiple workgroup sizes but caching is more likely to help with WG sizes than region offsets. Further changes to Vulkan would make this issue go away entirely (not any time soon though).

[...]

PR #542, opened 2020-03-30
POD arguments in push constants

alan-baker:
Contributes to #529 

Support passing POD arguments via push constants.

New options:
* -pod-pushconstant: instruct the compiler to pass pod args in push constants
  * requires clustered pod args 
  * cannot be used with module scope push constants
  * cannot be used with ubo pod args
  * does not support arrays currently
* -max-pushconstant-size: specify the max size of push constants
  * verified in the frontend
  * default to Vulkan minimum of 128B

Changes:
* updated docs
*  cluster pod kernel args now records pod arg kinds correctly
  * simplified uses throughout the flow since it is more reliable now
*  allocate descriptors can allocate push constant pods

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -304,7 +304,7 @@ implementation limit for push constants. This is validated at the start of the c
 Note: Module scope push constanst are currently incompatible with plain-old-data
 arguments sent as push constants.
 
-TODO(alan-baker): fix this.
+TODO(alan-baker): See #529 for the overall plan to address this.
 
 The descriptor map entry for kernel arguments will not contain `descriptorSet` or
 `binding` entries. For example, an integer arg `f` to kernel `foo` might look like:
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -92,11 +92,17 @@ following way:
   types, set `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER`.
 - If the argument to the kernel is a plain-old-data type, the matching Vulkan
   descriptor set type is `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` by default.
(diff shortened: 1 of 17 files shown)


docs/OpenCLCOnVulkan.md
```diff
@@ -92,12 +92,18 @@ following way:
   types, set `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER`.
 - If the argument to the kernel is a plain-old-data type, the matching Vulkan
   descriptor set type is `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` by default.
-  If option `-pod-ubo` is used the `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER`.
+  If option `-pod-ubo` is used the `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER`. If the
```
dneto0, line 95: Pre-existing problem: this is not grammatically correct.
Recommend starting the new sentence on a new line.

lib/AllocateDescriptorsPass.cpp
```diff
@@ -512,7 +512,11 @@ bool AllocateDescriptorsPass::AllocateKernelArgDescriptors(Module &M) {
       for (Argument &Arg : f_ptr->args()) {
         set_and_binding_list.emplace_back(kUnallocated, kUnallocated);
         if (discriminants_list[arg_index].index >= 0) {
-          set_and_binding_list.back().first = set;
+          if (clspv::GetArgKindForType(Arg.getType()) !=
+              clspv::ArgKind::PodPushConstant) {
+            // Don't assign a descriptor set to push constants.
+            set_and_binding_list.back().first = set;
+          }
           set_and_binding_list.back().second = binding++;
```
dneto0, line 520: Seems the binding assignment should be guarded too?

alan-baker: The bindings are assigned in an unused descriptor set to make sure that unique discriminants are kept separate.

Issue #541, opened 2020-03-27
SplatArgPass hangs


jrprice:
Clspv hangs in SplatArgPass for any kernel that triggers its functionality (we seem to be lacking test coverage for it?).

e.g.:
```
kernel void foo(global int4 *data, int value) {
  data[0] = min(data[0], value);
}
```

The cause seems to be [this loop](https://github.com/google/clspv/blob/04f3a95d01078d644f6ea9710061e3e0aac854d1/lib/SplatArgPass.cpp#L161-L163) introduced by #531, presumably because it updates the use list of the function inside the loop that iterates over it (instead of building a work list).

sjw36:
I am investigating..

sjw36:
Well I do see 54 tests exercising this code and your testcase does pass for me but I agree we shouldn't be adding to a list that we are iterating over. I will submit a fix.

PR #543, opened 2020-03-30
BugFix(#541): collect function list first, then replace usages with n…

sjw36:
…ew vectorized version

### DIFF
--- a/lib/SplatArgPass.cpp
+++ b/lib/SplatArgPass.cpp
@@ -135,10 +135,11 @@ void SplatArgPass::replaceCall(Function *NewCallee, CallInst *Call) {
 }
 
 bool SplatArgPass::runOnModule(Module &M) {
-  std::list<Function *> func_list;
-  for (auto &F : M) {
+  std::list<std::pair<Function *, const Builtins::FunctionInfo &>> func_list;
+  // Collect candidates
+  for (auto &F : M.getFunctionList()) {
     // process only function declarations
-    if (F.empty() && !F.use_empty()) {
+    if (F.isDeclaration() && !F.use_empty()) {
       auto &func_info = Builtins::Lookup(&F);
       auto func_type = func_info.getType();
       switch (func_type) {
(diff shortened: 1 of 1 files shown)


lib/SplatArgPass.cpp
```diff
@@ -152,32 +153,31 @@ bool SplatArgPass::runOnModule(Module &M) {
...
+    auto &func_info = FI.second;
+    auto func_type = func_info.getType();
+    auto &param_info = func_info.getParameter(0);
+    bool has_3_params =
+        func_type == Builtins::kClamp || func_type == Builtins::kMix;
+    std::string NewFName = getSplatName(func_info, param_info, has_3_params);
+    Function *NewCallee = getReplacementFunction(*F, NewFName);
+    // Replace the users of the function.
+    for (User *U : F->users()) {
+      replaceCall(NewCallee, dyn_cast<CallInst>(U));
```
jrprice: This is still modifying the user list while iterating over it, so my test case still hangs with this code.

sjw36:
> FWIW `ninja check-spirv` passes for me even though the kernel I posted in #541 hangs, so I don't think we do have coverage for this specific issue.

Interested to know your build os/env @jrprice, thanks.

lib/SplatArgPass.cpp
alan-baker: Thanks for catching that.

@sjw36 I'd suggest using a SmallVector to copy the users before modifying them.

sjw36: Too right, I will fix that..

sjw36: I was just going to :
  while (!F->use_empty()) {
    User *U = *F->users().begin();
    replaceCall(...);
  }
Good?

jrprice:
> Interested to know your build os/env @jrprice

Debian-based Linux distro, building with GCC 9.

lib/SplatArgPass.cpp
alan-baker: Looks ok, but I don't have time to try it out.

sjw36:
Any ideas why this keeps failing?

CMake Error at CMakeLists.txt:1518 (add_custom_command):
  No TARGET 'libGLESv2' has been created in this directory.

alan-baker:
Rebase on top of master. That has been fixed.

lib/SplatArgPass.cpp
```diff
@@ -135,10 +135,11 @@ void SplatArgPass::replaceCall(Function *NewCallee, CallInst *Call) {
...
-  for (auto &F : M) {
+  std::list<std::pair<Function *, const Builtins::FunctionInfo &>> func_list;
+  // Collect candidates
+  for (auto &F : M.getFunctionList()) {
     // process only function declarations
-    if (F.empty() && !F.use_empty()) {
+    if (F.isDeclaration() && !F.use_empty()) {
       auto &func_info = Builtins::Lookup(&F);
       auto func_type = func_info.getType();
       switch (func_type) {
```
jrprice: The `default` clause in this switch seems to have been dropped for some reason:
```warning: 289 enumeration values not handled in switch:```

sjw36: Great, I will add the default clause back as well.

PR #548, opened 2020-04-01
Report all kernels in the descriptor map

### LINKED ISSUE #536: Report kernels with no arguments in the descriptor map?
From https://github.com/kpet/clvk/issues/23:

>clvk currently relies on the descriptor map obtained from clspv to ascertain the existence of a kernel in a given program. Kernels with no arguments aren't described at all in the descriptor map and thus can't be used (kernel creation will fail).

How would you feel about adding a new `kernelnoargs,<name>` type of entry to the descriptor map? This would arguably be an abuse of the descriptor map but one that doesn't feel too bad. Alternatively, clvk could parse the SPIR-V module to get the list of entrypoints but that's one more dependency on SPIRV-Tools that I'd rather avoid if possible.

kpet:
This make it easier for clvk to support kernel with no arguments.

Fixes #536.

### DIFF
--- a/include/clspv/DescriptorMap.h
+++ b/include/clspv/DescriptorMap.h
@@ -50,7 +50,7 @@ const unsigned kSamplerFilterMask = 0x30;
 
 struct DescriptorMapEntry {
   // Type of the entry.
-  enum Kind { Sampler, KernelArg, Constant, PushConstant } kind;
+  enum Kind { Sampler, KernelArg, KernelDecl, Constant, PushConstant } kind;
 
   // Common data.
   uint32_t descriptor_set;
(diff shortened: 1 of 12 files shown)


kpet:
I'm not a huge fan of this solution but it does the job. The descriptor map interface will probably be due a redesign once we're done adding the missing features.

jrprice:
FYI you'll need to rebase on top of master to get the fix for the issue breaking the clvk Kokoro bot.

kpet:
Ok, will do, thanks.

kpet:
Now rebased, PTAL.

kpet:
Just had a look at the clvk bot failure: it's expected as clvk fails when there's an entry in the descriptor map it can't parse. Maybe that should be relaxed to make future changes easier.

alan-baker:
@kpet if you could prioritize a fix for the clvk side of things I'd appreciate it.

kpet:
@alan-baker Here you go: https://github.com/kpet/clvk/pull/186

PR #554, opened 2020-04-06
Change default of -cluster-pod-kernel-args

alan-baker:
Contributes to #529

* Clustered pod args is now on by default
* updated tests
* updated docs

Test changes are NFC.


docs/OpenCLCOnVulkan.md
```diff
@@ -290,7 +291,7 @@ argument as `pod_ubo` rather than the default of `pod`.
 
 Normally plain-old-data arguments are passed into the kernel via a storage buffer.
 Use the option `-pod-pushconstant` to pass these parameters in via push constants. The
-option `-cluster-pod-kernel-args` must also be specified. Push constants are intended
+option `-cluster-pod-kernel-args=0` cannot be specified. Push constants are intended
```
dneto0, line 294: Weird. That's how the LLVM command line options work?
Would have expected -no-cluster-pod-kernel-args
But hey, it's documented, so that's fine with me.

Issue #545, opened 2020-04-01
Proposed refactor SPIRVProducerPass


sjw36:
As a precursor to adding sub_groups instruction support I would like to refactor a bit of the SPIRVProducerPass as follows:

- Create SPIRV Module Sections for SPIRVInstructions according to 2.4 of the [spec](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_logicallayout_a_logical_layout_of_a_module).
  - This removes the need for looking up and inserting OpDecorates
- Remove Constant collection and gather/insert on-the-fly
- Add a Builder class to create and insert new SPIRVInstructions, the only place to ++nextID based on spv::HasResultAndType.
- Create a placeholder for the DeferredInsts instead of capturing an insertion point
- Collect a set of capabilities on-the-fly, and generate list after all instructions generated.

Does this plan sound acceptable?

sjw36:
BTW, it doesn't look like SPIRVProducerPass handles function declarations. Is there a reason? Seems like it may be needed for library support.

alan-baker:
Still thinking about the proposal, answering the question:
>BTW, it doesn't look like SPIRVProducerPass handles function declarations. Is there a reason? 

SPIR-V requires that all function declarations be decorated with the Import Linkage Type. That, in turn, requires the Linkage capability, which Vulkan does not allow. Right now clspv isn't setup to do proper linking for libraries.

sjw36:
That's all for now. I will create a new issue for sub_group support

PR #560, opened 2020-04-07
SPIRVProducerPass refactor - phase1

sjw36:
* Create SPIRV Module Sections for SPIRVInstructions according to 2.4 of the spec.
  * Sections: Capabilities, Extensions, Imports, MemoryModel, EntryPoints, ExecutionModes, Debug, Annotations, Types, Constants, GlobalVariables, Functions
  * This removes the need for looking up where to insert OpDecorates

Phase1 enhancement for issue #545

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -90,6 +90,28 @@ const glsl::ExtInst kGlslExtInstBad = static_cast<glsl::ExtInst>(0);
 
 const char *kCompositeConstructFunctionPrefix = "clspv.composite_construct.";
 
+// SPIRV Module Sections (per 2.4 of the SPIRV spec)
+// These are used to collect SPIRVInstructions by type on-the-fly.
+enum SPIRVSection {
+  kCapabilities,
+  kExtensions,
+  kImports,
+  kMemoryModel,
+  kEntryPoints,
+  kExecutionModes,
+
+  kDebug,
+  kAnnotations,
+
+  kTypes,
+  kConstants,
+  kGlobalVariables,
+
+  kFunctions,
+
+  kSectionCount
+};
+
 enum SPIRVOperandType {
   NUMBERID,
   LITERAL_INTEGER,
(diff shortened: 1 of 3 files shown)


sjw36:
@alan-baker does this look like the right direction?

lib/SPIRVProducerPass.cpp
```diff
@@ -90,6 +90,28 @@ const glsl::ExtInst kGlslExtInstBad = static_cast<glsl::ExtInst>(0);
...
+  kExtensions,
+  kImports,
+  kMemoryModel,
+  kEntryPoints,
+  kExecutionModes,
+
+  kDebug,
+  kAnnotations,
+
+  kTypes,
```
alan-baker, line 106: This shouldn't be broken up like this. Types can depend on constants (e.g. array sizes). Forcing this extra separation will lead to invalid IR. Types, constants and global variables can all be intermixed.

lib/SPIRVProducerPass.cpp
```diff
@@ -90,6 +90,28 @@ const glsl::ExtInst kGlslExtInstBad = static_cast<glsl::ExtInst>(0);
 
 const char *kCompositeConstructFunctionPrefix = "clspv.composite_construct.";
 
+// SPIRV Module Sections (per 2.4 of the SPIRV spec)
+// These are used to collect SPIRVInstructions by type on-the-fly.
+enum SPIRVSection {
+  kCapabilities,
+  kExtensions,
+  kImports,
+  kMemoryModel,
```
alan-baker, line 99: Seems a bit odd to define a list for a single instruction. That being said, I'm not sure there's a better alternative.

lib/SPIRVProducerPass.cpp
sjw36: Agreed. In a future CL I set kConstants = kTypes. That way we can still collect by type but will be the same bucket.

sjw36: I can make the change for this CL.

lib/SPIRVProducerPass.cpp
sjw36: Yes, it's an odd one-off.

sjw36:
Thanks @alan-baker , hope that fixes all the windows build problems. I don't have my windows environment anymore so cannot be sure.

PR #562, opened 2020-04-14
Add generic support to allocate spec ids

alan-baker:
* Specializations ids recorded in metadata
  * used to generate new descriptor map entries
* Changed local memory spec ids to use new allocation method
* Added new enum for specialization constants
  * currently only workgroup size is added to the descriptor map

Example:
```
spec_constant,workgroup_size_x,spec_id,0
spec_constant,workgroup_size_y,spec_id,1
spec_constant,workgroup_size_z,spec_id,2
```

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -192,10 +192,18 @@ Consider this example:
 
 It generates the following descriptor map:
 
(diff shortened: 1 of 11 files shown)


alan-baker:
clvk error due to unrecognized descriptor map entry.

PR #565, opened 2020-04-17
SPIRVProducerPass refactor - phase2

sjw36:
* addSPIRVInst for creating/inserting new SPIRVInstruction
  * Creates/manages ResultID based on spv::HasResultAndType in one place (almost, nextID still used for second pass)
* Consolidated on SmallVector storage for SPIRVOperandVec (replaces SPIRVOperandList)
* Consolidated Workgroup variable creation

Enhancement for clspv issue #545

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2863,10 +2863,10 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
 
       // Allocate spec constants for workgroup size.
       clspv::AddWorkgroupSpecConstants(module);
-      
+
       SPIRVOperandVec Ops;
-      uint32_t result_type_id =
-          getSPIRVType(dyn_cast<VectorType>(Ty->getPointerElementType())->getElementType());
+      uint32_t result_type_id = getSPIRVType(
+          dyn_cast<VectorType>(Ty->getPointerElementType())->getElementType());
 
       // X Dimension
       Ops << MkId(result_type_id) << MkNum(1);
@@ -2913,7 +2913,8 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
     // Ops[1] : Default literal value
     InitializerID = nextID++;
 
-    Ops << MkId(getSPIRVType(IntegerType::get(GV.getContext(), 32))) << MkNum(3);
+    Ops << MkId(getSPIRVType(IntegerType::get(GV.getContext(), 32)))
+        << MkNum(3);
 
     auto *Inst = new SPIRVInstruction(spv::OpSpecConstant, InitializerID, Ops);
     getSPIRVInstList(kConstants).push_back(Inst);
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2910,13 +2910,11 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
     //
     // Ops[0] : Result Type ID
     // Ops[1] : Default literal value
-    InitializerID = nextID++;
 
     Ops << MkId(getSPIRVType(IntegerType::get(GV.getContext(), 32)))
         << MkNum(3);
 
-    auto *Inst = new SPIRVInstruction(spv::OpSpecConstant, InitializerID, Ops);
-    getSPIRVInstList(kConstants).push_back(Inst);
+    InitializerID = addSPIRVInst<kConstants>(spv::OpSpecConstant, Ops);
 
     //
     // Generate SpecId decoration.
@@ -2929,13 +2927,12 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
     Ops << MkId(InitializerID) << MkNum(spv::DecorationSpecId)
         << MkNum(spec_id);
 
-    Inst = new SPIRVInstruction(spv::OpDecorate, Ops);
-    getSPIRVInstList(kAnnotations).push_back(Inst);
+    addSPIRVInst<kAnnotations>(spv::OpDecorate, Ops);
   } else if (BuiltinType == spv::BuiltInGlobalOffset) {
     // 1. Generate a spec constant with a default of {0, 0, 0}.
     // 2. Allocate and annotate SpecIds for the constants.
     // 3. Use the spec constant as the initializer for the variable.
-    SPIRVOperandList Ops;
+    SPIRVOperandVec Ops;
 
     //
     // Generate OpSpecConstant for each dimension.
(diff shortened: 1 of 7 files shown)


sjw36:
This is a bit more extensive than phase1 but addresses my biggest concern with nextID access everywhere. ResultIDs should be captured directly from the newly created SPIRVInstruction.

Apologies for dropping this on Friday, no rush.

lib/SPIRVProducerPass.cpp
```diff
@@ -866,15 +827,6 @@ void SPIRVProducerPass::GenerateLLVMIRInfo(Module &M, const DataLayout &DL) {
...
 
-          // Handle image type specially.
-          if (IsImageBuiltin(callee_name)) {
-            TypeMapType &OpImageTypeMap = getImageTypeMap();
-            Type *ImageTy =
-                Call->getArgOperand(0)->getType()->getPointerElementType();
-            OpImageTypeMap[ImageTy] = 0;
-            getImageTypeList().insert(ImageTy);
-          }
-
```
sjw36, line 877: Moved this to getSPIRVType under the image type.

lib/SPIRVProducerPass.cpp
```diff
@@ -1568,18 +1521,40 @@ void SPIRVProducerPass::FindWorkgroupVars(Module &M) {
...
+
+    SPIRVID VariableID = addSPIRVInst<kGlobalVariables>(spv::OpVariable, Ops);
+
+    Ops.clear();
+    Ops << MkId(ArraySizeID) << MkNum(spv::DecorationSpecId) << MkNum(spec_id);
+    addSPIRVInst<kAnnotations>(spv::OpDecorate, Ops);
+
+    LocalArgInfo info{VariableID,  ElemTy,         ArraySizeID,
+                      ArrayTypeID, PtrArrayTypeID, spec_id};
+    LocalSpecIdInfoMap[spec_id] = info;
```
sjw36, line 1557: With SPIRV Sections the different components of Workgroup variables can be created all at once.

lib/SPIRVProducerPass.cpp
```diff
@@ -1925,716 +1900,598 @@ spv::BuiltIn SPIRVProducerPass::GetBuiltin(StringRef Name) const {
...
+}
 
-    Ops.clear();
-    Ops << MkNum(spv::StorageClassWorkgroup) << MkId(arg_info.array_type_id);
-    SPIRVInstList.push_back(new SPIRVInstruction(
-        spv::OpTypePointer, arg_info.ptr_array_type_id, Ops));
+void SPIRVProducerPass::GenerateSPIRVTypes(LLVMContext &Context,
+                                           Module &module) {
+  for (Type *Ty : getTypeList()) {
+    getSPIRVType(Ty);
```
sjw36, line 2305: At some point types can be created on demand but many tests would need updated. So for now still gather a TypeList and generate types up front.

lib/SPIRVProducerPass.cpp
```diff
@@ -1925,716 +1900,598 @@ spv::BuiltIn SPIRVProducerPass::GetBuiltin(StringRef Name) const {
...
+  return 0;
+}
 
+void SPIRVProducerPass::GenerateSPIRVConstants() {
+  ValueList &CstList = getConstantList();
+
+  for (uint32_t i = 0; i < CstList.size(); i++) {
+    // UniqueVector ids are 1-based.
+    Constant *Cst = cast<Constant>(CstList[i + 1]);
+    getSPIRVValue(Cst);
```
sjw36, line 2490: Same concept applies as GenerateSPIRVTypes above.

lib/SPIRVProducerPass.cpp
```diff
@@ -112,6 +114,8 @@ enum SPIRVSection {
   kSectionCount
 };
 
+typedef uint32_t SPIRVID;
```
alan-baker, line 117: Prefer not to introduce this typedef. Just use uint32_t.

lib/SPIRVProducerPass.cpp
```diff
@@ -496,8 +427,40 @@ struct SPIRVProducerPass final : public ModulePass {
   // Returns true if |Arg| is called with a coherent resource.
   bool CalledWithCoherentResource(Argument &Arg);
 
+  //
+  // Primary interface for adding SPIRVInstructions to a SPIRVSection.
+  template <enum SPIRVSection TSection = kFunctions>
```
alan-baker, line 432: Why templating over just taking an extra parameter? Seems like a waste for the compiler to generate extra functions to start with. I'd also rather be explicit everywhere than just making kFunction instructions easier to insert.

lib/SPIRVProducerPass.cpp
```diff
@@ -496,8 +427,40 @@ struct SPIRVProducerPass final : public ModulePass {
...
+  SPIRVID addSPIRVInst(spv::Op Op, const char *V) {
+    SPIRVOperandVec Ops;
+    Ops << MkString(V);
+    return addSPIRVInst<TSection>(Op, Ops);
+  }
+
 private:
   static char ID;
+
+  Module *module;
```
alan-baker, line 462: I dislike the mixing this introduces. It's ok to store the module pointer, but I'd rather stop passing around the module when that happens. This would ideally be changed in a separate PR.

lib/SPIRVProducerPass.cpp
```diff
@@ -1568,18 +1521,40 @@ void SPIRVProducerPass::FindWorkgroupVars(Module &M) {
...
-    FindConstant(ConstantInt::get(IdxTy, 0));
-    FindType(IdxTy);
-    FindType(arg->getType());
+    Ops.clear();
+    Ops << MkNum(spv::StorageClassWorkgroup) << MkId(ArrayTypeID);
+    SPIRVID PtrArrayTypeID = addSPIRVInst<kTypes>(spv::OpTypePointer, Ops);
+
+    // Generate OpVariable.
+    //
+    // GIDOps[0] : Result Type ID
```
alan-baker, line 1544: Nit: What are GIDOps?

lib/SPIRVProducerPass.cpp
```diff
@@ -1925,716 +1900,598 @@ spv::BuiltIn SPIRVProducerPass::GetBuiltin(StringRef Name) const {
...
+  }
+  case Type::PointerTyID: {
+    PointerType *PTy = cast<PointerType>(Ty);
+    unsigned AddrSpace = PTy->getAddressSpace();
+
+    if (AddrSpace != AddressSpace::UniformConstant) {
+      auto PointeeTy = PTy->getElementType();
+      if (PointeeTy->isStructTy() &&
+          dyn_cast<StructType>(PointeeTy)->isOpaque()) {
+        // TODO(sjw): assert always an image?
```
alan-baker, line 1944: Samplers should also be in UniformConstant.

lib/SPIRVProducerPass.cpp
sjw36: Since every call uses an explicit Section the compiler can optimize for it and should be pretty insignificant for compile time since there are only 11 specializations.

lib/SPIRVProducerPass.cpp
sjw36: It was a convenience in a few places but I will forgo this for now and make another PR.

lib/SPIRVProducerPass.cpp
sjw36: Copied from another location, it was wrong there also. It should just be Ops.

lib/SPIRVProducerPass.cpp
sjw36: It does make more explicit what is being returned vs a random number.

lib/SPIRVProducerPass.cpp
sjw36: Actually needed in getSPIRVType unless I capture the DataLayout.

alan-baker: For clarity, I was proposing a first PR to make that change, then this one.

sjw36: Working on it..

sjw36:
Updated and rebased.

sjw36:
I can't understand why windows is so triggered by this file. If I turn on -Werror-unused-parameter I get errors in many files in lib..

sjw36:
Ok, this is too much. We are setting up a windows build env.

sjw36:
Another case where turning on -Werror-shadow causes compile failures in lib directory..

sjw36:
Apologies @alan-baker , didn't realize that was also part of the windows error.

alan-baker:
Sorry, you'll need to rebase again @sjw36.

sjw36:
No problem @alan-baker , should be ready now.

PR #566, opened 2020-04-20
Update LLVM

jrprice:
* Use element count/type getters from VectorType subclass


lib/ReplaceLLVMIntrinsicsPass.cpp
```diff
@@ -149,7 +149,7 @@ bool ReplaceLLVMIntrinsicsPass::replaceMemcpy(Module &M) {
       } else if (OutType->isArrayTy()) {
         OutType = OutType->getArrayElementType();
       } else if (OutType->isVectorTy()) {
```
alan-baker, line 151: Would be better to do
```
} else if (auto vec_ty = dyn_cast<VectorType>(OutType)) {
  OutType = vec_ty->getElementType();
} else {
```

alan-baker: Basically anytime there's a check on isVectorTy().

PR #567, opened 2020-04-20
SPIRVProducerPass refactor - phase1.1

sjw36:
* cache current Module

Related to issue google#545


sjw36:
@alan-baker here it is..

sjw36:
Can we turn off compile error on unused variable for windows or turn it on everywhere?

alan-baker:
I'll look into the option changes.

PR #568, opened 2020-04-21
Generate work dimension as a spec constant

alan-baker:
Contributes to #529 

* Change work dim to be a spec constant instead of a push constant
  * generates an initialized private variable
* Update tests
* Update docs


jrprice:
Now that we're using a spec constant for `get_work_dim()`, do we really still need the `-work-dim` flag? We could just always generate the spec constant if `get_work_dim()` is used (which I imagine will be a very small number real kernels), and then the client can decide whether they want to override the default of `3` or not.

alan-baker:
> Now that we're using a spec constant for `get_work_dim()`, do we really still need the `-work-dim` flag? We could just always generate the spec constant if `get_work_dim()` is used (which I imagine will be a very small number real kernels), and then the client can decide whether they want to override the default of `3` or not.

I could change the default to true on -work-dim as part of this change.

jrprice:
Presumably `Dimensions` needs to be removed from the `PushConstant` enum:
https://github.com/google/clspv/blob/cf53df47266a60930bdad2fa7570605cb716d949/include/clspv/PushConstant.h#L21
(there would be some other code in `PushConstant.cpp` that needs to be deleted too).

There's also one extra reference to work-dim as push constant in a comment here:
https://github.com/google/clspv/blob/cf53df47266a60930bdad2fa7570605cb716d949/include/clspv/Passes.h#L61

alan-baker:
> Presumably `Dimensions` needs to be removed from the `PushConstant` enum:
> https://github.com/google/clspv/blob/cf53df47266a60930bdad2fa7570605cb716d949/include/clspv/PushConstant.h#L21

I decided against removing from that enum yet. Clspv simply won't generate it. It could be removed later.

lib/SPIRVProducerPass.cpp
```diff
@@ -1923,6 +1923,7 @@ spv::BuiltIn SPIRVProducerPass::GetBuiltin(StringRef Name) const {
       .Case("__spirv_WorkgroupSize", spv::BuiltInWorkgroupSize)
       .Case("__spirv_NumWorkgroups", spv::BuiltInNumWorkgroups)
       .Case("__spirv_WorkgroupId", spv::BuiltInWorkgroupId)
+      .Case("__spirv_WorkDim", spv::BuiltInWorkDim)
```
kpet, line 1926: This is a bit dodgy/confusing IMHO as we're not using the SPIR-V `WorkDim` built-in. Pragmatically speaking, no harm done though.

PR #569, opened 2020-04-21
Remove unused variables

alan-baker:
* Add compiler error to clang and gcc for unused variables
* Remove unused variables throughout the code
* Fix a bug in get_global_offset that potentially accessed out-of-bounds
memory
  * updated related test
* Fixed some deprecated uses of CreateShuffleVector


sjw36:
👍  thanks @alan-baker

PR #572, opened 2020-04-26
Initial support for non-uniform NDRanges

### LINKED ISSUE #522: Support for OpenCL 2.x non-uniform NDRanges
Here's the approach I'm proposing we use to emulate support for non-uniform workgroups as defined by OpenCL 2.x until support appears in Vulkan:
- Non-uniform NDRanges are split into 1 to 8 uniform regions depending on the number of dimensions for which the NDRange is not uniform.
- The application/runtime uses one compute dispatch per uniform region. These can be run in parallel.
- `get_enqueued_local_size` returns the workgroup size for the entire NDRange as passed by the application/runtime via a push constant (as done in #521)
- A new push constant for a region offset + global offset is passed by the application/runtime. Global IDs are calculated by adding the global invocation ID to that new (region offset + global offset) push constant.

kpet:
This implementation uses the approach outlined in #522 and push constants
for all the values that need to be communicated by the application/runtime.
The push constant space required is quite large in the worst case (which is
unlikely to be a common occurrence in typical code). There are definitely
applications where using specialisation constants everywhere would result
in a ridiculous amount of recompilation and likely others where saving some
push constant space would be a win (especially once POD arguments will
use push constants).

The CTS tests require that reqd_work_group_size be treated not as a hard
requirement but more as an upper bound/hint which seems odd. I've created
https://github.com/KhronosGroup/OpenCL-Docs/issues/267 to discuss this.
This change satisfies the CTS and is good enough to pass the entire
non_uniform_work_group CTS suite with the appropriate clvk changes.

Fixes #522.

### DIFF
--- a/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
+++ b/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
@@ -45,13 +45,11 @@ struct DefineOpenCLWorkItemBuiltinsPass final : public ModulePass {
                            AddressSpace::Type AddrSpace = AddressSpace::Input);
 
   bool defineGlobalIDBuiltin(Module &M);
-
+  bool defineNumGroupsBuiltin(Module &M);
+  bool defineGroupIDBuiltin(Module &M);
   bool defineGlobalSizeBuiltin(Module &M);
-
   bool defineGlobalOffsetBuiltin(Module &M);
-
   bool defineWorkDimBuiltin(Module &M);
-
   bool defineEnqueuedLocalSizeBuiltin(Module &M);
 
   bool addWorkgroupSizeIfRequired(Module &M);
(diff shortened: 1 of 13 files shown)


lib/DefineOpenCLWorkItemBuiltinsPass.cpp
```diff
@@ -194,13 +190,21 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineGlobalIDBuiltin(Module &M) {
...
 
-  // If we have a global offset we need to add it
   Value *Ret = GidBase;
-  if (clspv::Option::GlobalOffset()) {
-    auto Goff =
-        Builder.CreateCall(M.getFunction("_Z17get_global_offsetj"), Dim);
-    Goff->setCallingConv(CallingConv::SPIR_FUNC);
-    Ret = Builder.CreateAdd(GidBase, Goff);
+  if (clspv::Option::NonUniformNDRangeSupported()) {
+    auto Ptr = GetPushConstantPointer(BB, clspv::PushConstant::RegionOffset);
```
alan-baker, line 195: The docs were a bit vague, is region offset the sum of global_offset and the clvk specific split of global ids? If not, should the offset still be added?

kpet: Yes, it is. Maybe I could rename it to make this clearer (`global_plus_region_offset`?). Agree the docs need to say this explicitly.

test/WorkItemBuiltins/get_global_id-non-uniform.cl
```diff
@@ -0,0 +1,34 @@
...
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: FileCheck --check-prefix=DMAP %s < %t.dmap
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
+// DMAP: pushconstant,name,region_offset,offset,0,size,12
+
+// CHECK:     OpMemberDecorate %[[_struct_3:[0-9a-zA-Z_]+]] 0 Offset 0
+// CHECK:     OpDecorate %[[_struct_3]] Block
+// CHECK:     OpMemberDecorate %[[_struct_12:[0-9a-zA-Z_]+]] 0 Offset 0
+// CHECK:     OpDecorate %[[_struct_12]] Block
```
alan-baker, line 12: I would just drop these. The push constant tests should provide coverage that the right decorations are generated. Same edit for most of the tests to reduce overall LOC.

kpet: Yes, good idea. It'll reduce the amount of changes needed in the future as well. Now done.

PR #575, opened 2020-04-27
Automate pod arg implementation decision

alan-baker:
Contributes to #529 

Add a new pass to decide on a per kernel basis how pod args should be implemented. Currently handles three possibilities: ssbo, ubo and per-kernel push constant. Soon I want to support a type-mangled global push constant interface. The pass attempts to use push constants if possible, then UBOs and finally falls back on SSBOs.

Changes:
 * new pass
   * automated choice can be overridden by command line options
 * updated cluster pod args to copy metadata
 * refactored ArgKind utilities to use the actual argument everywhere
   * new utility looks at the kernel's pod arg implementation to decide
   * old utility based on type is used for pointer args
 * refactored some push constant related helper methods into PushConstant.cpp/h
 * refactored layout checks into separate files
   * added a new version to check a whole struct at once
 * updated many tests to account for different pod arg choices
    * picked a spread of implementations

### DIFF
--- a/lib/PushConstant.cpp
+++ b/lib/PushConstant.cpp
@@ -101,4 +101,25 @@ Value *GetPushConstantPointer(BasicBlock *BB, PushConstant pc) {
   return Builder.CreateInBoundsGEP(GV, Indices);
 }
 
+bool UsesGlobalPushConstants(Module &M) {
+  return clspv::Option::NonUniformNDRangeSupported() ||
+         ShouldDeclareEnqueuedLocalSize(M) || ShouldDeclareGlobalOffset(M);
+}
+
+bool ShouldDeclareEnqueuedLocalSize(Module &M) {
+  bool isEnabled = ((clspv::Option::Language() ==
+                     clspv::Option::SourceLanguage::OpenCL_C_20) ||
+                    (clspv::Option::Language() ==
+                     clspv::Option::SourceLanguage::OpenCL_CPP));
+  bool isUsed = M.getFunction("_Z23get_enqueued_local_sizej") != nullptr;
+  return isEnabled && isUsed;
+}
+
+bool ShouldDeclareGlobalOffset(Module &M) {
+  bool isEnabled = clspv::Option::GlobalOffset();
+  bool isUsed = (M.getFunction("_Z17get_global_offsetj") != nullptr) ||
+                (M.getFunction("_Z13get_global_idj") != nullptr);
+  return isEnabled && isUsed;
+}
+
 } // namespace clspv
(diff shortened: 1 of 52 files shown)


lib/PushConstant.cpp
```diff
@@ -85,4 +87,24 @@ Value *GetPushConstantPointer(BasicBlock *BB, PushConstant pc) {
   return Builder.CreateInBoundsGEP(GV, Indices);
 }
 
+bool UsesGlobalPushConstants(Module &M) {
+  return ShouldDeclareEnqueuedLocalSize(M) || ShouldDeclareGlobalOffset(M);
```
jrprice, line 91: Presumably this needs to rebased and updated to include the new global push constants added by #572.

test/AutoPodArgs/force_pushconstant.ll
```diff
@@ -0,0 +1,14 @@
+; RUN: clspv-opt -AutoPodArgs -pod-pushconstant %s -o %t.ll
+; RUN: FileCheck %s < %t.ll
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK: define spir_kernel void @foo(i32 addrspace(1)* %out, i32 %pod) !clspv.pod_args_impl [[MD:![0-9]+]]
+; CHECK: [[MD]] = !{i32 2}
+define spir_kernel void @foo(i32 addrspace(1)* %out, i32 %pod) {
```
jrprice, line 9: The `auto-pod-args` pass will make this a POD arg a push constant anyway, so this test doesn't really test the `-pod-pushconstant` override. Maybe needs a `-max-pushconstant-size=1` like the `max_size_prevents_push constants` test so that the pass would pick UBO instead? Though I don't know whether forcing push constants that would exceed the size should actually be an error case...

lib/PushConstant.cpp
alan-baker: Do that right now.

test/AutoPodArgs/force_pushconstant.ll
alan-baker: The test you describe would be rejected by the frontend so I'd prefer not to add it. The option requests push constants and the compiler delivers. Yes maybe it would anyway, but I'm not actually too worried about the case.

lib/PushConstant.cpp
alan-baker: Done

PR #576, opened 2020-05-01
Add ability to generate global offset as spec constant

alan-baker:
Contributes to #529

* -global-offset will now generate either a spec constant or push
constant
  * push constant is used for OpenCL 2.0 and OpenCL C++
  * push constant can be forced with -global-offset-push-constant
* global offset as a spec constant generates a private variable with
spec constant initializer
  * new SpecConstant enum values for x, y and z dimensions of global
  offset
* refactored some logic for push constants
* updated docs
* new tests

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -624,7 +624,7 @@ The OpenCL C work-item functions map to Vulkan SPIR-V as follows:
   `get_local_size()` by the result from `get_num_groups()`.
 - `get_global_id()` is mapped to a SPIR-V variable decorated with
   `GlobalInvocationId`. The global offset is added to that variable when
-   `-global-offset` is enabled.
+   `-global-offset` or `-global-offset-push-constant` is enabled.
 - `get_local_size()` is mapped to a SPIR-V variable decorated with
   `WorkgroupSize`.
 - `get_local_id()` is mapped to a SPIR-V variable decorated with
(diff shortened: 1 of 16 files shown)


alan-baker:
clvk will need an update to support the spec constants.

lib/DefineOpenCLWorkItemBuiltinsPass.cpp
```diff
@@ -382,11 +384,21 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineGlobalOffsetBuiltin(Module &M) {
     auto Dim = &*F->arg_begin();
     auto InBoundsDim = inBoundsDimensionIndex(Builder, Dim);
     Value *Indices[] = {Builder.getInt32(0), InBoundsDim};
-    auto GoffPtr =
-        GetPushConstantPointer(BB, clspv::PushConstant::GlobalOffset);
-    auto GoffDimPtr = Builder.CreateInBoundsGEP(GoffPtr, Indices);
-    auto GoffDim = Builder.CreateLoad(GoffDimPtr);
-    auto Ret = inBoundsDimensionOrDefaultValue(Builder, Dim, GoffDim, 0);
+    Value *gep = nullptr;
+    const bool uses_push_constant = clspv::ShouldDeclareGlobalOffset(M);
```
jrprice, line 388: A little pedantic, but the function name `clspv::ShouldDeclareGlobalOffset()` is a little confusing now, since it specifically means "should declare global offset _as a push constant_". When I first read this line of code I thought it was a bug until I read the definition of that function.

docs/OpenCLCOnVulkan.md
```diff
@@ -630,8 +633,10 @@ The OpenCL C work-item functions map to Vulkan SPIR-V as follows:
   `NumWorkgroups`.
 - `get_group_id()` is mapped to a SPIR-V variable decorated with
   `WorkgroupId`.
-- `get_global_offset()` is mapped to the `global_offset` push constant when
-  `-global-offset` is enabled, otherwise it always returns 0.
+- `get_global_offset()` is mapped to the `global_offset` spec constant or push
+  constant when `-global-offset` is enabled, otherwise it always returns 0.
```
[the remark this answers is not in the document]
alan-baker: I'll clarify about the options, but since this is talking about how specific functions are mapped it doesn't need to be mentioned that it only occurs if used.

lib/DefineOpenCLWorkItemBuiltinsPass.cpp
alan-baker: I'll clarify the name.

PR #580, opened 2020-05-06
Add fine-grained storage restrictions

alan-baker:
Contributes to #529

* Remove -f16bit_storage in favour of new options
  * updated uses
* New options to restrict which storage classes support 8- and 16-bit
types
  * -no-storage-16bit and -no-storage-8bit
  * Default behaviour assumes support for all storage classes
* Added frontend diagnostics to check 8- and 16-bit storage support of
kernel arguments
* Modified pod arg determination to check storage class support
* Added a restriction to not generate push constant pod args if pod args
contain an array
* Added tests

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -20,9 +20,6 @@
 namespace clspv {
 namespace Option {
 
-// Returns true if code generation can use SPV_KHR_16bit_storage.
-bool F16BitStorage();
-
 // Returns true if each kernel must use its own descriptor set for all
 // arguments.
 bool DistinctKernelDescriptorSets();
@@ -167,6 +164,18 @@ static bool NonUniformNDRangeSupported() {
   return (Language() == SourceLanguage::OpenCL_CPP) ||
          ((Language() == SourceLanguage::OpenCL_C_20));
 }
+
+enum class StorageClass : int {
+  kSSBO = 0,
+  kUBO,
+  kPushConstant
+};
+
+// Returns true if |sc| supports 16-bit storage.
+bool Supports16BitStorageClass(StorageClass sc);
+
+// Returns true if |sc| supports 8-bit storage.
+bool Supports8BitStorageClass(StorageClass sc);
 
 } // namespace Option
 } // namespace clspv
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -168,7 +168,7 @@ static bool NonUniformNDRangeSupported() {
 enum class StorageClass : int {
   kSSBO = 0,
   kUBO,
-  kPushConstant
+  kPushConstant,
 };
 
 // Returns true if |sc| supports 16-bit storage.
--- a/lib/AutoPodArgsPass.cpp
+++ b/lib/AutoPodArgsPass.cpp
@@ -49,6 +49,14 @@ private:
 
   // Makes kernel |F| use |impl| as the pod arg implementation.
   void AddMetadata(Function &F, clspv::PodArgImpl impl);
+
+  // Returns true if |type| contains an array. Does not consider pointers since
+  // we are dealing with pod args.
+  bool ContainsArrayType(Type *type) const;
+
+  // Returns true if |type| contains a 16-bit integer or floating-point type.
+  // Does not consider pointer since we are dealing with pod args.
+  bool ContainsSizedType(Type *type, uint32_t width) const;
 };
 } // namespace
 
@@ -92,6 +100,13 @@ void AutoPodArgsPass::runOnFunction(Function &F) {
 
     pod_types.push_back(arg_type);
 
+    // If the type contains an 8- or 16-bit type UBO storage must be supported.
+    satisfies_ubo &= !ContainsSizedType(arg_type, 16) ||
+                     clspv::Option::Supports16BitStorageClass(
+                         clspv::Option::StorageClass::kUBO);
+    satisfies_ubo &= !ContainsSizedType(arg_type, 8) ||
+                     clspv::Option::Supports8BitStorageClass(
+                         clspv::Option::StorageClass::kUBO);
     if (auto struct_ty = dyn_cast<StructType>(arg_type)) {
       // Only check individual arguments as clustering will fix the layout with
       // padding if necessary.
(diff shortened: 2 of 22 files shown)


include/clspv/Option.h
```diff
@@ -168,6 +165,18 @@ static bool NonUniformNDRangeSupported() {
          ((Language() == SourceLanguage::OpenCL_C_20));
 }
 
+enum class StorageClass : int {
+  kSSBO = 0,
```
dneto0, line 169: These are not quite identical to storage class as a SPIR-V concept.  :-/

But close enough for your use.
Really it's "kinds of storage used for passing directly accessible laid-out data".  But that doesn't have a good name.

no request for change

lib/AutoPodArgsPass.cpp
```diff
@@ -49,6 +49,14 @@ class AutoPodArgsPass : public ModulePass {
 
   // Makes kernel |F| use |impl| as the pod arg implementation.
   void AddMetadata(Function &F, clspv::PodArgImpl impl);
+
+  // Returns true if |type| contains an array. Does not consider pointers since
```
dneto0, line 53: Nit:  Consider "does not look through pointers" instead of "does not consider pointers"  ?

lib/AutoPodArgsPass.cpp
```diff
@@ -49,6 +49,14 @@ class AutoPodArgsPass : public ModulePass {
...
+  // we are dealing with pod args.
+  bool ContainsArrayType(Type *type) const;
+
+  // Returns true if |type| contains a 16-bit integer or floating-point type.
+  // Does not consider pointer since we are dealing with pod args.
```
dneto0, line 58: nit: similar here.

lib/AutoPodArgsPass.cpp
```diff
@@ -104,12 +119,24 @@ void AutoPodArgsPass::runOnFunction(Function &F) {
...
+  const bool support_16bit_pc = !ContainsSizedType(pod_struct_ty, 16) ||
+                                clspv::Option::Supports16BitStorageClass(
+                                    clspv::Option::StorageClass::kPushConstant);
+  const bool support_8bit_pc = !ContainsSizedType(pod_struct_ty, 8) ||
+                               clspv::Option::Supports8BitStorageClass(
+                                   clspv::Option::StorageClass::kPushConstant);
   const bool satisfies_push_constant =
-      !(!clspv::Option::ClusterPodKernelArgs() ||
-        clspv::UsesGlobalPushConstants(M) ||
+      clspv::Option::ClusterPodKernelArgs() && support_16bit_pc &&
```
dneto0, line 134: This is pretty hard to unpack. 
Recommend breaking down that last clause.  Can be a future CL.

lib/AutoPodArgsPass.cpp
```diff
@@ -141,3 +168,35 @@ void AutoPodArgsPass::AddMetadata(Function &F, clspv::PodArgImpl impl) {
...
+  }
+
+  return false;
+}
+
+bool AutoPodArgsPass::ContainsSizedType(Type *type, uint32_t width) const {
+  if (auto int_ty = dyn_cast<IntegerType>(type)) {
+    return int_ty->getBitWidth() == width;
+  } else if (type->isHalfTy()) {
+    return true;
```
dneto0, line 189: This should be return width == 16.

alan-baker:
```
isHalfTy () const
--
  | Return true if this is 'half', a 16-bit IEEE fp type
```

That's from LLVM's docs. LLVM doesn't have a derived type for floating point types.

alan-baker: oh nvm i see what you meant.

alan-baker:
I thought about it, but remain uncertain. I can take a stab at something.

dneto0:
Only truly needed if the -f16bit_storage option was already described.

alan-baker:
I added some docs in the latest update.

dneto0:
Doc update is good.!

kpet:
@alan-baker Are you working on the corresponding clvk changes? I've started work on that to fix some CTS regressions introduced by your recent changes to POD arguments.

lib/AutoPodArgsPass.cpp
```diff
@@ -49,6 +49,14 @@ class AutoPodArgsPass : public ModulePass {
 
   // Makes kernel |F| use |impl| as the pod arg implementation.
   void AddMetadata(Function &F, clspv::PodArgImpl impl);
+
+  // Returns true if |type| contains an array. Does not look through pointers
+  // since we are dealing with pod args.
+  bool ContainsArrayType(Type *type) const;
+
+  // Returns true if |type| contains a 16-bit integer or floating-point type.
```
jrprice, line 57: Replace "16-bit" with "{width}-bit"?

alan-baker:
Yes, I was planning on working on the clvk changes for this. Needs to query the right extensions and feature bits.

alan-baker:
If you have WIP to do this sort of thing, I'm also ok if you want to continue it.

kpet:
@jrprice has got the patch for these already apparently (discussed by email). I suggest you guys go first with the missing features/extensions patch as I haven't started yet. I believe I've got the missing piece of the puzzle in the form of a fix to the push constant update code (Vulkan VU states it's only valid to do so in aligned chunks of 4 bytes).

lib/Option.cpp
```diff
@@ -203,6 +193,28 @@ static llvm::cl::opt<bool> cluster_non_pointer_kernel_args(
                    "other arguments. Use this to reduce storage buffer "
                    "descriptors."));
 
+static llvm::cl::list<clspv::Option::StorageClass> no_16bit_storage(
+    "no-16bit-storage",
+    llvm::cl::desc("Specify fine-grained 16-bit storage capabilities. SSBO "
+                   "support is assumed."),
```
jrprice, line 199: "SSBO support is assumed" -> don't we assume all storage classes are supported by default? (possible hangover from when this was going to be an enabling option rather than a disabling one?)

Maybe "Specify" should also be changed to "Disable".

lib/Option.cpp
```diff
@@ -203,6 +193,28 @@ static llvm::cl::opt<bool> cluster_non_pointer_kernel_args(
...
+    llvm::cl::values(
+        clEnumValN(clspv::Option::StorageClass::kSSBO, "ssbo", "ssbo"),
+        clEnumValN(clspv::Option::StorageClass::kUBO, "ubo", "ubo"),
+        clEnumValN(clspv::Option::StorageClass::kPushConstant, "pushconstant",
+                   "push constant")));
+
+static llvm::cl::list<clspv::Option::StorageClass> no_8bit_storage(
+    "no-8bit-storage",
+    llvm::cl::desc("Specify fine-grained 8-bit storage capabilities. SSBO "
+                   "support is assumed."),
```
jrprice, line 210: Same comment as above.

kpet:
So with this PR, https://github.com/kpet/clvk/pull/228 and my patch, the regression I observed on basic/parameters_type is gone. buffers/sub_buffers_read_write was passing for me but didn't pass validation. It now does.

alan-baker:
@kpet, great news, thanks!

PR #585, opened 2020-05-11
Expand functionality to catch odd truncates

### LINKED ISSUE #583: Invalid integer widths from mask + switch
Compiling this kernel generates SPIR-V that uses 2-bit integers:
```
kernel void foo(global float *output)
{
  size_t i = get_global_id(0);

  float result = 0.f;
  switch (i & 0x3) {
    case 1:
      result = 2.f;
      break;
    case 2:
      result = 4.f;
      break;
  }

  output[i] = result;
}
```

Commands:
```
$ bin/clspv foo.cl
$ spirv-val a.spv 
error: line 26: Invalid number of bits (2) used for OpTypeInt.
  %u2 = OpTypeInt 2 0
```

alan-baker:
Fixes #583

* Changes UndoTruncatedSwitchPass to look for truncates to look for
truncates to non-power-of-2 integer sizes instead of switches on odd
sizes
  * rename pass
  * new logic to update instructions that aren't full rewritten
  * new tests

The first commit has the heart of the changes. The second commit just renames things.

### DIFF
--- a/lib/UndoTruncateToOddIntegerPass.cpp
+++ b/lib/UndoTruncateToOddIntegerPass.cpp
@@ -44,7 +44,6 @@ private:
   // values are created.
   // TODO(dneto): Handle 64 bit case as well, but separately.
   Value *ZeroExtend(Value *v, uint32_t desired_bit_width) {
-    //outs() << "Zero extend to " << desired_bit_width << " bits: " << *v << "\n";
     auto bit_width = 0;
     if (v->getType()->isIntegerTy())
       bit_width = v->getType()->getIntegerBitWidth();
@@ -249,8 +248,6 @@ bool UndoTruncateToOddIntegerPass::runOnModule(Module &M) {
     if (!zombie->hasNUsesOrMore(1))
       zombie->eraseFromParent();
   }
-
-  //outs() << M << "\n";
 
   return Changed;
 }
--- a/lib/UndoTruncateToOddIntegerPass.cpp
+++ b/lib/UndoTruncateToOddIntegerPass.cpp
@@ -222,33 +222,12 @@ bool UndoTruncateToOddIntegerPass::runOnModule(Module &M) {
     }
   }
 
-  // Remove the zombies if we can.  We expect to. Generally this can be done in
-  // 1 or 2 passes.
-  SmallVector<Instruction *, 8> extra;
-  for (auto *zombie : zombies_) {
-    if (!zombie->hasNUsesOrMore(1)) {
+  // Remove the zombies if we can.  We expect to. We've ordered zombies in
+  // reverse.
+  for (int i = zombies_.size(); i >= 1; --i) {
+    auto zombie = zombies_[i];
+    if (!zombie->hasNUsesOrMore(1))
       zombie->eraseFromParent();
-    } else {
-      extra.push_back(zombie);
-    }
-  }
-
-  // This is not ideal, but the way we identify instructions doesn't lend
-  // itself to a single pass well.
-  bool updated = true;
-  while (updated) {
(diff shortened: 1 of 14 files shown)


lib/UndoTruncateToOddIntegerPass.cpp
```diff
@@ -160,32 +206,50 @@ bool UndoTruncatedSwitchConditionPass::runOnModule(Module &M) {
...
+          WorkList.push_back(user_inst);
+        }
+      }
     }
   }
 
-  // Remove the zombies if we can.  We expect to.  They are ordered from
-  // combinations down to their supporting values.
+  // Remove the zombies if we can.  We expect to. Generally this can be done in
+  // 1 or 2 passes.
```
jrprice, line 226: I couldn't come with a scenario where we *wouldn't* be able to remove all zombie instructions (i.e. no instruction that is replaced should ever still have users).

Since we're doing the replacement recursively, can't we just build a stack of instructions to remove and then do them all bottom up (in one pass)? I just hacked this into your changes and they worked fine for all existing tests. If not, we're missing test coverage for the case where this doesn't work.

alan-baker: Hmm I think I must have tested it wrong in development because UniqueVector indexing starts at 1.

lib/UndoTruncateToOddIntegerPass.cpp
```diff
@@ -160,31 +206,28 @@ bool UndoTruncatedSwitchConditionPass::runOnModule(Module &M) {
...
 
-  // Remove the zombies if we can.  We expect to.  They are ordered from
-  // combinations down to their supporting values.
-  for (auto *zombie : zombies_) {
-    if (!zombie->hasNUsesOrMore(1)) {
+  // Remove the zombies if we can.  We expect to. We've ordered zombies in
+  // reverse.
+  for (int i = zombies_.size(); i >= 1; --i) {
+    auto zombie = zombies_[i];
+    if (!zombie->hasNUsesOrMore(1))
```
jrprice, line 229: Arguably this should just be an assertion; if there's still uses then something has gone wrong.

alan-baker:
I didn't change zero extend, but yes it does look incorrect.

alan-baker:
I've made changes to fix zext handling and other integer sizes.

lib/UndoTruncateToOddIntegerPass.cpp
```diff
@@ -56,44 +58,45 @@ struct UndoTruncateToOddIntegerPass : public ModulePass {
...
+              (uint32_t)APInt::getAllOnesValue(bit_width).getZExtValue()),
           "", trunc);
     } else if (auto *zext = dyn_cast<ZExtInst>(v)) {
-      result = new ZExtInst(zext->getOperand(0), i32_, "", zext);
+      auto tmp = ZeroExtend(zext->getOperand(0), desired_bit_width);
+      auto operand_bit_width = tmp->getType()->getIntegerBitWidth();
+      if (isPowerOf2_32(operand_bit_width)) {
+        result = tmp;
+      } else {
+        result = new ZExtInst(zext->getOperand(0), desired_int_ty, "", zext);
```
jrprice, line 86: I don't quite understand what situation this code is handling. When is `result = tmp` not the right thing to do? Doesn't `ZeroExtend()` always produce a result with the desired bit width? I couldn't come up with a sequence of instructions that triggers the `else` clause.

alan-baker: I've reworked this and added a few tests with mixed sizes that need extended (not all 32-bits).

PR #586, opened 2020-05-11
SPIRVProducerPass refactor - phase2.2

sjw36:
* Removed pre-pass collection of Constants (caused some reorder in tests)
* Collect Capabilitys on-the-fly

* This is the last refactor phase (for now). This enables easy implementation for extensions (e.g. SubGroups)

Enhancement for clspv issue #545


sjw36:
Testing windows now.. strange that linux passed the same tests.

lib/SPIRVProducerPass.cpp
```diff
@@ -165,9 +165,8 @@ struct SPIRVOperand {
 
 typedef SmallVector<SPIRVOperand, 4> SPIRVOperandVec;
 
-SPIRVOperandVec &operator<<(SPIRVOperandVec &list, SPIRVOperand elem) {
+void operator<<(SPIRVOperandVec &list, SPIRVOperand elem) {
```
alan-baker, line 168: I don't follow why this wasn't a problem previously. This change is resulting in a lot of churn in the diff and an unintuitive usage requirement.

sjw36: I agree. The problem is that now constants are added on-the-fly so the getSPIRVValue call has side effect of possibly creating a constant. As well the getSPIRVType can also create a new type, however these are mostly created up front (I would also like to remove the FindType pass but it causes alot of changes in tests).

And so we get to the << operator which can process the inputs in any order (apparently reverse order on Windows). I was considering a new interface where << could take a Type * or a Value *. This way it could preserve the order since the operator overloads would be called in order. I can make that change here and keep the chaining of << but with passing these types instead.

alan-baker: So the tests were too fragile and expecting instructions in too specific an order?

sjw36: Yes, that is correct (did not use CHECK-DAG). The generated code was correct in both cases but not consistent across platforms.

sjw36:
@alan-baker please review latest. This is where I wanted to go, cleans up alot and makes operands more obvious.

sjw36:
please hold off, 1 more debug fix.

sjw36:
Ok, all clear.

lib/SPIRVProducerPass.cpp
```diff
@@ -3340,104 +2917,16 @@ void SPIRVProducerPass::GenerateFuncPrologue(Function &F) {
...
-  if (hasVariablePointers()) {
-    //
-    // Generate OpCapability.
-    //
-    // Ops[0] = Capability
-    //
-    addSPIRVInst<kCapabilities>(spv::OpCapability,
-                                spv::CapabilityVariablePointers);
-  } else if (hasVariablePointersStorageBuffer()) {
+  for (auto Capability : CapabilitySet) {
```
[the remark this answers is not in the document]
sjw36: Hmm, I didn't think I changed the logic on capabilites, just moved where the tests happen. This does create an ordered tree, so they should be the same every time.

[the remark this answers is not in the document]
alan-baker: We used to test:
```
if (hasVariablePointers()) {
} else if (hasVariablePointersStorageBuffer()) {
}
```

So there would only be a single variable pointers capability, but now you iterate through the set and potentially add both capabilities. Similar things happened for Sampled1D and Image1D.

PR #588, opened 2020-05-15
Support global type-mangled pod args in push constants

alan-baker:
Contributes to #529 

* AutoPodArgsPass can now decide to put pod kernel arguments into a type mangled structure shared by all kernels with that style of pod arg implementation
* The heavy lifting is in ClusterPodKernelArgumentsPass
  * for this style pod args are removed from the kernel, but still appear in the kernel arg map
  * instead the pod args are implemented as a struct of integers in the same variable that holds the other global push constants (e.g. region offset)
  * only considers unpacked struct sizes right now
  * offsets account for global push constants also being present
  * When generating the wrapper function, the pass converts an appropriate number of integers back into the appropriate type
    * the type conversion does a pretty good job of generating a minimal number of instructions
    * though the type conversion results in more instructions it can also workaround the storage restrictions since only i32s are loaded from the push constant variable
* FIxed a bug in AutoPodArgsPass where byval pod args were not properly considered for size restrictions
* Push constant loads are now hidden and unhidden the same as __constant loads
  * this prevents generating some awkward nested constant expressions
* updated the spirv producer to handle the new style
* bunch of new tests

### DIFF
--- a/include/clspv/PushConstant.h
+++ b/include/clspv/PushConstant.h
@@ -25,6 +25,7 @@ enum class PushConstant : int {
   RegionOffset,
   NumWorkgroups,
   RegionGroupOffset,
+  KernelArgument,
 };
 
 } // namespace clspv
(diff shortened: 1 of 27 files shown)


alan-baker:
@kpet this is the final step for #529 I talked about. We'll have to benchmark the type conversion code vs the other styles of pod args.

kpet:
@alan-baker Thanks for letting me know.

PR #598, opened 2020-06-01
Builtins refactor - phase3

sjw36:
* moved all function names in Constants.h to builtins (mangled names in Constants.cpp)
  * add some missing functions (pack, unpack, composite_construct, copy_memory, atomic_xor)
* replaced all new function creation with name generation from Builtins::GetMangledFunctionName (ReplaceOpenCLBuiltinPass.cpp)
* moved most CallInst codegen for Builtins to GenerateInstruction (from HandleDeferredInstruction)
* removed all image test routines from Builtins (replaced with Builtins::Lookup)
* updated IR tests

Enhancement as part of https://github.com/google/clspv/issues/545


sjw36:
@alan-baker please restart testing, the formatting should be good now. Thanks.

lib/Builtins.cpp
```diff
@@ -273,113 +273,117 @@ Builtins::Lookup(const std::string &mangled_name) {
...
+std::string Builtins::GetMangledTypeName(Type *Ty) {
+  std::string mangled_type_str;
+
+  switch (Ty->getTypeID()) {
+  case Type::VoidTyID:
+    return "v";
+  case Type::HalfTyID:
+  case Type::FloatTyID:
+  case Type::DoubleTyID:
+    switch (Ty->getScalarSizeInBits()) {
```
alan-baker, line 315: This seems like it could be simplified to:
```
case Type::HalfTyID:
  return "Dh";
case Type::FloatTyID:
  return "f";
case Type::DoubleTyID:
  return "d";
```

lib/Constants.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright 2018 The Clspv Authors. All rights reserved.
```
alan-baker, line 1: Wrong year.

lib/SpecializeImageTypes.cpp
```diff
@@ -143,8 +143,15 @@ Type *SpecializeImageTypesPass::RemapType(Argument *arg) {
 Type *SpecializeImageTypesPass::RemapUse(Value *value, unsigned operand_no) {
   if (CallInst *call = dyn_cast<CallInst>(value)) {
     auto called = call->getCalledFunction();
-    if (IsSampledImageRead(called) || IsUnsampledImageRead(called) ||
-        IsImageWrite(called)) {
+    auto func_info = Builtins::Lookup(called);
+    // TODO(sjw): should this handle half types also?
```
alan-baker, line 147: read_imageh is converted in ReplaceOpenCLBuiltins with read_imagef and a conversion as Vulkan requires the sampled type to be a 32-bit scalar. Similar for write_imageh. So those shouldn't appear at this point in the flow.

lib/Builtins.cpp
```diff
@@ -258,7 +257,8 @@ Builtins::FunctionInfo::getParameter(size_t _arg) const {
 
 // Test for OCL Sampler parameter type
 bool Builtins::ParamTypeInfo::isSampler() const {
-  return type_id == Type::StructTyID && name == "ocl_sampler";
+  return type_id == Type::StructTyID &&
+         (name == "ocl_sampler" || name == "opencl.sampler_t");
```
alan-baker, line 261: Is this a result of just using the struct name?

lib/Builtins.cpp
sjw36: Yes, true, since this is OpenCL kernel language I guess we can rely on the type.

lib/Builtins.cpp
sjw36: Now in ReplaceOpenCLBuiltinPass when using Builtins::GetMangledFunctionName to generate replacement functions (such as replacing read_imageh with read_imagef) the actual arguments are used to mangle the name. So subsequent Unmangling will have a different name for sampler struct. This is the actual name of the struct when we see it in the IR.

Issue #602, opened 2020-07-01
cl_khr_subgroups support


sjw36:
We would like to enable the cl_khr_subgroups extension for the following functions:

These functions will map to Builtin constants:
* get_sub_group_size - SubgroupSize
* get_max_sub_group_size - SubgroupMaxSize
* get_num_sub_groups - NumSubgroups
* get_sub_group_id - SubgroupId
* get_sub_group_local_id - SubgroupLocalInvocationId

And these functions will map to spv::Op's accordingly:
* sub_group_all - OpGroupNonUniformAll
* sub_group_any - OpGroupNonUniformAny
* sub_group_broadcast - OpGroupNonUniformBroadcast
* sub_group_barrier - OpControlBarrier

And the reduction functions as follows:
* sub_group_<group_operation>_add
  * INT: OpGroupNonUniformIAdd
  * FLOAT: OpGroupNonUniformFAdd
* sub_group_<group_operation>_min
  * INT: OpGroupNonUniformSMin
  * UINT: OpGroupNonUniformUMin
  * FLOAT: OpGroupNonUniformFMin 
* sub_group_<group_operation>_max
  * INT: OpGroupNonUniformSMax
  * UINT: OpGroupNonUniformUMax
  * FLOAT: OpGroupNonUniformFMax 
With <group_operation> correspondingly:
* reduce - GroupOperationReduce
* scan_exclusive - GroupOperationExclusiveScan
* scan_inclusive - GroupOperationInclusiveScan

Adding the cl_khr_subgroups extension support requires OpenCL 2.0 or greater (clang front-end already supports with -cl-std=CL2.0) and also requires SPIR-V 1.3 or greater support. A command line switch will be required to specify the -spirv-std for the output version.

Setting -spirv-std=1.4 or greater requires updating the EntryPointInterface with all GlobalVariable IDs.

All tests should be run with -spirv-std=1.5 to test conformance.

sjw36:
The remainder of cl_khr_subgroups functions that should be implemented for full support:
* get_enqueued_num_subroups
* sub_group_reserve_read_pipe
* sub_group_reserve_write_pipe
* sub_group_commit_read_pipe
* sub_group_commit_write_pipe
* get_kernel_sub_group_count_for_ndrange
* get_kernel_max_sub_group_size_for_ndrange

alan-baker:
#610 added `sub_group_barrier` support.

dokipen3d:
[...]

PR #603, opened 2020-07-01
cl_khr_subgroups support

sjw36:
* Added hidden switch '-spirv-std', default 1.0.
* Added SPIRVProducerPass::addSPIRVGlobalVariable to also capture entry_point_interface when -spirv-std=1.4 or greater
* Added builtin function mapping:
  * get_sub_group_size - SubgroupSize
  * get_max_sub_group_size - SubgroupMaxSize
  * get_num_sub_groups - NumSubgroups
  * get_sub_group_id - SubgroupId
  * get_sub_group_local_id - SubgroupLocalInvocationId

  * sub_group_all - OpGroupNonUniformAll
  * sub_group_any - OpGroupNonUniformAny
  * sub_group_broadcast - OpGroupNonUniformBroadcast
  * sub_group_barrier - OpControlBarrier

  * sub_group_<group_operation>_add
    * INT: OpGroupNonUniformIAdd
    * FLOAT: OpGroupNonUniformFAdd
  * sub_group_<group_operation>_min
    * INT: OpGroupNonUniformSMin
    * UINT: OpGroupNonUniformUMin
    * FLOAT: OpGroupNonUniformFMin
  * sub_group_<group_operation>_max
    * INT: OpGroupNonUniformSMax
    * UINT: OpGroupNonUniformUMax
    * FLOAT: OpGroupNonUniformFMax
  * With <group_operation> correspondingly:
    * reduce - GroupOperationReduce
    * scan_exclusive - GroupOperationExclusiveScan
    * scan_inclusive - GroupOperationInclusiveScan

Added tests in test/SubGroup:

* sub_group_barrier.cl
* sub_group_broadcast.cl
* sub_group_id.cl
* sub_group_reduce.cl
* sub_group_scan_exclusive.cl
* sub_group_scan_inclusive.cl

Tested all tests with -spirv-std=1.5:

Failing Tests (2):
  clspv :: LLVMIntrinsics/lifetime_start.cl
  clspv :: mfmt_c.cl

 - both fail for different SPIRV version

This PR addresses issue #602.


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -619,7 +623,8 @@ bool ReplaceOpenCLBuiltinPass::replaceBarrier(Function &F) {
                                                 ConstantScopeWorkgroup, "", CI);
 
     // Lastly, the Execution Scope is always Workgroup Scope.
```
dneto0, line 625: Update the comment, please.  :-)

lib/SPIRVProducerPass.cpp
```diff
@@ -3220,144 +3262,802 @@ spv::Op SPIRVProducerPass::GetSPIRVBinaryOpcode(Instruction &I) {
...
+  case Builtins::kSubGroupReduceMax:
+  case Builtins::kSubGroupScanExclusiveMax:
+  case Builtins::kSubGroupScanInclusiveMax: {
+    capability = spv::CapabilityGroupNonUniformArithmetic;
+    auto &param = FuncInfo.getParameter(0);
+    if (param.type_id == Type::IntegerTyID) {
+      op = param.is_signed ? spv::OpGroupNonUniformSMax
+                           : spv::OpGroupNonUniformUMax;
+    } else {
+      // TODO(sjw): what about Half, Double
```
alan-baker, line 3741: This is the correct opcode, but Vulkan requires extra feature bits from subgroup extended types extension to enable support.

lib/SPIRVProducerPass.cpp
```diff
@@ -3220,144 +3262,802 @@ spv::Op SPIRVProducerPass::GetSPIRVBinaryOpcode(Instruction &I) {
...
+    capability = spv::CapabilityGroupNonUniformVote;
+    op = spv::OpGroupNonUniformAll;
+    break;
+  case Builtins::kSubGroupAny:
+    capability = spv::CapabilityGroupNonUniformVote;
+    op = spv::OpGroupNonUniformAny;
+    break;
+  case Builtins::kSubGroupBroadcast:
+    capability = spv::CapabilityGroupNonUniformBallot;
+    op = spv::OpGroupNonUniformBroadcast;
```
alan-baker, line 3706: Prior to 1.5 the broadcast id must be a constant value. This deserves a note in the docs (or an error if we limit ourselves to generating SPIR-V 1.3).

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -231,6 +231,8 @@ bool ReplaceOpenCLBuiltinPass::runOnFunction(Function &F) {
   case Builtins::kBarrier:
   case Builtins::kWorkGroupBarrier:
     return replaceBarrier(F);
+  case Builtins::kSubGroupBarrier:
```
alan-baker, line 234: OpenCL defines a version of this function with an explicit memory scope operand. The memory scopes are already problematic (see #604). I'd suggesting leaving to the other issue to fix.

lib/SPIRVProducerPass.cpp
```diff
@@ -83,6 +83,9 @@ cl::opt<bool>
     ShowProducerIR("show-producer-ir", cl::init(false), cl::ReallyHidden,
                    cl::desc("Dump the IR at the start of SPIRVProducer"));
 
+cl::opt<std::string> SPIRVVersionOpt("spirv-std", cl::init("1.0"), cl::Hidden,
```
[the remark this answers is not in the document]
sjw36: Ok, I had it marked as hidden but will make it inferred since it is a translation tool.

I am sure there are other problems (I noted that the entry point interfaces are common for all kernels) but it is encouraging that spirv-val passed for all tests when defaulting to SPIRV 1.5.

lib/SPIRVProducerPass.cpp
```diff
@@ -1516,6 +1556,30 @@ SPIRVID SPIRVProducerPass::getOpExtInstImportID() {
...
+  SPIRVOperandVec Ops;
+  Ops << TypeID << SC;
+  if (InitID.isValid()) {
+    Ops << InitID;
+  }
+
+  SPIRVID VID = addSPIRVInst<kGlobalVariables>(spv::OpVariable, Ops);
+
+  if (SC == spv::StorageClassInput || SPIRVMinorVersion >= 4) {
+    getEntryPointInterfacesList().push_back(VID);
```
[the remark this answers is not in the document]
sjw36: Yes, I had a feeling there were nuances to this.

Do we need to track usage per entry point?

lib/SPIRVProducerPass.cpp
```diff
@@ -1927,6 +1991,12 @@ void SPIRVProducerPass::GenerateSPIRVTypes() {
   }
 }
 
+SPIRVID SPIRVProducerPass::getSPIRVConstValue(uint32_t CstVal) {
```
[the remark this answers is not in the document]
sjw36: Ok, I was considering overloading the function for float, int32_t but those are rarely used.

lib/SPIRVProducerPass.cpp
```diff
@@ -3220,144 +3262,802 @@ spv::Op SPIRVProducerPass::GetSPIRVBinaryOpcode(Instruction &I) {
   return Map.at(I.getOpcode());
 }
 
-void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
-  ValueMapType &VMap = getValueMap();
-  LLVMContext &Context = I.getParent()->getParent()->getParent()->getContext();
+SPIRVID SPIRVProducerPass::getSPIRVBuiltin(spv::BuiltIn BID) {
```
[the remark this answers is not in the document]
sjw36: Was thinking of generalizing for others in DefineOpenCLWorkItemBuiltinsPass but that would be a bit more complicated for most.

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
sjw36: Actually, thinking more about this. Shouldn't the version be driven by the clvk driver based on the underlying vulkan implementation version? Currently, clvk is hard coded to SPIRV 1.0 but it does query the vk version from the underlying driver.

lib/SPIRVProducerPass.cpp
```diff
@@ -3134,7 +3176,7 @@ void SPIRVProducerPass::GenerateFuncBody(Function &F) {
 }
 
 spv::Op SPIRVProducerPass::GetSPIRVCmpOpcode(CmpInst *I) {
-  const std::map<CmpInst::Predicate, spv::Op> Map = {
+  static const std::map<CmpInst::Predicate, spv::Op> Map = {
```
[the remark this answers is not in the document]
sjw36: constexpr only works for LiteralTypes, std::map doesn't qualify.. I will have to revert to non-static.

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
alan-baker: clvk will catch the problem when it looks module's capabilities. clspv doesn't have useful error handling except in the frontend plugin right now. I'd prefer not to hit llvm_unreachable because you're on a 1.0 implementation and used subgroups. If you can check everything in the frontend, it might be ok.

alan-baker:
When you update this PR, please also rebase on master to pick up the new changes for the windows bots.

lib/SPIRVProducerPass.cpp
```diff
@@ -796,7 +837,8 @@ bool SPIRVProducerPass::runOnModule(Module &M) {
 void SPIRVProducerPass::outputHeader() {
   binaryOut->write(reinterpret_cast<const char *>(&spv::MagicNumber),
                    sizeof(spv::MagicNumber));
-  const uint32_t spv_version = 0x10000; // SPIR-V 1.0
+  const uint32_t spv_version =
+      (SPIRVMajorVersion << 16) | (SPIRVMinorVersion << 8);
```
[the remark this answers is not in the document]
sjw36: How about just encapsulating in a class that has a 'get()' returning the munged version?

lib/ReplaceOpenCLBuiltinPass.cpp
sjw36: work_group_barrier also has an explicit memory scope overload. Is that supported?

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
sjw36: For now I will limit version to 1.3.

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
alan-baker: That sounds ok.

lib/ReplaceOpenCLBuiltinPass.cpp
alan-baker: I'll deal with that in #604 too. Certainly the scope operand isn't handled correctly.

lib/SPIRVProducerPass.cpp
```diff
@@ -83,6 +83,9 @@ cl::opt<bool>
     ShowProducerIR("show-producer-ir", cl::init(false), cl::ReallyHidden,
                    cl::desc("Dump the IR at the start of SPIRVProducer"));
 
+cl::opt<std::string> SPIRVVersionOpt("spv-version", cl::init("1.0"), cl::Hidden,
```
alan-baker, line 86: I've been thinking about this some more and it would be better to use an enum here I think.

```
enum SPIRVVersion {
  SPIRV_1_0,
  SPIRV_1_3,
  SPIRV_1_5, // note that lack of command line option to specify it
}

cl::option<SPIRVVersion> version("spv-version", cl::init(SPIRV_1_0),
                                  cl::values(
                                    clEnumValN(SPIRV_1_0, "1.0", "SPIR-V version 1.0 (Vulkan 1.0)")
                                    clEnumValN(SPIRV_1_3, "1.3", "SPIR-V version 1.3 (Vulkan 1.1). Experimental")),
                                  cl::desc("Specify the SPIR-V binary version"));              
```

LLVM will do all the parsing and generate a reasonable error. Then just have a method that converts to the packed uint32_t version for the binary and use the enum everywhere else.

Also, I'm sort of leaning to towards exposing the option in the help and the docs now and not supporting inference. We can defer the frontend error checking for now. Sorry for the change on this.

lib/SPIRVProducerPass.cpp
```diff
@@ -467,6 +526,12 @@ struct SPIRVProducerPass final : public ModulePass {
     return RID;
   }
 
+  //
+  // Add global variable and capture entry point interface
+  // TODO(sjw): make entry point interfaces unique per entry point?
```
alan-baker, line 531: Remove this TODO since we only generate up to 1.3 currently.

lib/SPIRVProducerPass.cpp
```diff
@@ -3220,191 +3283,850 @@ spv::Op SPIRVProducerPass::GetSPIRVBinaryOpcode(Instruction &I) {
...
+  case Builtins::kGetSubGroupSize:
+    return loadBuiltin(spv::BuiltInSubgroupSize);
+  case Builtins::kGetNumSubGroups:
+    return loadBuiltin(spv::BuiltInNumSubgroups);
+  case Builtins::kGetSubGroupId:
+    return loadBuiltin(spv::BuiltInSubgroupId);
+  case Builtins::kGetSubGroupLocalId:
+    return loadBuiltin(spv::BuiltInSubgroupLocalInvocationId);
+  case Builtins::kGetEnqueuedNumSubGroups:
+    return loadBuiltin(spv::BuiltInNumEnqueuedSubgroups, spv::CapabilityKernel);
```
alan-baker, line 3714: Kernel is incompatible with Shader. We can't implement num_enqueued_subgroups this way. Leave it unimplemented. Update the docs too.

sjw36:
Forgot to remove the sub_group_barrier.cl test 👎 .. should be good now. @alan-baker I think this covers all the review comments.

lib/SPIRVProducerPass.cpp
sjw36: Sounds good. Also like it better to use the supported version from the underlying vulkan driver. Will need some changes in clvk.

docs/OpenCLCOnVulkan.md
```diff
@@ -850,7 +859,8 @@ The `group_op` qualifier translates as follows:
 
 These extension built-in functions are not supported:
 
-- `get_max_sub_group_size()`
+- `get_max_sub_group_size()` requires CapabiliryKernel (incompatible with Shader)
```
alan-baker, line 862:
```suggestion
- `get_max_sub_group_size()` requires CapabilityKernel (incompatible with Shader)
```

sjw36:
Thanks @alan-baker. This should do it.

PR #605, opened 2020-07-07
Update LLVM

alan-baker:
* add alignments to AtomicRMWInst


alan-baker:
Going to have to live with windows failures until I upgrade MSVC on the bots.

Issue #506, opened 2020-02-14
Incorrect mapping of fmax/fmin


mantognini:
Hi,

I've noticed that OpenCL C `fmax` function is mapped to GLSL SPIR-V `FMax` instruction. But the behaviour is not the same w.r.t. NaNs. Instead, I believe the `NMax` instruction could be used. Similarly, OpenCL C `fmin` is mapped to `FMin` instead of `NMin`.

For convenience, here are the various definitions:
##### OpenCL C `fmax(x, y)`
> Returns y if x < y, otherwise it returns x. If one argument is a NaN, fmax() returns the other argument. If both arguments are NaNs, fmax() returns a NaN.

##### GLSL SPIR-V `FMax`
> Result is y if x < y, either x or y when both x and y are zeros, otherwise x. Which operand is the result is undefined if one of the operands is a NaN.

##### GLSL SPIR-V `NMax`
> Result is y if x < y, either x or y when both x and y are zeros, otherwise x. If one operand is a NaN, the other operand is the result. If both operands are NaN, the result is a NaN.

I guess it doesn't help that the correct OpenCL SPIR-V instructions are `fmax` and `fmin`...

The following program can be used as input to test this.
```
kernel void kern(global float* x) {
  x[0] = fmin(x[1], x[2]);
  x[3] = fmin(x[1], x[2]);
}

alan-baker:
FClamp -> NClamp still needs to be addressed.

PR #609, opened 2020-07-16
Use NMax/NMin for fmax/fmin (#506)

mantognini:
OpenCL C fmax/fmin maps to GLSL SPIR-V instruction NMax/NMin due to the
behaviour requirement on NaN inputs.

Applies by transitivity to OpenCL C fract.


dneto0:
Indeed, OpenCL 2.0 says "Returns fmin(fmax(x, minval), maxval). Results are undefined if minval > maxval."

mantognini:
That is a good observation. Would you rather extend the scope of #506 or have a new issue for clamp?

alan-baker:
@mantognini I'm ok either way. Let me know your preference. I'll merge this PR if you prefer to open another.

mantognini:
Given that I'm not sure when I'll be able to update this PR, it might be more beneficial to merge this one as-is. As for having a new issue or extending the existing one, I don't have a preference.

PR #610, opened 2020-07-17
Fix barrier and mem_fence scopes

### LINKED ISSUE #604: barrier and mem_fence memory scope should not be based on flags
Currently barrier produces a device memory scope based on whether CLK_GLOBAL_MEM_FENCE is specified in the flags and mem_fence always produces a device memory scope. For the OpenCL 1.2 functions these should always be workgroup. 

OpenCL 2.0 introduces new variants that take an explicit memory scope operand. For those variants the scope should be translated into the SPIR-V memory scope.

alan-baker:
Fixes #604

* barrier, work_group_barrier and mem_fence should default to workgroup
  memory scope
* support CL 2.0 explicit memory scope barriers
* support sub_group_barrier
* use AcquireRelease semantics instead of SequentiallyConsistent

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -121,7 +121,7 @@ struct ReplaceOpenCLBuiltinPass final : public ModulePass {
   bool replaceFmod(Function &F);
   bool replaceExp10(Function &F, const std::string &basename);
   bool replaceLog10(Function &F, const std::string &basename);
-  bool replaceBarrier(Function &F);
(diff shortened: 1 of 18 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -606,20 +611,39 @@ bool ReplaceOpenCLBuiltinPass::replaceBarrier(Function &F) {
     // MemorySemanticsSequentiallyConsistentMask.
```
jrprice, line 611: Comment needs updating too (SeqCst -> AcqRel).

sjw36:
👍

PR #612, opened 2020-07-30
Embedded reflection

### LINKED ISSUE #592: Re-design descriptor map api
The descriptor map api is out-dated compared to more recent clspv functionality. In particular the assumption that entries all get a descriptor and binding assigned is no longer true.

I can imagine the new design being more hierarchical and the overall api having more functionality. For example, grouping all of a kernel's arguments under the kernel declaration.

alan-baker:
Fixes #592 

Changes the descriptor map to be implemented as a non-semantic extended instruction set embedded directly in the SPIR-V.

* remove -descriptormap
* added a new tool `clspv-reflection` to dump the old style descriptor map
  * legacy tests use this instead of -descriptormap where they checked the descriptor results
* remove DescriptorMap.{cpp,h}
  * add Sampler.h as a replacement for the sampler enums in descriptor map
* docs update

Amber and Clvk bots will fail on this PR. They will require changes to continue to work. This PR has been tested against version of both codebases that support the reflection and their tests are passing.


docs/OpenCLCOnVulkan.md
```diff
@@ -203,16 +211,30 @@ The shaders produced use the GLSL450 memory model. As such, there is an assumpti
 no aliasing by default. The compiler does not generate *Aliased* decorations
 currently. Users should be aware of this and ensure they are not relying on aliasing.
 
-#### Descriptor map
+#### Embedded Reflection Instructions
+
+Clspv embeds reflection information via use of a non-semantic extended
+instruction set. The instruction set is defined
+[here](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/nonsemantic/NonSemantic.ClspvReflection.html).
```
dneto0, line 218: I would recommend actually naming the extended instruction set in the visible words, so only reading the text tells me what it is. That supports and promotes the extended instruction set itself.

> Clspv embeds reflection information via use of the `[NonSemantic.ClspvReflection.0](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/nonsemantic/NonSemantic.ClspvReflection.html)` non-semantic extended
instruction set.

docs/OpenCLCOnVulkan.md
```diff
@@ -203,16 +211,30 @@ The shaders produced use the GLSL450 memory model. As such, there is an assumpti
...
+
+#### Descriptor map (DEPRECATED)
 
 The compiler can report the descriptor set and bindings used for samplers
 in the sampler map and for the kernel arguments, and also array sizing information for
 pointer-to-local arguments.
-Use option `-descriptormap` to name a file that should contain the mapping information.
+Run `clspv-reflection <spirv> -o <descriptor map>` to produce the descriptor map.
 
 Example:
```
dneto0, line 235: Delete the example as it's no longer valid. ??

dneto0: Or change it to show use of clspv-reflection instead.

Issue #613, opened 2020-07-31
Support for "long vector" types


mantognini:
Because "long vector" types (i.e. any vector type of arity 8 or 16) cannot be directly represented in Vulkan SPRI-V, their usages are currently rejected by the frontend.

We (arm) have been working on a prototype that lowers these vector types into equivalent aggregate of scalars and massages OpenCL C builtin function calls and other LLVM instructions to produce valid, equivalent Vulkan SPIR-V. The core functionality is implemented as an LLVM module pass. It is, in some ways, similar to the LLVM scalariser but it selectively scalarises instructions and handles further builtin functions as well as function parameters.

The benefit of this pass is to increase support for arm compute library (#349). Before providing our implementation, we wanted to give a high level description of how it works and make sure that the proposed approach looks sensible to you.

Here is an example of what the prototype currently supports:

```
static size_t getId() { return get_global_id(0); }
static float8 load(global float *ptr) { return vload8(getId(), ptr); }
static void process(float8 a, private float8 *b) { *b = fma(a, 4, 7); }
static void store(global float *ptr, float8 value) { vstore8(value, getId(), ptr); }

kernel void k(global float *in, global float *out) {
  float8 x = load(in);
  process(x * 2, &x);
  store(out, x);
}
```

Long vector types are supported as variable types and parameters types for user-defined functions but not for kernel parameters. It might be possible to support the latter case, too, but we left it out for now and an appropriate error message is provided to the user. Pointers to these types are supported as well. The pass will rewrite user-defined functions, when needed. E.g. the `load` function shown above is rewritten as follows in LLVM IR:

```
define internal spir_func { float, float, float, float, float, float, float, float } @load.1(float addrspace(1)* %0) #1 {
entry:
  %call.i = call spir_func i32 @getId() #5
  %1 = shl i32 %call.i, 3
  %2 = getelementptr float, float addrspace(1)* %0, i32 %1
  %3 = load float, float addrspace(1)* %2, align 4
  %4 = or i32 %1, 1
  %5 = getelementptr float, float addrspace(1)* %0, i32 %4
  %6 = load float, float addrspace(1)* %5, align 4
  %7 = or i32 %1, 2
  %8 = getelementptr float, float addrspace(1)* %0, i32 %7
  %9 = load float, float addrspace(1)* %8, align 4
  %10 = or i32 %1, 3
  %11 = getelementptr float, float addrspace(1)* %0, i32 %10
  %12 = load float, float addrspace(1)* %11, align 4
[...]

PR #621, opened 2020-08-25
Various improvements to the compilation process

mantognini:
I've noticed some issues with the build process. Namely, the available options were not the same for a static build than for a shared build. For example, the `--MultiVersionUBOFunctions` option of `clspv-opt` wasn't available in a static build. I had exactly the same issue for the pass I want to introduce for #613 so here is a general fix.

I also had to fix compilation for shared build. I tested this using `cmake --target check-spirv` with the four {Debug,Release} x {shared,static} configurations on Linux.

Below are the commit messages for the two commits in this PR.

---

1. Improve support for shared build

Fix compilation for clspv-reflection when doing a shared-library build.
 - Add clspv_passes to the public list of dependencies of clspv_core so
   that clspv-opt, clspv-reflection and any other target depending on
   clspv_core don't have to link against the dependencies of
   clspv_passes.
 - Remove unnecessary dependencies of clspv-opt now that they are
   transitively added thanks to the previous point.
 - Move clangCodeGen dependency to clspv_core because it isn't used by
   clsvp_passes.
 - Because SPIRV-Tools-shared has hidden symbol visibility, some symbols
   are undefined when linking against SPIRV-Tools (which is an alias for
   SPIRV-Tools-shared or SPIRV-Tools-static based on build options).
   This is worked around by always linking against SPIRV-Tools-static,
   regardless of the build options.
 - Don't explicitly list LLVM components as dependencies as this is
   resolved automatically by CMake.

2. Improve support for static build

Use OBJECT library for clspv_passes to ensure all the symbols for global
static variables are present in clspv-opt, ensuring passes using a
static RegisterPass<> object are registered. This is effectively
equivalent to using the linker option --whole-archive for static build,
but in a more portable way as not all toolchains support this option.

Previously, the output of `clspv-opt --help-hidden` and `clspv --help-hidden`
would differ from a shared library build to a static
library build. There are still some differences but they seem to be
unrelated to clspv passes but rather related to LLVM passes.

The MultiVersionUBOFunctionsPass pass is using such a global static
object. Update existing test to cover this pass.

Bump required CMake version. Chose 3.13.4 as it is required by the
latest LLVM and it supports linking against OBJECT libraries.

Add a missing dependency of clspv_passes that was identified by using
OBJECT library when using a shared build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-cmake_minimum_required(VERSION 3.4.3)
+cmake_minimum_required(VERSION 3.13.4)
 
 if(NOT COMMAND find_host_package)
   macro(find_host_package)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -96,7 +96,7 @@ foreach(clspv_lib clspv_core clspv_passes)
 
 endforeach(clspv_lib)
 
-set(CLSPV_LLVM_COMPONENTS clangCodeGen LLVMAnalysis LLVMScalarOpts LLVMCore)
+set(CLSPV_LLVM_COMPONENTS LLVMAnalysis LLVMScalarOpts LLVMCore)
 
 if(${EXTERNAL_LLVM} EQUAL 1)
   include(${CLSPV_LLVM_BINARY_DIR}/lib/cmake/llvm/LLVMConfig.cmake)
@@ -107,11 +107,13 @@ if(${EXTERNAL_LLVM} EQUAL 1)
 endif()
 
 # clspv_c_strings, clspv_glsl, and clspv_reflection are used by SPIRVProducerPass.cpp.
-add_dependencies(clspv_passes clspv_c_strings clspv_glsl clspv_reflection ${CLSPV_LLVM_COMPONENTS})
+add_dependencies(clspv_passes clspv_c_strings clspv_glsl clspv_reflection)
+target_link_libraries(clspv_passes PRIVATE ${CLSPV_LLVM_COMPONENTS})
 
 # clspv_baked_opencl_header is used by Compiler.cpp.
 add_dependencies(clspv_core clspv_baked_opencl_header)
-target_link_libraries(clspv_core PRIVATE clspv_passes ${CLSPV_LLVM_COMPONENTS})
+target_link_libraries(clspv_core PUBLIC clspv_passes)
+target_link_libraries(clspv_core PRIVATE clangCodeGen)
 
 if (MSVC)
   set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/SPIRVProducerPass.cpp"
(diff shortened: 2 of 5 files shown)


dneto0:
> Because SPIRV-Tools-shared has hidden symbol visibility, some symbols
are undefined when linking against SPIRV-Tools (which is an alias for
SPIRV-Tools-shared or SPIRV-Tools-static based on build options).

Should this be considered a bug in the list of visible items in the SPIRV-Tools-shared library?
What symbols are missing?

mantognini:
I see these undefined references:
```
tools/reflection/CMakeFiles/clspv-reflection.dir/main.cpp.o: In function `main':
PATH/clspv/tools/reflection/main.cpp:114: undefined reference to `spvtools::SpirvTools::SpirvTools(spv_target_env)'
PATH/clspv/tools/reflection/main.cpp:115: undefined reference to `spvtools::SpirvTools::Validate(std::vector<unsigned int, std::allocator<unsigned int> > const&) const'
PATH/clspv/tools/reflection/main.cpp:117: undefined reference to `spvtools::SpirvTools::~SpirvTools()'
tools/reflection/CMakeFiles/clspv-reflection.dir/ReflectionParser.cpp.o: In function `clspv::ParseReflection(std::vector<unsigned int, std::allocator<unsigned int> > const&, spv_target_env, std::ostream*)':
PATH/clspv/tools/reflection/ReflectionParser.cpp:337: undefined reference to `spvtools::Context::Context(spv_target_env)'
PATH/clspv/tools/reflection/ReflectionParser.cpp:338: undefined reference to `spvtools::Context::SetMessageConsumer(std::function<void (spv_message_level_t, char const*, spv_position_t const&, char const*)>)'
PATH/clspv/tools/reflection/ReflectionParser.cpp:341: undefined reference to `spvtools::Context::CContext()'
PATH/clspv/tools/reflection/ReflectionParser.cpp:337: undefined reference to `spvtools::Context::~Context()'
```

These symbols are present in both `libSPIRV-Tools.a` and `libSPIRV-Tools-shared.so` but they are defined as local in the latter.

Yet, these are declared in `include/spirv-tools/libspirv.hpp` so I would assume they should be exported. Would you expect the same? If so, I can create a bug report against the SPIR-V Tools.

mantognini:
Thanks for merging.

Re. visibility of symbols: after digging into SPIR-V Tools issues, I've found [this comment](https://github.com/KhronosGroup/SPIRV-Tools/issues/3214#issuecomment-615320065) which seems relevant (i.e. confirming the symbols are expected to be visible):

> `SPIRV-Tools` can be properly generated and consumed as a shared lib (with C and C++ API) if `SPIRV_TOOLS_EXPORT` is added in front of its C++ API (in `include/spirv-tools/libspirv.hpp`)

dneto0:
> Yet, these are declared in include/spirv-tools/libspirv.hpp so I would assume they should be exported. Would you expect the same? If so, I can create a bug report against the SPIR-V Tools.

Yes, they should be exported. Thanks for finding this!

Issue #624, opened 2020-08-27
Invalid vector typecast with select operators


gnl21:
When compiling the test kernel:
```
__kernel void ptr_cast_fail(float4 a, float4 b)
{
  float4 o = b > a ? a : (b <= a ? b : a);
}
```
I get a crash inside LLVM:
```
clspv: third_party/llvm/llvm/lib/IR/Instructions.cpp:2985: static llvm::CastInst* llvm::CastInst::CreatePointerCast(llvm::Value*, llvm::Type*, const llvm::Twine&, llvm::Instruction*): Assertion `S->getType()->isPtrOrPtrVectorTy() && "Invalid cast"' failed.
```
The problem does not occur if the vectors are switched for floats.

mantognini:
Interesting timing, I came across a variant of this issue too and was about to create a PR. It seems my patch solves this crash.

gnl21:
Excellent. I was planning on looking into it but I haven't had time yet, so the timing is perfect for me! Thanks.

PR #625, opened 2020-08-27
Improve SimplifyPointerBitcastPass

mantognini:
Ignore non-trivial casts on non-pointer types.
Add test covering this.

Developed independently of #624 but seems to address the issue as well.

### DIFF
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -216,8 +216,9 @@ bool SimplifyPointerBitcastPass::runOnBitcastFromBitcast(Module &M) const {
       Bitcast->replaceAllUsesWith(OtherBitcast);
     } else {
       // Create a new bitcast from the other bitcasts argument to our type.
-      auto NewBitcast = CastInst::CreatePointerCast(
-          OtherBitcast->getOperand(0), Bitcast->getType(), "", Bitcast);
+      auto NewBitcast =
+          CastInst::Create(Instruction::BitCast, OtherBitcast->getOperand(0),
+                           Bitcast->getType(), "", Bitcast);
 
       // And replace the original bitcast with our replacement bitcast.
       Bitcast->replaceAllUsesWith(NewBitcast);
--- a/test/PointerCasts/non_pointer_cast.ll
+++ b/test/PointerCasts/non_pointer_cast.ll
@@ -1,0 +1,19 @@
+; RUN: clspv-opt --SimplifyPointerBitcast %s -o %t
+; RUN: FileCheck %s < %t
+
+; Ensure the pass doesn't crash on this input and properly simplifies the cast.
+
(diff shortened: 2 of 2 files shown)


test/PointerCasts/non_pointer_cast.ll
```diff
@@ -0,0 +1,19 @@
...
+; RUN: FileCheck %s < %t
+
+; Ensure SimplifyPointerBitcast doesn't crash on this input and doesn't modify the casts.
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+define spir_func void @test(float %x) {
+entry:
+  %y = bitcast float %x to i32
```
alan-baker, line 11: This should be handled and generate:
```
%cast = bitcast float %x to <2 x i16>
```

lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -196,11 +196,15 @@ bool SimplifyPointerBitcastPass::runOnBitcastFromBitcast(Module &M) const {
...
         // If we have a bitcast instruction...
-        if (auto Bitcast = dyn_cast<BitCastInst>(&I)) {
+        if (auto *Bitcast = dyn_cast<BitCastInst>(&I)) {
           // ... whose source is a bitcast instruction...
-          if (isa<BitCastInst>(Bitcast->getOperand(0))) {
-            // ... record the bitcast as something we need to process.
-            WorkList.push_back(Bitcast);
+          if (auto *OtherBitcast = dyn_cast<BitCastInst>(Bitcast->getOperand(0))) {
+            // ... record the bitcast as something we need to process...
+            if (OtherBitcast->getType() == Bitcast->getType() ||
```
alan-baker, line 203: This seems overly conservative.  The real problem seems to be the way the cast instruction is created. It should instead use `CastInst::Create(Instruction::BitCast,...)`.

mantognini: That works, indeed. I wonder though, should the pass be renamed to, say, `SimplifyBitcastPass`? The current name is the main reason my patch ignores non-trivial, non-pointer casts.

alan-baker: It had already grown past that point with the trivial bitcast transformation. I'm ok to leave as is or change it in this PR.

test/PointerCasts/non_pointer_cast.ll
mantognini: The test has been updated.

mantognini (commented):
I've updated the code as discussed. I ended up leaving out the file/class renaming as I wasn't 100% sure of which part of the doc/comments needed to be updated as well and didn't want to drag this PR for too long.

alan-baker:
@mantognini, there are some minor format errors.

mantognini:
Oh, yes, sorry about that. Should be fixed now.

PR #626, opened 2020-08-27
Implement a replacement for log1p function

### LINKED ISSUE #623: log1p not implemented
The CL builtin function log1p is not currently implemented and is used in a kernel that I'm trying to compile. I've implemented the function in the obvious way as log(1+x) and was going to make a PR, but a couple of questions:
* The implementation matches the precision of fast-math, but it does sort of defeat the object of having this function in the first place. Do you want the function implemented in that way?
 * I found two places in the compiler where derived functions are implemented (ReplaceCLBuiltins and during SPIR-V code generation). Is there a reason to prefer one over the other? I see that, for example, log10 is implemented in replace builtins, while asinpi is implemented during code generation. (FWIW, my implementation is based on that of log10, but could easily be changed).

gnl21:
Implemented as "log(1+x)", so may suffer from precision problems.

Fixes #623.


gnl21:
OK, great, thanks. I was going to update the tests and documentation, I just wanted to make sure it was going to be acceptable before doing that.

test/MathBuiltins/log/float2_log1p.cl
```diff
@@ -0,0 +1,19 @@
+// RUN: clspv %s -o %t.spv
```
alan-baker, line 1: Sorry for the hassle, but we're trying to move to targeted tests using `clspv-opt` where possible. Could you please convert these to be based on LLVM IR and just running ReplaceOpenCLBuiltins? The tests look correct.

alan-baker:
Also, you'll need to sign the CLA.

test/MathBuiltins/log/float2_log1p.cl
gnl21: Sure. I'm on vacation this week but I'll update the tests and get the  CLA sorted with our POC at the start of next week.

Could you point me to an example of a test that's in the correct new style?

alan-baker: Sure, take a look at the tests in `test/MathBuiltins/isfinite` (e.g. [isfinite_half2.ll](https://github.com/google/clspv/blob/master/test/MathBuiltins/isfinite/isfinite_half2.ll)).

gnl21:
@googlebot I signed it!

test/MathBuiltins/log/float2_log1p.cl
gnl21: I've changed the tests to LLVM IR.

alan-baker (approved):
Just need to fix the formatting.

alan-baker (approved):

PR #630, opened 2020-09-01
Support mixed reqd_work_group_size attributes

### LINKED ISSUE #629: Support mixed reqd_work_group_size attributes
Clspv doesn't properly handle a module with some kernels that have required workgroup size and others that don't. This is often only apparent on debug builds when llvm_unreachable is encountered; however, clspv should not generate LocalSize execution modes when the WorkgroupSize builtin also exists in the module (since it is overridden).

alan-baker:
Fixes #629

* Generate WorkgroupSize builtin if not all kernels specify
  reqd_work_group_size
  * Do not generate LocalSize execution mode if WorkgroupSize builtin is
    generated
  * new tests

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2399,8 +2399,8 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
           PrevZDimCst = CurZDimCst;
         } else if (CurXDimCst != PrevXDimCst || CurYDimCst != PrevYDimCst ||
                    CurZDimCst != PrevZDimCst) {
-          llvm_unreachable(
-              "reqd_work_group_size must be the same across all kernels");
+          HasMDVec.push_back(false);
+          continue;
         } else {
           continue;
         }
@@ -2439,8 +2439,8 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
       HasMD = HasMDVec[0];
       for (uint32_t i = 1; i < HasMDVec.size(); i++) {
         if (HasMD != HasMDVec[i]) {
-          llvm_unreachable(
-              "Kernels should have consistent work group size definition");
+          HasMD = false;
+          break;
         }
       }
     }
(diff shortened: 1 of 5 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -2439,8 +2439,8 @@ void SPIRVProducerPass::GenerateGlobalVar(GlobalVariable &GV) {
       HasMD = HasMDVec[0];
       for (uint32_t i = 1; i < HasMDVec.size(); i++) {
         if (HasMD != HasMDVec[i]) {
-          llvm_unreachable(
-              "Kernels should have consistent work group size definition");
+          HasMD = false;
```
jrprice, line 2442: If I've understood the logic correctly, I think this can all be simplified. We generate spec constants if there are any kernels that don't use `reqd_work_group_size`, or if there are conflicting values. You don't need this loop or a `bool` vector, you just need a single `bool` to track whether you can use a constant wgsize, initialize it to `true`, and set it to `false` when you encounter a kernel that hasn't got `reqd_work_group_size` or if there's a conflict.

jrprice (changes requested):

lib/SPIRVProducerPass.cpp
alan-baker: Conflicting values are ok. If you take a look at some of the tests we can generate `LocalSize` execution mode. It's only when some kernels don't have it we need to have the spec constants (and no execution mode). I'll double check the logic here.

jrprice: I think this part of the code is only hit if the kernels use the work-group size though (e.g. they call `get_local_size()`), and if they do (and they're using different `reqd_work_group_size`) then we don't generate `LocalSize` execution mode. Maybe worth adding a test case for that scenario.

I'm not disagreeing with the logic that's here - I think it's correct - I'm just saying the same thing can be written without the vector and the loop.

alan-baker: We add the global variable (which is what this deals with) if any kernel does not use required workgroup size **or** if one of the builtins (e.g. get_local_size()) that requires it is used.

In any event, I'll double check this logic.

alan-baker: I've removed the loop.

jrprice (approved):

PR #638, opened 2020-09-09
Introduce stubs for long-vector support

mantognini:
This relates to #613.

The first commit improves the existing checks to reject usage of long-vectors. The second one introduces the `--long-vector` option and a new pass to lower long-vectors (for now, doing nothing), plus registers that pass when appropriate. The frontend is updated to accept long-vectors when the option is given on the command line.


lib/LongVectorLoweringPass.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+bool LongVectorLoweringPass::runOnModule(Module &) {
+  // TODO implement long-vector lowering.
+  bool Modified = false;
+  return Modified;
+}
+
+} // namespace
+
+static RegisterPass<LongVectorLoweringPass>
```
alan-baker, line 47: Please use the initialization macro as in the other files (adding to passes.{cpp,h}). I've changed all other passes to do that and is easier to work with for some internal builds.

lib/FrontendPlugin.cpp
```diff
@@ -550,6 +558,20 @@ struct ExtraValidationConsumer final : public ASTConsumer {
                            "%select{SSBOs|UBOs|push constants}0");
   }
 
+  bool IsLongVectorType(QualType type) {
```
alan-baker, line 561: Why does this only check pointers or vectors? What if the kernel argument has a long vector in a struct or array? If they are meant to be supported, add a comment to indicate the difference otherwise perhaps just reuse IsSupportedType but provide it an extra argument.

Please also add tests to cover structs and arrays to cover those cases.

lib/Option.cpp
```diff
@@ -151,6 +151,10 @@ llvm::cl::opt<bool> keep_unused_arguments(
 llvm::cl::opt<bool> int8_support("int8", llvm::cl::init(true),
                                  llvm::cl::desc("Allow 8-bit integers"));
 
+llvm::cl::opt<bool>
+    long_vector_support("long-vector", llvm::cl::init(false),
```
alan-baker, line 155: Please add that this is experimental.

lib/Compiler.cpp
```diff
@@ -607,6 +607,9 @@ int PopulatePassManager(
     pm->add(clspv::createClusterPodKernelArgumentsPass());
   }
   pm->add(clspv::createReplaceOpenCLBuiltinPass());
+  if (clspv::Option::LongVectorSupport()) {
```
alan-baker, line 610: If there are any phase-ordering dependencies you are aware of, please add a comment to indicate the restrictions here.

alan-baker (changes requested):

lib/Compiler.cpp
mantognini: Will do. Only ReplaceOpenCLBuiltinPass is required to my knowledge.

lib/FrontendPlugin.cpp
mantognini: This is an oversight that I'll address, thanks for bringing it up. I'll add the tests too.

lib/LongVectorLoweringPass.cpp
mantognini: The downside of this macro is that we cannot use `--print-before/after=LongVectorLowering`. Are you aware of a workaround that I can use with the macro?

lib/Option.cpp
mantognini: Will do.

mantognini (commented):
Thanks for your feedback, I'll provide an updated patch shortly.

lib/LongVectorLoweringPass.cpp
alan-baker: I usually just dump everything into a file and scan for the particular pass.

lib/FrontendPlugin.cpp
mantognini: When addressing this, I realised `IsSupportedType` didn't cover all cases. The situation should be a bit better now, although there's an `llvm_unreachable` for situation not covered by the existing and additional test-cases.

`IsLongVectorType` has now been removed in favour of using `IsSupportedType`.

lib/LongVectorLoweringPass.cpp
mantognini: I see. I've updated the code to use the macro.

lib/FrontendPlugin.cpp
```diff
@@ -181,19 +184,29 @@ struct ExtraValidationConsumer final : public ASTConsumer {
...
       // Otherwise check recursively.
-      return IsSupportedType(Ty->getPointeeType(), SR);
+      return IsSupportedType(Ty->getPointeeType(), SR, IsKernelParameter);
     }
 
     const auto &canonicalType = QT.getCanonicalType();
     if (auto *VT = llvm::dyn_cast<ExtVectorType>(canonicalType)) {
-      // We don't support vectors with more than 4 elements.
+      // We don't support vectors with more than 4 elements under all
+      // circumstancies.
```
alan-baker, line 193: circumstancies -> circumstances

alan-baker (approved):
Approved with nits.

lib/FrontendPlugin.cpp
mantognini: Indeed. Fixed now.

PR #640, opened 2020-09-10
Fix how ArrayStride is generated for pointers

### LINKED ISSUE #637: ArrayStride decoration not transferred to function parameters
For the test kernel:
```
static int test_func(const int idx, global const unsigned char (*const d)[6])
{
  return d[idx][0];
}

kernel void k1(write_only image2d_t out, global float *padding, global const unsigned char (*const data)[6])
{
  int d = test_func(0, data);
  write_imagei(out, (int2)(0, 0), d);
}

kernel void k2(write_only image2d_t out, global const unsigned char (*const data)[6])
{
  int d = test_func(0, data);
  write_imagei(out, (int2)(0, 0), d);
}
```
clspv appears to generate invalid SPIR-V for `test_func`. The function parameter `d` is not decorated with `ArrayStride` so the consumer can't generate the correct memory accesses.

If I simplify the kernel (remove either `k1` or `k2`, or remove the `padding` argument) then the SPIR-V is transformed so that `data` is not passed as an argument at all, working around the problem.

alan-baker:
Fixes #637

* Pointer type for the base operand of an OpPtrAccessChain requires
  array stride, not the result of the pointer access chain
  * added some new tests
* Fixed various lint warnings
  * use empty() instead of size() compared to zero
  * remove some empty lines
  * remove unnecessary semi-clolons
* Fix typo in a test

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -5252,8 +5252,10 @@ Value *SPIRVProducerPass::GetBasePointer(Value *v) {
 bool SPIRVProducerPass::sameResource(Value *lhs, Value *rhs) const {
   if (auto *lhs_call = dyn_cast<CallInst>(lhs)) {
     if (auto *rhs_call = dyn_cast<CallInst>(rhs)) {
-      const auto &lhs_func_info = Builtins::Lookup(lhs_call->getCalledFunction());
-      const auto &rhs_func_info = Builtins::Lookup(rhs_call->getCalledFunction());
+      const auto &lhs_func_info =
+          Builtins::Lookup(lhs_call->getCalledFunction());
+      const auto &rhs_func_info =
+          Builtins::Lookup(rhs_call->getCalledFunction());
       if (lhs_func_info.getType() == Builtins::kClspvResource &&
           rhs_func_info.getType() == Builtins::kClspvResource) {
         // For resource accessors, match descriptor set and binding.
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -152,10 +152,10 @@ struct SPIRVOperand {
     }
   }
 
-  SPIRVOperandType getType() const { return Type; };
-  uint32_t getNumID() const { return LiteralNum[0]; };
-  std::string getLiteralStr() const { return LiteralStr; };
-  const uint32_t *getLiteralNum() const { return LiteralNum; };
+  SPIRVOperandType getType() const { return Type; }
+  uint32_t getNumID() const { return LiteralNum[0]; }
+  std::string getLiteralStr() const { return LiteralStr; }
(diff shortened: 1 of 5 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3899,21 +3896,22 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
...
         // generation until later, so we only make one decoration.
-        getTypesNeedingArrayStride().insert(ResultType);
+        getTypesNeedingArrayStride().insert(GEP->getPointerOperandType());
+        break;
+      case spv::StorageClassWorkgroup:
         break;
       default:
+        llvm_unreachable(
+            "OpPtrAccessChain is not supported for this storage class");
         break;
```
alan-baker, line 3915: This is the actual bugfix.

lib/SPIRVProducerPass.cpp
```diff
@@ -3899,21 +3896,22 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
...
-      // for something else in the SPIR-V.
-      // E.g. see test/PointerAccessChain/pointer_index_is_constant_1.cl
+      // Shader validation in the SPIR-V spec requires that the base pointer to
+      // OpPtrAccessChain (in StorageBuffer storage class) be decorated with
+      // ArrayStride.
       auto address_space = ResultType->getAddressSpace();
       setVariablePointersCapabilities(address_space);
       switch (GetStorageClass(address_space)) {
       case spv::StorageClassStorageBuffer:
-      case spv::StorageClassUniform:
```
jrprice, line 3911: What's the reason for removing the `StorageClassUniform` case? AFAICT the shader validation rule isn't specific to `StorageBuffer`.

lib/SPIRVProducerPass.cpp
```diff
@@ -3899,21 +3896,22 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
...
         // Save the need to generate an ArrayStride decoration.  But defer
         // generation until later, so we only make one decoration.
-        getTypesNeedingArrayStride().insert(ResultType);
+        getTypesNeedingArrayStride().insert(GEP->getPointerOperandType());
+        break;
+      case spv::StorageClassWorkgroup:
```
jrprice, line 3910: Just curious: do we ever generate `OpPtrAccessChain` for `Workgroup` storage class (presumably we don't have test coverage for this)?

lib/SPIRVProducerPass.cpp
alan-baker: Variables pointers only work with StorageBuffer and Workgroup storage classes and OpPtrAccessChain generates a variable pointer so clspv is generating invalid code if it does generate OpPtrAccessChain for a Uniform pointer. No point generating array stride.

jrprice (approved):

PR #643, opened 2020-09-22
Read write images

alan-baker:
Support read-write images.

* When targeting OpenCL 2.0 or later, treat write-only images as read-write images to prevent duplicate types
* Read-write images are translated as storage images (same as write-only images)
* Add support for CLK_IMAGE_MEM_FENCE for mem fence and barriers
* rename ArgKind::ReadOnlyImage and ArgKind::WriteOnlyImage to ArgKind::SampledImage and ArgKind::StorageImage respectively
* tests

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -1258,8 +1258,8 @@ void SPIRVProducerPass::FindTypesForResourceVars() {
   // To match older clspv codegen, generate the float type first if required
   // for images.
   for (const auto *info : ModuleOrderedResourceVars) {
-    if (info->arg_kind == clspv::ArgKind::ReadOnlyImage ||
-        info->arg_kind == clspv::ArgKind::WriteOnlyImage) {
+    if (info->arg_kind == clspv::ArgKind::SampledImage ||
+        info->arg_kind == clspv::ArgKind::StorageImage) {
       if (IsIntImageType(info->var_fn->getReturnType())) {
         // Nothing for now...
       } else if (IsUintImageType(info->var_fn->getReturnType())) {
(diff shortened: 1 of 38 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3351,8 +3400,10 @@ SPIRVProducerPass::GenerateImageInstruction(CallInst *Call,
       Ops.clear();
     }
     Ops << Image << Coordinate << TexelID;
-
     RID = addSPIRVInst(spv::OpImageWrite, Ops);
+
+    // Image writes require StorageImageWriteWithoutFormat.
+    addCapability(spv::CapabilityStorageImageWriteWithoutFormat);
```
dneto0, line 3406: This looks like a bugfix.

lib/SPIRVProducerPass.cpp
```diff
@@ -1661,19 +1661,20 @@ SPIRVID SPIRVProducerPass::getSPIRVType(Type *Ty) {
...
                  STy->getName().startswith("opencl.image2d_wo_t") ||
                  STy->getName().startswith("opencl.image2d_array_ro_t") ||
+                 STy->getName().startswith("opencl.image2d_array_rw_t") ||
                  STy->getName().startswith("opencl.image2d_array_wo_t") ||
                  STy->getName().startswith("opencl.image3d_ro_t") ||
+                 STy->getName().startswith("opencl.image3d_rw_t") ||
                  STy->getName().startswith("opencl.image3d_wo_t")) {
-
-        if (STy->getName().contains("_wo_t")) {
-          addCapability(spv::CapabilityStorageImageWriteWithoutFormat);
```
dneto0, line 1675: Ah, you're just making this more specialized to the cases where it's needed.
Sounds good.

dneto0 (approved):
Thanks!

PR #646, opened 2020-10-08
Lower function parameters and return type

mantognini:
This brings support for user-defined functions that have long-vectors as parameters and/or return type, improving support for lowering long-vectors (#613).

New tests cover this feature but also increase testing coverage of `BinaryOperators` with `fadd` and `fdiv`.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -91,6 +91,9 @@ private:
   /// by their removal.
   void cleanDeadInstructions();
 
+  /// Remove all long-vector functions that were lowered.
+  void cleanDeadFunctions();
+
 private:
   /// A map between long-vector types and their equivalent representation.
   DenseMap<Type *, Type *> TypeMap;
@@ -275,6 +278,8 @@ bool LongVectorLoweringPass::runOnModule(Module &M) {
   for (auto &F : M.functions()) {
     Modified |= runOnFunction(F);
   }
+
+  cleanDeadFunctions();
 
   return Modified;
 }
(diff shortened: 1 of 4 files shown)


alan-baker:
I'm confused by the scope of this PR. Is it intentional that calls are not handled? Right now, the pass lowers an equivalent function, but can only remove unused functions. So if the original function was called, both versions stick around.

Also the comments previously indicated that only non-kernel functions would be replaced, but there is no distinction made in the pass that I can see.  Is intentional to handle kernel functions now?

Separately, I will be away next week so expect this PR to get stalled temporarily.

mantognini:
> Is it intentional that calls are not handled? Right now, the pass lowers an equivalent function, but can only remove unused functions. So if the original function was called, both versions stick around.

Yes, calls will be handled by a later PR. It was a bit hard and time consuming to split features in fully independent steps while keeping size of PRs smallish so it's indeed a bit awkward that calling such function should result in an assertion when removing the original version:

```
  // L594-595:
  assert(DeadFunctions.empty() &&
         "Not all supposedly-dead functions were removed!");
```

Implementing support for function lowering early let us write tests more concisely. In two commits I'll also uncomment the `llvm_unreachable` in `LongVectorLoweringPass::visit` so `CallInst` will be more cleanly rejected. Support for calls will follow very closely in another PR. I should probably have mentioned that in the PR description, sorry for the confusion.

Adding support for calls at this stage is not impossible, of course, but it would imply moving a significant amount of code around and some local redesign, given that the subsequent commits rely on `InstVisitor` to handle instructions in a modular fashion, as well as dealing with merge conflicts. So if it is alright with you I'd rather keep it out of the scope of this PR for now.

> Also the comments previously indicated that only non-kernel functions would be replaced, but there is no distinction made in the pass that I can see. Is intentional to handle kernel functions now?

Yes, it is. The pass itself relies on the frontend to reject unsupported code in that matter. For the pass, kernel and non-kernel functions are the same and no specific distinctions are made.

My initial concerns about supporting kernel with long-vector parameters were more on the communication aspects between the runtime and the compiler (which would lie about the actual types). Recent experiments with clvk have shown this not to be an issue, so I should be able to provide a PR later on that removes this restriction on kernel parameters -- by mainly modifying the frontend checks.

alan-baker (approved):

PR #650, opened 2020-10-14
Update LLVM

jrprice:
* InstCombine replaces the code we generate for the rotate builtin
  with an llvm.fshl intrinsic, so handle that in ReplaceLLVMIntrinsicsPass

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -6,7 +6,7 @@
       "subrepo" : "llvm/llvm-project",
       "branch" : "master",
       "subdir" : "third_party/llvm",
-      "commit" : "7238faa4ae977523903192e287d442eb53c49ee5"
+      "commit" : "b21ad3b66bce942ee6e0f5b1fcfdea31928005a7"
     },
     {
       "name" : "SPIRV-Headers",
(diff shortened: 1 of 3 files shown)


jrprice:
The code in `replaceFshl` is almost the same as the code in `replaceRotate`, but slightly more general (`rotate` is essentially a specialized version of `llvm.fshl` where the first two arguments are the same).

We could avoid the code duplication by pre-empting `InstCombine` and just having `replaceRotate` generate a `llvm.fshl` directly, knowing that we run `ReplaceLLVMInstrincis` later in the flow. I can do this in a follow up PR if this sounds reasonable.

dnovillo (changes requested):
Are there any existing tests that will exercise this transformation?

jrprice:
> Are there any existing tests that will exercise this transformation?

The existing `rotate` tests cover it indirectly, but I've just added some additional tests that use it directly as well.

dnovillo:
> > Are there any existing tests that will exercise this transformation?
> 
> The existing `rotate` tests cover it indirectly, but I've just added some additional tests that use it directly as well.

Thanks!  LGTM.

dnovillo (approved):

PR #653, opened 2020-10-19
Add support for -cl-std=CL3.0

kpet:
Always enable support for non-uniform work-groups when compiling as
CL3.0.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -139,6 +139,7 @@ enum class SourceLanguage {
   OpenCL_C_11,
   OpenCL_C_12,
   OpenCL_C_20,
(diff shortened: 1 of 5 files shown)


kpet:
I don't see clspv gaining full support for the generic address space any time soon (clspv currently requires inlining entry points and hoping that address space inference can remove all uses) and generic address space is optional in OpenCL 3 so I thought that it would be cleaner to disable it initially. I think we'll probably want command line options ultimately for features that are optional in OpenCL 3 but the default IMHO should be OpenCL C 1.2 + what we can support well.

alan-baker (approved):

PR #655, opened 2020-10-21
Recusiverly lower instructions

mantognini:
The first commit is almost a NFC. It refactors the code to use InstVisitor to make the pass more modular. In future commits, support for a new instruction will be added with a new method overriding InstVisitor's default behaviour. The only change is that the pass will no longer print "Value not handled" (or any other message) for instruction that do not need handling, as determined by `handlingRequired`.

The second commit recursively visits instructions and lowers their operands. It adds support for a few additional kind of instructions and constants, with tests covering these, that, together with the existing support for BinaryOperator, allow turning on assertion/llvm_unreachable to alert users about features not yet supported.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -60,11 +60,17 @@ private:
   /// Returns nullptr if no lowering is required.
   Value *visit(Value *V);
 
+  /// Visit Constant. This is not provided by InstVisitor.
+  Value *visitConstant(Constant &Cst);
+
   /// InstVisitor impl, general "catch-all" function.
   Value *visitInstruction(Instruction &I);
 
   // InstVisitor impl, specific cases.
   Value *visitBinaryOperator(BinaryOperator &I);
+  Value *visitExtractElementInst(ExtractElementInst &I);
+  Value *visitInsertElementInst(InsertElementInst &I);
+  Value *visitShuffleVectorInst(ShuffleVectorInst &I);
 
 private:
   // Helpers for lowering values.
@@ -73,6 +79,13 @@ private:
   ///
   /// This only looks at the types involved, not the opcodes or anything else.
   bool handlingRequired(User &U);
+
+  /// Return the lowered version of @p U or @p U itself when no lowering is
+  /// required.
+  Value *visitOrSelf(Value *U) {
+    auto *V = visit(U);
+    return V ? V : U;
+  }
 
   /// Register the replacement of @p U with @p V.
   ///
@@ -88,7 +101,7 @@ private:
   Type *getEquivalentType(Type *Ty);
 
   /// Implementation details of getEquivalentType.
-  Type *getEquivalentTypeImpl(Type *Ty) const;
+  Type *getEquivalentTypeImpl(Type *Ty);
 
   /// Return the equivalent type for @p Ty or @p Ty if no lowering is needed.
   Type *getEquivalentTypeOrSelf(Type *Ty) {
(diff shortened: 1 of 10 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -291,44 +325,201 @@ Value *LongVectorLoweringPass::visit(Value *V) {
     return it->second;
   }
 
-  if (auto *BinOp = dyn_cast<BinaryOperator>(V)) {
-    if (auto *EquivalentType = getEquivalentType(BinOp->getType())) {
-      IRBuilder<> B(BinOp);
+  if (isa<Argument>(V)) {
```
alan-baker, line 328: `InstructionMap` is looking like a bit of a misnomer now. Maybe `ValueMap` or `RemappedValues`?

lib/LongVectorLoweringPass.cpp
```diff
@@ -291,44 +325,201 @@ Value *LongVectorLoweringPass::visit(Value *V) {
...
+    }
+
+    V = setScalar(V, Scalar, i);
+  }
+
+  registerReplacement(I, *V);
+  return V;
+}
+
+bool LongVectorLoweringPass::handlingRequired(Instruction &I) {
```
alan-baker, line 510: This might be better served taking a User as the parameter for reuse with constants. At some point I assume the lowering will have to break down aggregates that contain long vectors (e.g. arrays and structs) and I think this will be useful at that point.

lib/LongVectorLoweringPass.cpp
```diff
@@ -291,44 +325,201 @@ Value *LongVectorLoweringPass::visit(Value *V) {
...
+  auto *EquivalentTy = getEquivalentType(Cst.getType());
+  if (EquivalentTy == nullptr) {
+    return nullptr; // No replacement needed.
+  }
+
+  if (isa<UndefValue>(Cst)) {
+    return UndefValue::get(EquivalentTy);
+  }
+
+  if (auto *Vector = dyn_cast<ConstantDataVector>(&Cst)) {
```
alan-baker, line 365: I think you are missing constant null vectors. They should be handled now. 

Separately, I assume it is left to future changes to handle constant expressions?

lib/LongVectorLoweringPass.cpp
```diff
@@ -426,7 +634,8 @@ bool LongVectorLoweringPass::runOnFunction(Function &F) {
 
   bool Modified = (FunctionToVisit != &F);
   for (Instruction &I : instructions(FunctionToVisit)) {
-    Modified |= (visit(&I) != nullptr);
+    // Use the Value overload of visit to ensure cache is used.
+    Modified |= (visit(static_cast<Value *>(&I)) != nullptr);
```
alan-baker, line 638: Does LLVM's `cast` not suffice here? If it does it is preferable to `static_cast`.

alan-baker (changes requested):
Can you add some cases to cover an undef in the shuffle selector, constant null and other (non-undef) constants?

lib/LongVectorLoweringPass.cpp
alan-baker: I'm happy to track missing functionality in #613. To be clear, I'm totally ok with this coming together in small steps. These questions are more to help to me scope out the extent of a given PR.

mantognini: FYI, I've updated the description in #613.

mantognini (commented):
The latest commits should cover the additional functionalities and tests. Let me know if there's anything else I can improve.

alan-baker (approved):
Changes look good, thanks for the updates. Just small formatting issue.

mantognini:
Ha, yes. Should be fixed now.

PR #656, opened 2020-10-23
Implement add_sat

alan-baker:
* Modify strip_banned_opencl_features.py to not strip add_sat (or
  sub_sat) builtins
* Add an implementation of add_sat
  * uses three possible implementations
  * OpIAddCarry
  * extend, add and clamp
  * add with extra comparisons and clamping
* Add opcode support in the producer for OpIAddCarry


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2353,3 +2358,139 @@ bool ReplaceOpenCLBuiltinPass::replaceFract(Function &F, int vec_size) {
...
+        auto sclamp_callee = module->getOrInsertFunction(sclamp_name, func_ty);
+        auto clamp = CallInst::Create(sclamp_callee, {add, min, max}, "", Call);
+        result = CastInst::Create(Instruction::Trunc, clamp, ty, "", Call);
+      } else {
+        // Pseudo-code:
+        // c = a + b;
+        // if (a < 0 && b < 0)
+        //   c = c > 0 ? min : c;
+        // else if (a >= 0 && b >= 0)
+        //   c = c < 0 ? max : c;
```
jrprice, line 2446: I think you could simplify this a bit to something like:
```
// c = a + b;
// if (b < 0)
//   c = c > a ? min : c;
// else
//   c = c < a ? max : c;
```

jrprice (changes requested):

lib/ReplaceOpenCLBuiltinPass.cpp
alan-baker: Good catch, much simpler.

jrprice (approved):
LGTM.

PR #657, opened 2020-10-27
Add support for OpenCL 2.0 atomic functions

alan-baker:
* Support OpenCL 2.0 atomic functions (except flag related builtins)
* default ordering of sequentially consistent is strengthened as appropriate to acquire, release or acquire release
* Fixed builtins enum for `atomic_fetch_*` functions
  * previously there was just `kAtomicFetchKey` and `kAtomicFetchKeyExplicit` neither of which map to real functions
* Allows atomic types in the frontend plugin

Missing support:
* No support for atomic_flag
* no support for `all_svm_devices` or `all_devices` scopes
* Only supports int and uint types (due to Vulkan restrictions)
* atomic_compare_exchange_weak* is implemented as the strong version
* No initializer functions supported (macro or atomic_init)
* no support for atomic_work_item_fence


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -74,6 +76,116 @@ Type *getIntOrIntVectorTyForCast(LLVMContext &C, Type *Ty) {
...
+
+  // Constants for SPIR-V storage class semantics.
+  const auto UniformSemantics =
+      builder.getInt32(spv::MemorySemanticsUniformMemoryMask);
+  const auto WorkgroupSemantics =
+      builder.getInt32(spv::MemorySemanticsWorkgroupMemoryMask);
+
+  // Instead of sequentially consistent, use acquire, release or acquire
+  // release semnatics.
+  Value *base_order = SeqCstSemantics;
```
jrprice, line 117: Could just do ` = nullptr;` and remove the `SeqCstSemantics` declaration above since it's never actually used?

Unless the point was to highlight that we're replacing `SeqCst` with something else...

jrprice (approved):

PR #660, opened 2020-10-29
Implement hadd and rhadd

alan-baker:
Add implementations of hadd and rhadd.


jrprice (approved):
Not a blocker for this PR, but I wonder if it's worth grouping the tests so that all vector widths for a given type are in the same file. Once we throw 8- and 16-wide vectors into the mix we're looking at a large number of test files in total across all builtins.

(Incidentally, we could arguably enable 8- and 16-wide vectors for these tests already, since they don't go through the frontend.)

alan-baker:
I don't mind the extra files personally. The tests shard better that way. Also, if there is a break it is easier to localize the problem. As for the vectors, yes 8- and 16-wide vectors could be done already, but I was actually considering reducing which vector sizes are tested. The tests aren't particularly interesting between 2, 3 and 4 as it is, so there likely isn't much value in adding the 8- and 16-wide versions in the future.

PR #664, opened 2020-10-30
Fix non-32-bit clz

alan-baker:
* Implement non-32-bit clz in terms of 32-bit clz
  * GLSL extended instruction FindUMsb only supports 32-bit integers
* Add amber and unit tests


alan-baker:
The char and short versions are a little bit ugly, but the way clz is handled in the producer is a bit awkward. Longer term this should be refactored to generate the code a bit cleaner. Probably something like `InsertSPIRVOp` for glsl extended instructions. Also may want to first lower to `llvm.ctlz` and then handle that in llvm intrinsic replacement.

amber/run_tests.py
```diff
@@ -32,7 +32,9 @@
     'images/write_image2d_r32i.amber',
     'images/write_image2d_rg32i.amber',
     'images/write_image2d_rgba32i.amber',
-    'integer/add_sat_short.amber']
+    'integer/add_sat_short.amber',
+    'integer/clz_long.amber',
+    'integer/clz_short.amber']
```
jrprice, line 37: Missing the `clz_int.amber` script that you added?

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2745,3 +2749,58 @@ bool ReplaceOpenCLBuiltinPass::replaceAtomicCompareExchange(Function &F) {
...
+      ty = VectorType::get(ty, vec_ty->getElementCount());
+    }
+    auto clz_32bit_ty = FunctionType::get(ty, {ty}, false);
+    std::string clz_32bit_name = Builtins::GetMangledFunctionName("clz", ty);
+    auto clz_32bit =
+        F.getParent()->getOrInsertFunction(clz_32bit_name, clz_32bit_ty);
+    if (bitwidth < 32) {
+      // Extend the input to 32-bits and perform a clz.  The clz for 32-bit is
+      // tranlated as 31 - FindUMsb(in). Adjust to that result to the right
+      // size.
```
jrprice, line 2776: Couple of typos here: "tranlated" and "Adjust ~to~ that".

jrprice (approved):

amber/run_tests.py
alan-baker: No, those are swiftshader suppressions. Right now swiftshader doesn't support non-32-bit types so only those variants are filtered out.

jrprice: Ah sorry, I should've expanded the diff to get that context.

PR #665, opened 2020-10-31
Fix hadd/rhadd for signed types

kpet:
Now passes the OpenCL conformance tests.


alan-baker (approved):
Thanks for catching this.

PR #669, opened 2020-11-04
Implement ctz

alan-baker:
* Similar to clz, all versions are implemented in terms of 32-bit ctz
* 32-bit ctz is implemented as FindILsb, a comparison against -1 and a
  select
* Added lit and amber tests
  * amber tests cover the 0 cases not tested in CTS
* passes OpenCL CTS for short and int types


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2779,56 +2782,75 @@ bool ReplaceOpenCLBuiltinPass::replaceAtomicCompareExchange(Function &F) {
...
+      if (leading) {
+        // For clz, if clz(top) is 32, return 32 + clz(bot).
+        auto cmp = builder.CreateICmpEQ(top_func, c32);
+        auto adjust = builder.CreateAdd(bot_func, c32);
+        tmp = builder.CreateSelect(cmp, adjust, top_func);
+      } else {
+        // For ctz, if clz(bot) is 32, return 32 + ctz(top)
+        auto bot_cmp = builder.CreateICmpEQ(bot_func, c32);
+        auto adjust = builder.CreateAdd(top_func, c32);
+        tmp = builder.CreateSelect(bot_cmp, adjust, bot_func);
```
jrprice, line 2850: Could avoid the `icmp` and `select` by instead `or`ing the extended value with `(1 << bitwidth)` before calling the 32-bit ctz.

jrprice (approved):

lib/ReplaceOpenCLBuiltinPass.cpp
alan-baker: Good idea, making that change.

PR #670, opened 2020-11-05
Lower OpenCL vloadN and vstoreN

mantognini:
Continuing on #613, this time lowering a few additional kind of instructions (AllocaInst, CastInst, LoadInst and StoreInst) which enable lowering of OpenCL vload/vstore builtin functions.

I've also replaced `isa<ConstantAggregateZero>(Cst)` with `Cst.isZeroValue()` to cover more cases ellegantly.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -72,11 +72,15 @@ private:
   Value *visitInstruction(Instruction &I);
 
   // InstVisitor impl, specific cases.
+  Value *visitAllocaInst(AllocaInst &I);
   Value *visitBinaryOperator(BinaryOperator &I);
   Value *visitCallInst(CallInst &I);
+  Value *visitCastInst(CastInst &I);
   Value *visitExtractElementInst(ExtractElementInst &I);
   Value *visitInsertElementInst(InsertElementInst &I);
+  Value *visitLoadInst(LoadInst &I);
   Value *visitShuffleVectorInst(ShuffleVectorInst &I);
+  Value *visitStoreInst(StoreInst &I);
   Value *visitUnaryOperator(UnaryOperator &I);
 
 private:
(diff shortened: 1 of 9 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -503,6 +521,35 @@ Value *LongVectorLoweringPass::visitCallInst(CallInst &I) {
...
+  // We expect something to lower, or this function shouldn't have been called.
+  assert(((OriginalValue != EquivalentValue) ||
+          (OriginalDestTy != EquivalentDestTy)) &&
+         "nothing to lower");
+
+  Value *V = nullptr;
+  switch (I.getOpcode()) {
+  case Instruction::BitCast: {
+    IRBuilder<> B(&I);
+    V = B.CreateBitCast(EquivalentValue, EquivalentDestTy, I.getName());
```
alan-baker, line 539: This doesn't seem correct. You can't bitcast to/from aggregate types. So `bitcast <4 x i32> %in to <8 x i16>` would be invalid and need to be broken down to element-wise bitcasts and reconstructing the output aggregate.

alan-baker (changes requested):

lib/LongVectorLoweringPass.cpp
mantognini: Indeed, you're correct. I missed the fact that bitcast apply only to _non-aggregate_ first class types. I've addressed this by keeping the scope to pointers for now and introduced a new tests with some TODOs for later.

Mind the fact that other instructions (alloca, load, store) do not have a dedicated explicit test. They are covered implicitely by the OpenCL vload/vstore tests. Let me know if you feel they deserve dedicated tests in LLVM IR format.

alan-baker: I think it would be best if we had explicit coverage for each instruction in pass test.

mantognini: I've added a testfile for these three instructions and addressed an issue with alignment (that wasn't handled).

alan-baker (approved):

PR #681, opened 2020-12-02
Lower global variables

mantognini:
Next feature for #613 is support for global variables. This one should be a bit simpler than previous PR, but comments are always welcome.

### DIFF
--- a/test/LongVectorLowering/globals.ll
+++ b/test/LongVectorLowering/globals.ll
@@ -1,0 +1,56 @@
+; RUN: clspv-opt --LongVectorLowering --instcombine %s -o %t
+; RUN: FileCheck %s < %t
+; RUN: FileCheck --check-prefix=NEGATIVE %s < %t
+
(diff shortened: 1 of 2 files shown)


test/LongVectorLowering/globals.ll
```diff
@@ -0,0 +1,26 @@
+; RUN: clspv-opt --LongVectorLowering --instcombine %s -o %t
+; RUN: FileCheck %s < %t
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+@a_global = external addrspace(2) constant <8 x float>, align 32
```
alan-baker, line 7: It would be good to have a version of this test where the global variable has an initializer.

alan-baker (changes requested):
Please add a test to cover a GV with an initializer.

test/LongVectorLowering/globals.ll
mantognini: Good idea. I've just pushed a change to address that.

alan-baker (approved):

PR #682, opened 2020-12-07
Refactor some builtin support

alan-baker:
* Don't replace builtins or intrinsics if they have a definition
* add implmemetations for llvm.trunc
* small change to name mangling
* fix an odd debug assertion in spirvop when reusing another function

Implement clz and ctz via llvm.ctlz and llvm.cttz

* Remove run of llvm intrinsic replacement before opencl builtin
  replacement
* add cleaner methodology for replacement ReplaceLLVMIntrinsicsPass
  * convert fshl to this style
* All clz and ctz functions are replaced with equivalent llvm intrinsics
  * Replace llvm intrinsics implements non-32-bit versions
  * update tests
* spirv producer updated to handle more llvm intrinsics
* Rotate, clz and ctz still pass CTS

Refactor copysign support

* Replace opencl builtin with llvm intrinsic
* move copysign implementation to llvm intrinsic replacement

### DIFF
--- a/lib/Builtins.cpp
+++ b/lib/Builtins.cpp
@@ -281,7 +281,8 @@ std::string Builtins::GetMangledFunctionName(const char *name, Type *type) {
   if (auto *func_type = dyn_cast<FunctionType>(type)) {
     Type *last_arg_type = nullptr;
     for (auto *arg_type : func_type->params()) {
-      if (arg_type == last_arg_type) {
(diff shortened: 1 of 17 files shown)


jrprice (approved):
LGTM.

I made some suggestions for simplifications using alternative LLVM APIs that take care of vector vs non-vector cases automatically, but they don't have to block this PR.

PR #686, opened 2020-12-10
Lower CmpInst, SelectInst, and OpenCL select()

mantognini:
Continuing submission for #613, this adds support for some additional instructions. As always, feedback welcome.

---

Mark PHINode as unsupported for now. Test cases are required to
implement this feature.

Add tests covering icmp and the two forms of select instructions. Add
test covering OpenCL select builtin function. Update fast-math flag
test.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -31,6 +31,7 @@
 #include "Builtins.h"
 #include "Passes.h"
 
+#include <array>
 #include <functional>
 
 using namespace llvm;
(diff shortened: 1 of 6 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -720,6 +749,62 @@ Value *LongVectorLoweringPass::visitLoadInst(LoadInst &I) {
...
+
+  // We have two cases to handle here:
+  // - when the condition is a scalar to select one of the two long-vector
+  //   alternatives. In this case, we create a single select instruction.
+  // - when the condition is a long-vector, too. In this case, we do an
+  //   element-wise select and construct an aggregate for the result.
+  Value *V = nullptr;
+  if (EquivalentCondition->getType()->isSingleValueType()) {
+    IRBuilder<> B(&I);
+    V = B.CreateSelect(EquivalentCondition, EquivalentTrueValue,
```
alan-baker, line 782: This is slightly problematic at present. The SPIRV producer does not handle aggregate selections correctly. SPIR-V before 1.4 cannot select between aggregates; however, they are not usually generated in normal flows. There are two options I see:
1. Change this code to always scalarize
2. Fix the spirv producer

Fixing the producer is more general, but also more complicated as it would have to recursively scalarize the selection.

alan-baker (changes requested):

lib/LongVectorLoweringPass.cpp
mantognini: I see. I've opted for the first option for now as it was much simpler for me to implement.

alan-baker (approved):

alan-baker:
@mantognini just a small formatting issue.

mantognini:
Yes indeed. It should now be fixed.

PR #690, opened 2020-12-17
Lower additional OpenCL builtin functions

mantognini:
Continuing on #613, this covers builtins similar to max: exp, fma, fmax and fmin, as well as frexp, which has a different signature.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -270,6 +270,10 @@ Function *getBIFScalarVersion(Function &Builtin) {
 
   // TODO Add support for other builtins by providing testcases and listing the
   // builtins here.
+  case clspv::Builtins::kExp:
+  case clspv::Builtins::kFma:
+  case clspv::Builtins::kFmax:
+  case clspv::Builtins::kFmin:
   case clspv::Builtins::kMax: {
     // Scalarise all the input/output types. Here we intentionally do not rely
     // on getEquivalentType because we want the scalar overload.
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -274,20 +274,36 @@ Function *getBIFScalarVersion(Function &Builtin) {
   case clspv::Builtins::kFma:
   case clspv::Builtins::kFmax:
   case clspv::Builtins::kFmin:
+  case clspv::Builtins::kFrexp:
   case clspv::Builtins::kMax: {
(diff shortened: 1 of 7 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -270,20 +270,38 @@ Function *getBIFScalarVersion(Function &Builtin) {
...
-    SmallVector<Type *, 16> ParamTys;
+    SmallVector<Type *, 16> ScalarParamTys;
     for (auto &Param : Builtin.args()) {
-      // TODO Need support for other types, like pointers. Need test case.
-      assert(Param.getType()->isVectorTy());
-      ParamTys.push_back(Param.getType()->getScalarType());
+      auto *ParamTy = Param.getType();
+
+      Type *ScalarParamTy = nullptr;
+      if (ParamTy->isVectorTy()) {
```
alan-baker, line 286: I missed this for previously, but several of these (e.g. fmax) the second argument maybe be a scalar to be splatted. I would suggest flipping this around to test if the parameter is a pointer so that you can grab the right type in such situations. This will avoid needing the splat arg pass earlier in the flow.
```
if (ParamTy->isPointerTy()) {
  ...
} else {
  ScalarParamTy = ParamTy->getScalarTy();
}
```

This deserves a test.

lib/LongVectorLoweringPass.cpp
```diff
@@ -364,15 +382,27 @@ Value *convertVectorOperation(Instruction &I, Type *EquivalentReturnTy,
...
   IRBuilder<> B(&I);
   for (unsigned i = 0; i < Arity; ++i) {
     SmallVector<Value *, 16> Args;
     Args.resize(EquivalentArgs.size());
 
     for (unsigned j = 0; j < Args.size(); ++j) {
-      assert(EquivalentArgs[j]->getType()->isStructTy());
-      Args[j] = B.CreateExtractValue(EquivalentArgs[j], i);
+      auto *ArgTy = EquivalentArgs[j]->getType();
+      if (ArgTy->isStructTy()) {
```
alan-baker, line 397: This will also need updated to handle the vector and scalar mix case.

alan-baker (changes requested):

lib/LongVectorLoweringPass.cpp
mantognini: Good catch. I've updated the code & test to handle this.
(PS: I'll be away for a bit but if there is any other comments I'll address them when I'm back early next year.)

alan-baker (approved):

PR #694, opened 2021-01-05
Lower kernel parameters

mantognini:
This brings support for long-vector parameters for kernel functions. To implement this support, I added a limited support for GEPs -- I've updated #613 to list this as a current limitation of the prototype we've designed so far.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -1,4 +1,4 @@
-// Copyright 2020 The Clspv Authors. All rights reserved.
+// Copyright 2020-2021 The Clspv Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -752,6 +753,42 @@ Value *LongVectorLoweringPass::visitExtractElementInst(ExtractElementInst &I) {
...
+  // cases.
+#ifndef NDEBUG
+  {
+    assert(I.getNumIndices() == 1 &&
+           "Need test case to implement GEP with multiple indices.");
+
+    assert(I.isInBounds() && "Need test case to implement GEP not 'inbound'.");
+
+    auto *ResultTy = I.getResultElementType();
+    assert(ResultTy->isVectorTy() &&
```
alan-baker, line 767: Why the restriction that the whole vector must be accessed? The code below would seem to handle accessing an individual element just as well.

mantognini: This is really down to the OpenCL test cases I had at my disposal when working on the initial prototype: I used bugpoint to reduce inputs when constructs where not supported and only this form of GEP was generated.

I suppose someone more familiar with GEPs could generate further test cases by hand to drive the improvement of this implementation.

PR #698, opened 2021-01-11
Add stub for array types in the long vector pass

jrprice:
Without this the pass bails when it encounters any array type, even if it does not contain wide vectors.

CC @mantognini: hopefully this doesn't cause you any rebase pain. I assume there is a patch for array types coming later but adding this stub now means that I can start testing this pass on some workloads that I'm looking at now.


mantognini:
Yes, it's perfectly fine from my side.

alan-baker (approved):

PR #704, opened 2021-02-06
Change LLVM branch name to 'main' in deps.json

kpet:


alan-baker:
I'm in the process of doing this and updating llvm. The 3.0 feature macros were added and some minor changes are necessary.

PR #710, opened 2021-02-15
Lower arrays and more GEPs

mantognini:
Support additional forms for getelementptr instructions for the provided use cases: multiple indices and constant expressions.

When lowering globals, keep their name intact.

Related to #613.

### DIFF
--- a/test/LongVectorLowering/gep.ll
+++ b/test/LongVectorLowering/gep.ll
@@ -4,13 +4,80 @@
 target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir-unknown-unknown"
 
-define spir_func void @test() {
+define spir_func void @test1() {
 entry:
   %ptr = load <8 x float> addrspace(1)*, <8 x float> addrspace(1)** undef, align 4
   %data = getelementptr inbounds <8 x float>, <8 x float> addrspace(1)* %ptr, i32 undef
   ret void
 }
 
-; CHECK: getelementptr inbounds
-; CHECK-SAME: [[FLOAT8:{ float, float, float, float, float, float, float, float }]],
-; CHECK-SAME: [[FLOAT8]] addrspace(1)* {{%[^ ]+}}, i32 undef
+define dso_local spir_kernel void @test2([1 x <16 x half>] addrspace(1)* %in, <16 x half> addrspace(1)* %out) {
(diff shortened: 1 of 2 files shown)


test/LongVectorLowering/gep.ll
```diff
@@ -4,13 +4,67 @@
...
+  ret void
+}
+
+@global = external dso_local addrspace(3) global [1 x <8 x float>], align 32
+; CHECK: @global = external dso_local addrspace(3) global
+; CHECK-SAME: [1 x [[FLOAT8:{ float, float, float, float, float, float, float, float }]]], align 32
+
+; This test covers ConstantExpr.
+define dso_local spir_kernel void @test4(<8 x float> addrspace(1)* %out) {
+  %ptr = getelementptr inbounds [1 x <8 x float>], [1 x <8 x float>] addrspace(3)* @global, i32 0, i32 undef
```
alan-baker, line 34: I'm confused how this covers constant expression gep? I would have expected the load to contain a contantexpr as the operand. It does cover a constant (global).

test/LongVectorLowering/gep.ll
```diff
@@ -4,13 +4,67 @@
...
+  %vec = load <8 x float>, <8 x float> addrspace(3)* %ptr, align 32
+  store <8 x float> %vec, <8 x float> addrspace(1)* %out, align 32
+  ret void
+}
+
+; This test covers ConstantExpr.
+define dso_local spir_kernel void @test5(<8 x float> addrspace(1)* %out) {
+  %ptr = getelementptr inbounds [1 x <8 x float>], [1 x <8 x float>] addrspace(3)* @global, i32 0, i32 0
```
alan-baker, line 42: Same issue here.

alan-baker (changes requested):
Code looks ok, just a question about the tests.

test/LongVectorLowering/gep.ll
mantognini: It got me confused as well but this form does result in an constant expression on the load when this IR is deserialised. If I comment out the code for constant gep, it fails.

I'll try to add a 6th test that explicitly use a constant expression as well, that shouldn't hurt.

mantognini: I've pushed the new test: [diff](https://github.com/google/clspv/compare/87f454da18278acdb3d9b693b945c29a40318356..747bfec47d92ee671d0f860caf0880b788515c41).

Let me know what you think of it.

alan-baker (approved):

PR #717, opened 2021-03-11
Fix compilation for 32-bit systems

mantognini:
Include `HandleLLVMOptions` to ensure LLVM and clspv components are compiled using the same defines (for example `-D_FILE_OFFSET_BITS=64` `-D_LARGEFILE_SOURCE`). This essentially fixes this undefined reference to
```
    clang::FileManager::getVirtualFile(llvm::StringRef, long, long)
```
while the following was actually defined
```
    clang::FileManager::getVirtualFile(llvm::StringRef, long long, long)
```
.

As a consequence, the compilation and link options are updated and require fixes to the dependency lists of `clspv_passes` and `clspv_core`.

The second commit address various warnings:

 * `-Wunused-function`: use `inline` instead of `static` in header.
 * `-Wunused-but-set-variable`: code is protected in `#ifndef NDEBUG` when it is only used for the purpose of assertions.
 * `-Wunused-function` and `-Wunused-but-set-variable`: dead code is removed.
 * `-Wsign-compare`: uses of `auto` are replaced with `unsigned` or `size_t`, or use `static_cast`.
 * `-Wimplicit-fallthrough`: this warning is silenced using special comments.
 * `-Wpedantic`: remove extra `;` on closing brace for namespace.
 * `-Woverflow`: use types explicitly sized (e.g. `std::uint64_t`) and appropriate macros for minimum-width integer constants instead of `size_t`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,10 @@ else()
 
   # Then pull in LLVM for building.
   add_subdirectory(${CLSPV_LLVM_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}/third_party/llvm EXCLUDE_FROM_ALL)
+
+  # Ensure clspv and LLVM use the same build options (e.g. -D_FILE_OFFSET_BITS=64).
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/third_party/llvm/llvm/cmake/modules/)
+  include(HandleLLVMOptions)
 endif()
 
 set(CLSPV_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -146,7 +146,7 @@ enum class SourceLanguage {
 SourceLanguage Language();
 
 // Returns true when the source language makes use of the generic address space.
-static bool LanguageUsesGenericAddressSpace() {
+inline bool LanguageUsesGenericAddressSpace() {
   return (Language() == SourceLanguage::OpenCL_CPP) ||
          ((Language() == SourceLanguage::OpenCL_C_20));
 }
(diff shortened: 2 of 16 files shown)


mantognini:
I should mention, the warnings I've addressed are for GCC 7. Here is the new compilation flags I've noticed:
```
common to the configurations noted below:
    -D__STDC_CONSTANT_MACROS
    -D__STDC_FORMAT_MACROS
    -D__STDC_LIMIT_MACROS
    -fPIC
    -fvisibility-inlines-hidden
    -Werror=date-time
    -Wall
    -Wextra
    -Wno-unused-parameter
    -Wwrite-strings
    -Wcast-qual
    -Wno-missing-field-initializers
    -pedantic
    -Wno-long-long
    -Wimplicit-fallthrough
    -Wno-noexcept-type
    -Wdelete-non-virtual-dtor
    -Wno-comment
    -fdiagnostics-color


shared, debug, Linux 64-bit, GCC 7.5:
    -D_DEBUG

static, release, Linux 64-bit, GCC 7.5:
    -ffunction-sections
    -fdata-sections

static, default, Linux Armv7, GCC 7.4:
    -D_FILE_OFFSET_BITS=64
    -D_LARGEFILE_SOURCE
    -ffunction-sections
    -fdata-sections
```

These options were identified by diffing the `compile_commands.json` that CMake generates with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` before and after this PR.

As expected, this flags will vary with other compilers so we should expect different warnings as well. I've noticed [clang](https://source.cloud.google.com/results/invocations/4e3a5808-21e3-4081-85a3-d27bbfd74008/targets/graphics_shader_compiler%2Fclspv%2Fmacos-clang-debug%2Fpresubmit/log) reports some more.

Would it be okay if this is addressed separately? I'm not sure I have the bandwidth at the moment to look into these.

alan-baker (approved):

PR #722, opened 2021-03-25
Add options for fp16 and fp64 and disable unsupported CL3 features

kpet:
Some OpenCL applications (e.g. clpeak) conditionally enable FP16/FP64 in
kernels based on the presence of the cl_khr_fp{16,64} macros. When using
the SPIR target Clang always enables all OpenCL extensions and features
potantially resulting in applications attempting to use features not
supported on the target device.

Since Clang always enables all features when using the SPIR target,
it means that clspv must either pass a list of features to be disabled
to Clang (not sure this would be accepted upstream) or disable features
after the target-specific defaults have been applied. The latter is the
approach proposed in this change: extensions and features are disabled
after the creation of the TargetInfo and before the CompilerInstance
is configured to use it.

The new options are enabled by default for backwards compatibility.

As a side note, I looked into the issue with cl_khr_3d_image_writes
that is described in #705 and can confirm it is a Clang bug that
is trivial to fix, I'll upstream a fix there.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -372,6 +372,52 @@ int ParseSamplerMap(const std::string &sampler_map,
   return 0;
 }
 
+clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
+  // Create target info
+  auto TargetInfo = clang::TargetInfo::CreateTargetInfo(
+      instance.getDiagnostics(),
+      std::make_shared<clang::TargetOptions>(instance.getTargetOpts()));
+
+  // The SPIR target enables all possible options, disable the ones we don't
+  // want
+  auto &Opts = TargetInfo->getSupportedOpenCLOpts();
+
+  // Conditionally disable extensions based on support
+  if (!clspv::Option::FP16()) {
+    Opts["cl_khr_fp16"] = false;
+  }
+  if (!clspv::Option::FP64()) {
+    Opts["cl_khr_fp64"] = false;
+  }
+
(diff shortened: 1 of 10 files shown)


lib/Compiler.cpp
```diff
@@ -372,6 +372,52 @@ int ParseSamplerMap(const std::string &sampler_map,
...
+    // TODO(kpet): This is a Clang bug (the pragma isn't enabled when the macro
+    // is defined)
+    Opts["cl_khr_3d_image_writes"] = false;
+
+    // The following features are never supported
+    Opts["__opencl_c_pipes"] = false;
+    Opts["__opencl_c_generic_address_space"] = false;
+    Opts["__opencl_c_atomic_order_seq_cst"] = false;
+    Opts["__opencl_c_device_enqueue"] = false;
+    Opts["__opencl_c_read_write_images"] = false;
```
alan-baker, line 406: These should actually work. There is code elsewhere that assumes read/write images for >= 2.0

alan-baker (changes requested):

lib/Compiler.cpp
kpet: Yes, you're right. I'd forgotten you'd added support.

kpet: Now, updated. PTAL.

alan-baker (approved):

kpet:
Sorry for that. Now fixed (and tested!). PTAL.

PR #734, opened 2021-04-10
Make it possible to specify workgroups as uniform

mehmetoguzderin:
This PR makes it possible to specify workgroups as uniform to disable some push constants' emission on frontends of variety.

# Why this hack (or unspecified flag)?

Using SPIR-V generated by `clspv` with C++ frontend enabled from simple Vulkan applications, having push constants emitted for "region offset" included causes vkCreateComputePipelines to throw an unknown error. When layout accommodates these push constant ranges, the error goes away, but this can be unexpected/unnecessary from the application's perspective. Despite what the application developer thinks, some (not all) drivers care about that part of the layout.

Maybe a more holistic (and nicer-looking) solution would be to "optimize away" these push constants when detected as not used by the module, but this works and is explicit.

### DIFF
--- a/lib/Option.cpp
+++ b/lib/Option.cpp
@@ -298,9 +298,6 @@ bool WorkDim() { return work_dim; }
 bool GlobalOffset() { return global_offset; }
 bool GlobalOffsetPushConstant() { return global_offset_push_constant; }
 bool NonUniformNDRangeSupported() {
-  assert(!(cl_arm_non_uniform_work_group_size && uniform_workgroup_size) &&
-         "uniform-workgroup-size overrides cl-arm-non-uniform-work-group-size. "
-         "Use only one.");
   return ((Language() == SourceLanguage::OpenCL_CPP) ||
           (Language() == SourceLanguage::OpenCL_C_20) ||
           (Language() == SourceLanguage::OpenCL_C_30) ||
@@ -333,6 +330,9 @@ bool NativeMath() { return cl_native_math; }
 
 bool FP16() { return fp16; }
 bool FP64() { return fp64; }
+
+bool ArmNonUniformWorkGroupSize() { return cl_arm_non_uniform_work_group_size; }
+bool UniformWorkgroupSize() { return uniform_workgroup_size; }
 
 } // namespace Option
 } // namespace clspv
--- a/lib/Option.cpp
+++ b/lib/Option.cpp
@@ -251,6 +251,10 @@ static llvm::cl::opt<bool>
          llvm::cl::desc(
              "Enable support for FP64 (cl_khr_fp64 and/or __opencl_c_fp64)."));
 
+static llvm::cl::opt<bool> uniform_workgroup_size(
+    "uniform-workgroup-size", llvm::cl::init(false),
+    llvm::cl::desc("Disable support for non-uniform workgroup size."));
+
 } // namespace
 
 namespace clspv {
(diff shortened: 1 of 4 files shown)


lib/Option.cpp
```diff
@@ -251,6 +251,10 @@ static llvm::cl::opt<bool>
          llvm::cl::desc(
              "Enable support for FP64 (cl_khr_fp64 and/or __opencl_c_fp64)."));
 
+static llvm::cl::opt<bool> uniform_workgroup_size(
+    "uniform-workgroup-size", llvm::cl::init(false),
+    llvm::cl::desc("Disable support for non-uniform workgroup size."));
```
alan-baker, line 256: A better description might be "Assume all workgroups are uniformly sized."

lib/Option.cpp
```diff
@@ -294,10 +298,11 @@ bool WorkDim() { return work_dim; }
...
 bool NonUniformNDRangeSupported() {
-  return (Language() == SourceLanguage::OpenCL_CPP) ||
-         (Language() == SourceLanguage::OpenCL_C_20) ||
-         (Language() == SourceLanguage::OpenCL_C_30) ||
-         cl_arm_non_uniform_work_group_size;
+  return ((Language() == SourceLanguage::OpenCL_CPP) ||
+          (Language() == SourceLanguage::OpenCL_C_20) ||
+          (Language() == SourceLanguage::OpenCL_C_30) ||
+          cl_arm_non_uniform_work_group_size) &&
+         !uniform_workgroup_size;
```
alan-baker, line 305: It should be an error to have both `cl_arm_non_uniform_work_group_size` and `uniform_workgroup_size` set.

alan-baker (changes requested):

lib/Option.cpp
mehmetoguzderin: Makes sense 👍

lib/Option.cpp
mehmetoguzderin: PTAL if my change does that as expected

alan-baker: Please remove the assert and instead add a check in `ParseOptions()` in lib/Compiler.cpp. You can see examples of error checking for option conflicts there, but the following should be sufficient (after the call to `ParseCommandLineOptions`):
```
if (cl_arm_non_uniform_work_group_size && uniform_workgroup_size) {
  llvm::errs() << "cannot enable Arm non-uniform workgroup extension support and assume uniform workgroup sizes\n";
  return -1;
}
```

mehmetoguzderin: @alan-baker oops, did I interpret it correctly in the last commit?

alan-baker (changes requested):
Latest change looks good, but I forgot to ask for a test. Could you please add a test that runs OpenCL C 2.0+ (or OpenCL C with C++) and the new option and checks that the reflection does not contain the any region offset or related push constant? Something along the lines of:

```
// RUN: clspv %s -o %t.spv -cl-std=CL2.0 -uniform-workgroup-size
// RUN: spirv-dis %t.spv -o %t.spvasm
// RUN: FileCheck %s < %t.spvasm
// RUN: spirv-val --target-env vulkan1.0 %t.spv

// verify this kernel would otherwise generate the region offset.
kernel void test(global int* out) {
  int gid = get_global_id(0);
  out[gid] = gid;
}

// CHECK-NOT: PushConstantRegionOffset
...
```

`test/Reflection/global_push_constants.cl` can be used as guide, this test will be simpler.

mehmetoguzderin:
@alan-baker I made an attempt with the latest commit, does it make the point as a test? (It did pass.)

alan-baker (changes requested):
Test looks, small nit about using the accessors everywhere.

lib/Option.cpp
```diff
@@ -294,10 +298,11 @@ bool WorkDim() { return work_dim; }
...
 bool GlobalOffsetPushConstant() { return global_offset_push_constant; }
 bool NonUniformNDRangeSupported() {
-  return (Language() == SourceLanguage::OpenCL_CPP) ||
-         (Language() == SourceLanguage::OpenCL_C_20) ||
-         (Language() == SourceLanguage::OpenCL_C_30) ||
-         cl_arm_non_uniform_work_group_size;
+  return ((Language() == SourceLanguage::OpenCL_CPP) ||
+          (Language() == SourceLanguage::OpenCL_C_20) ||
+          (Language() == SourceLanguage::OpenCL_C_30) ||
+          cl_arm_non_uniform_work_group_size) &&
```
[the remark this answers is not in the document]
mehmetoguzderin: @alan-baker looks cleaner with accessors indeed

alan-baker (approved):

PR #735, opened 2021-04-30
Replace llvm.{smax,smin} intrinsics with builtins

jrprice:
LLVM passes are generating these intrinsics from certain nested loop
patterns. Do this replacement early instead of in SPIRVProducerPass so
that -hack-scf can take effect if necessary.


alan-baker (commented):
Why not make the fixup pass handle the llvm intrinsics instead?

jrprice:
> Why not make the fixup pass handle the llvm intrinsics instead?

Done. Thought I was keeping things simple by handling the intrinsics in just a single place, but I was wrong.

alan-baker (approved):

PR #740, opened 2021-06-09
Support SPIR-V 1.4 and 1.5

alan-baker:
Major changes:
* SPIRVProducerPass is now unit testable
  * uses a simple constructor and outputs spir-v to a special file (specified via an option)
* SPIR-V 1.4 support
  * Layout decorations are only applied to storage buffer, uniform and push constant storage classes
    * uses OpCopyLogical to convert aggregates 
    * when loading, copying and storing only the memory has a layout and internal values have no layout
    * type mapping now tracks laid out and non-laid out versions of the type
  * Interfaces include all used global variables
    * resources are specific to the entry point
    * static globals are added to all entry points
  * storage buffer and variable pointer extensions no longer added to 1.3 and higher
  * select condition splatting does not happen for 1.4 or higher
  * Add ZeroExtend and SignExtend image operands for reads and writes
  * Memcpy uses both memory operands now
* Fixed a bug in memcpy replacement that erased some casts too early
* Fixed some type caching issues

### DIFF
--- a/include/clspv/Passes.h
+++ b/include/clspv/Passes.h
@@ -153,9 +153,10 @@ llvm::ModulePass *createReplaceOpenCLBuiltinPass();
 /// Create a pass to emit SPIR-V for the module.
 /// @return An LLVM module pass.
 llvm::ModulePass *createSPIRVProducerPass(
-    llvm::raw_pwrite_stream &out,
-    llvm::ArrayRef<std::pair<unsigned, std::string>> samplerMap,
+    llvm::raw_pwrite_stream *out,
+    llvm::SmallVectorImpl<std::pair<unsigned, std::string>> *samplerMap,
     bool outputCInitList);
+llvm::ModulePass *createSPIRVProducerPass();
 
 /// Undo LLVM's bitcast instructions with pointer type.
 /// @return An LLVM module pass.
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -3353,6 +3353,18 @@ SPIRVProducerPass::GenerateImageInstruction(CallInst *Call,
                                             const FunctionInfo &FuncInfo) {
   SPIRVID RID;
 
+  auto GetExtendMask = [this](Type *sample_type,
+                              bool is_int_image) -> uint32_t {
+    if (SpvVersion() >= SPIRVVersion::SPIRV_1_4 &&
+        sample_type->getScalarType()->isIntegerTy()) {
+      if (is_int_image)
+        return spv::ImageOperandsSignExtendMask;
+      else
+        return spv::ImageOperandsZeroExtendMask;
+    }
+    return 0;
+  };
+
   LLVMContext &Context = module->getContext();
   switch (FuncInfo.getType()) {
   case Builtins::kReadImagef:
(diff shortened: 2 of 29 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -727,7 +755,14 @@ bool SPIRVProducerPass::runOnModule(Module &M) {
   if (ShowProducerIR) {
     llvm::outs() << *module << "\n";
   }
-  binaryOut = outputCInitList ? &binaryTempOut : &out;
+
+  SmallVector<char, 10000> *binary = nullptr;
```
dneto0, line 759: 10000 kind of breaks the model of "Small".
But I'm glad it's not on the stack.

lib/SPIRVProducerPass.cpp
```diff
@@ -1471,19 +1530,60 @@ Type *SPIRVProducerPass::CanonicalType(Type *type) {
...
+  }
+  return getSPIRVType(Ty, needs_layout);
+}
+
+SPIRVID SPIRVProducerPass::getSPIRVType(Type *Ty, bool needs_layout) {
+  // Only pointers, structs and arrays should have layout decorations.
+  if (!(isa<PointerType>(Ty) || isa<ArrayType>(Ty) || isa<StructType>(Ty))) {
+    needs_layout = false;
+  }
+  const unsigned layout = needs_layout ? 1 : 0;
```
dneto0, line 1562: What's the meaning of layout.  

Oh, each type in the TypeMap can have up to two entries. 0 is the one without layout, and 1 is the one with layout.
Suggest a comment here saying something like this.  (I haven't reached the change to the TypeMap decl yet.)

lib/SPIRVProducerPass.cpp
```diff
@@ -1471,19 +1530,60 @@ Type *SPIRVProducerPass::CanonicalType(Type *type) {
...
 
   auto Canonical = CanonicalType(Ty);
   if (Canonical != Ty) {
     auto CanonicalTI = TypeMap.find(Canonical);
     if (CanonicalTI != TypeMap.end()) {
-      assert(CanonicalTI->second.isValid());
-      return CanonicalTI->second;
+      assert(layout < CanonicalTI->second.size());
+      if (CanonicalTI->second[layout].isValid()) {
+        // return CanonicalTI->second[layout];
```
dneto0, line 1578: nit: remove commented-out line.

dneto0 (approved):
LGTM with some commenting nits.
Thanks!

include/clspv/Passes.h
```diff
@@ -153,9 +153,10 @@ llvm::ModulePass *createReplaceOpenCLBuiltinPass();
 /// Create a pass to emit SPIR-V for the module.
 /// @return An LLVM module pass.
 llvm::ModulePass *createSPIRVProducerPass(
-    llvm::raw_pwrite_stream &out,
-    llvm::ArrayRef<std::pair<unsigned, std::string>> samplerMap,
+    llvm::raw_pwrite_stream *out,
+    llvm::SmallVectorImpl<std::pair<unsigned, std::string>> *samplerMap,
```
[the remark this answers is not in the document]
alan-baker: The Impl is an abstraction so you can use any SmallVector of that data and not be tied to a specific size. So the caller could call this with `SmallVector<x, 4>` or `SmallVector<x, 8>`.

PR #742, opened 2021-06-17
Update LLVM

sudonatalie:
* Replace enum compatibility synonyms that have now been removed


sudonatalie:
Ah, I see there is a place to update the LLVM hash, adding that now...

sudonatalie:
@alan-baker I had this about ready already, but feel free to close if you've got another one. Tested with `ninja check-spirv`.

alan-baker (approved):

PR #744, opened 2021-06-22
Allow function private storage pointers

polsenferreira:
The SPIRV producer has special cases for GEP handling, particularly when
the first index is 0.  However, when using pointer variables incremented
in a for-loop, the GEP operand becomes a PHINode, where the first offset
is not 0.  This causes the associated code in the SPIRV generator to
crash when it does not expect this access chain type.

Signed-off-by: Pedro Olsen Ferreira <pedro.olsenferreira@arm.com>

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -4147,6 +4147,7 @@ void SPIRVProducerPass::GenerateInstruction(Instruction &I) {
         getTypesNeedingArrayStride().insert(GEP->getPointerOperandType());
         break;
       case spv::StorageClassWorkgroup:
+      case spv::StorageClassFunction:
         break;
       default:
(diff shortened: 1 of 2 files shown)


alan-baker (changes requested):
Unfortunately this doesn't generate valid code (and the validator doesn't catch it). It generates `OpPtrAccessChain` and `OpPhi` on pointers in the `Function` storage class. Those are variable pointers, but variable pointers are restricted to `StorageBuffer` and `Workgroup` storage classes.

polsenferreira:
I am not well versed in SPIRV, so I can't really make sense of the reasons you stated (not doubting them).
At this point, I can't make progress as I don't know what should be happening for this case. Fortunately, the test case replicates the assertion, so I can open a bug using that.

PR #745, opened 2021-06-28
Remove the need for the type list

alan-baker:
* Since the refactor of the producer, the type list is no longer needed
  as dependencies are automatically generated correctly
  * remove all references to the type list
* update tests to be more agnostic to type declaration order


dneto0 (approved):
This is the best kind of change.
Nice cleanup!

Issue #761, opened 2021-09-07
stores involving GEP of aggregates of pointers hits invalid BitCast assert


thewilsonator:
[This line](https://github.com/google/clspv/blob/3970681ca8144e9a8d3cdd3f0d37c12465434211/lib/Compiler.cpp#L526) overwrites the target triple to "spir-unknown-unknown" and triggers the following two warnings 

```
warning: overriding the module target triple with spir-unknown-unknown
warning: Linking two modules of different data layouts: 'Unknown buffer' is 'e-p:3
2:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024' whereas 'kernel.ll' is 'e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024'
```
followed by the following assertion failure:
```
Assertion failed: (New->getType() == getType() && "replaceAllUses of value with new value of different type!"), function doRAUW, file /Users/nicholaswilson/d/llvm/clspv/clspv/third_party/llvm/llvm/lib/IR/Value.cpp, line 494.
```
for the attached file [kernel.ll.txt](https://github.com/google/clspv/files/7120530/kernel.ll.txt).

I don't know a whole lot about clang, but it seems to me that the correct triple should be recoverable from the `kernelFile` parameter of [`SetCompilerInstanceOptions`](https://github.com/google/clspv/blob/3970681ca8144e9a8d3cdd3f0d37c12465434211/lib/Compiler.cpp#L415), either that or add an extra command line argument.

I'll try to figure it out and have a PR soon.

thewilsonator:
Hmm, so apparently it doesn't like the bitcasts present. The warnings still make me nervous.

rjodinchr:
https://godbolt.org/z/9Ka6jzbrY
This is now passing.

I still needed to fix a few things in the kernel as shown in compiler-explorer:
- I needed to use the same data layout as the one in the builtin library to avoid the link warning
- _Z13get_global_idj needs to return an i32. Plus adding bitcast to i64 to keep everyone coherent.

thewilsonator:
Thanks!

PR #764, opened 2021-09-09
Fix issue #761 stores involving GEP of aggregates of pointers hits in…

thewilsonator:
…valid BitCast assert

Handle pointer scalar elements in `lib/ReplacePointerBitcastsPass.cpp:BuildFromElements` so that we don't generate an invalid bitcast from `float addrspace(1)*` to `i32` as  the `else` branch for regular scalar conversions conversion generates bitcasts to integer types.

### DIFF
--- a/lib/ReplacePointerBitcastPass.cpp
+++ b/lib/ReplacePointerBitcastPass.cpp
@@ -126,6 +126,11 @@ Value *BuildFromElements(Type *dst_type, const ArrayRef<Value *> &src_elements,
       auto *prev = dst ? dst : UndefValue::get(dst_type);
       dst = builder.CreateInsertElement(prev, tmp_value, i);
     }
+  } else if (auto *dst_ptr_ty = dyn_cast<PointerType>(dst_type)) {
+    auto *ele = src_elements[*index];
+    dst = builder.CreateBitCast(ele, dst_ptr_ty);
+    assert(dst_ptr_ty == ele->getType() && "pointer type mismatch");
+    ++(*index);
   } else {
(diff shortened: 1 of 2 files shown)


alan-baker:
Your test doesn't indicate what the expected code is. Bitcasting from a pointer to an integer is not generally allowed in shader SPIR-V and that appears to be what is happening.

When I checkout this branch I still get assertions that the replace all uses with is being used incorrectly.

thewilsonator:
>Bitcasting from a pointer to an integer is not generally allowed in shader SPIR-V and that appears to be what is happening.

Correct, but this is not present in the source input. The bad cast is generated by the ReplacePointerBitcastsPass because it (currently) doesn't handle pointers in aggregates properly and generates pointer-to-integer bit casts for (aggregate of pointer)-to-pointer bit casts (or the other way around, I don't remember off the top of my head). It should be noted that it does handle pointer-to-pointer bit casts outside of aggregates properly.

> Your test doesn't indicate what the expected code is.
> When I checkout this branch I still get assertions that the replace all uses with is being used incorrectly.

The test is a simple saxpy kernel, but the intent of the test is to validate that the compiler does not hit assert generating the code for it. Now that CI is enabled I will look into this more, because "It works on my machine".

Thanks!

thewilsonator:
>  "It works on my machine".

Seems I accidentally committed the 64-bit test case file, which won't work because compute SPIR-V is effectively 32 bit  (the width of `gl_InvocationID`. Fixed the style and unused variable warnings.

thewilsonator:
I've managed to work around this by not generating those bitcasts in the first place.

PR #765, opened 2021-09-23
Implement async_work_group_{strided_}copy

rjodinchr:
Replace OpenCL asynchronous copy by a loop doing OpCopyMemory
operations.
All work-items performs the copy in parallel.
Replace OpenCL wait_group_event by a OpControlBarrier to ensure the
synchronization between work-items.

Add test for the 3 functions

Update documentation about those functions (prefetch was already implemented and tested)


rjodinchr:
I have fix locally the format issue.
How do you proceed from there? Shall I push a new commit that will be squash later, or do I push force an updated commit?

alan-baker:
Feel free to push multiple commits. I will eventually squash them before merging.

alan-baker (commented):
Have you tested these against the OpenCL CTS?

rjodinchr:
Yes I did. It passes the OpenCL CTS for the types supported by clspv.

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -660,6 +675,191 @@ Type *ReplaceOpenCLBuiltinPass::GetPairStruct(Type *type) {
...
+   */
+
+  IRBuilder<> Builder(CI);
+
+  auto Cst0 = Builder.getInt32(0);
+  auto Cst1 = Builder.getInt32(1);
+  auto Cst2 = Builder.getInt32(2);
+
+  // get_local_id({0, 1, 2});
+  GlobalVariable *GVId = M.getGlobalVariable("__spirv_LocalInvocationId");
```
alan-baker, line 765: I think it's time to move "__spirv_LocalInvocationId" and "__spirv_WorkgroupSize" into lib/Constants.h and replace their uses with functions.

alan-baker: Separately, have you tested this against a kernel that doesn't already use `get_local_id()`? I'm not sure the local invocation id variable will be automatically generated in all cases.

test/AsyncWorkGroupCopy/wait_group_event.ll
```diff
@@ -0,0 +1,21 @@
...
+  %e = alloca %opencl.event_t*, align 4
+  store %opencl.event_t* null, %opencl.event_t** %e, align 4
+  call spir_func void @_Z17wait_group_eventsiP9ocl_event(i32 1, %opencl.event_t** %e)
+  ret void
+}
+
+declare spir_func void @_Z17wait_group_eventsiP9ocl_event(i32, %opencl.event_t**)
+
+; CHECK: [[tmp:%[a-zA-Z0-9_.]+]] = shl i32 1, 8
+; CHECK: call void @_Z8spirv.op.224.jjj(i32 224, i32 2, i32 2, i32 %0)
```
alan-baker, line 21: Should the barrier be using `[[tmp]]`? Also, as mentioned above this barrier doesn't include any memory orderings. Which implementations did you test against?

alan-baker (changes requested):

test/AsyncWorkGroupCopy/wait_group_event.ll
rjodinchr: I have tested this implementation using the OpenCL-CTS test suite. Maybe it did not stress enough the code to see the issue.

lib/ReplaceOpenCLBuiltinPass.cpp
rjodinchr: No, I did not test this against a kernel without a call to `get_local_id`. I will do that

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -660,6 +675,191 @@ Type *ReplaceOpenCLBuiltinPass::GetPairStruct(Type *type) {
...
+  /*
+   * event_t *async_work_group_strided_copy(T *dst, T *src, size_t num_gentypes,
+   *                                        size_t stride, event_t event) {
+   *   size_t start_id = ((get_local_id(2) * get_local_size(1))
+   *                     + get_local_id(1)) * get_local_size(0)
+   *                     + get_local_id(0);
+   *   size_t incr = get_local_size(0) * get_local_size(1) * get_local_size(2);
+   *   for (size_t it = start_id; it < num_gentypes; it += incr) {
+   *     dst[it] = src[it * stride];
+   *   }
```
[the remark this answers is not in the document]
rjodinchr: You are talking of this kind of implementation I suppose
```
    size_t nb_element_per_wi = num_gentypes / group_size;
    size_t remainder = num_gentypes % group_size; // OR num_gentypes - nb_element_per_wi * group_size
    size_t first_element = id * nb_element_per_wi + ((id < remainder) ? id : remainder);
    size_t last_element = first_element + nb_element_per_wi + ((id < remainder) ? 1 : 0);
    for (size_t i = first_element; i < last_element; i++) {
        dst[i] = src[i];
    }
```
I agree it could be more efficient. I will try it to have a look, but my first idea was to go for something simpler as this feature is mostly not used. Just enough to have the right behavior felt enough to me. I will try to implement this as well

lib/ReplaceOpenCLBuiltinPass.cpp
rjodinchr: I fix this with this commit 2e1889c
It adds tests as well for it

rjodinchr:
I think I would prefer to address it in another commit.
It would keep this one simpler.
Also, I'm not sure about the potential gain for it. I would need to really test it, but I would guess that this version is more cache-friendly for the GPU architecture where the access from consecutive work-items can be aggregate into 1 bigger access by the hardware.

Issue #768, opened 2021-09-24
dead pointer Bitcast remains for simple kernels


dneto0:
When compiling this kernel:

```C
kernel void foo(global float4* foo) {
  vstore_half4_rtz((float4)(0.0,0.0,0.0,0.0), 0, (global half*) foo);
}
```

There are two pointer bitcasts that are left behind by the ReplacePointerBitcasts pass:

         %28 = OpBitcast %_ptr_StorageBuffer_uint %21
         ...
         %34 = OpBitcast %_ptr_StorageBuffer_uint %33

They are unused, and should be swept away.

That pass is called in several places, and is often, but not always followed by LLVM's DeadCodeEliminationPass.

PR #769, opened 2021-09-24
Clean up dead code after simplify/replace pointer bitcasts

dneto0:
Fixes: #768


dneto0:
I'm not sure how you'd like me to test this.  Earlier you replaced end-to-end tests for half-storage with single-pass tests, that miss this issue.

alan-baker:
There are two possibilities:
1. Just run an end-to-end test
2. Use clspv-opt to run a couple of passes (simplify pointer bitcasts followed by dce)

Separately, it seems this change has broken the expectations on some other tests.

dneto0:
Ready for review. PTAL @alan-baker

alan-baker (approved):

PR #775, opened 2021-10-21
RFC: Rewrite private-AS GEP increments in loops

polsenferreira:
When a pointer in the private address space is incremented in a loop, the first offset in the GEP will not be a 0. This breaks the assumptions in the SPV writer pass, leading to an assertion.
Rewrite GEPs in loops to use an offset counter instead.

I have not done extensive testing on this pass, as I am not sure this is the right solution going forward. The pattern searched for in this pass is fragile, and this is effectively working-around a deficiency in the SPV spec (or the SPV writer), but I can't think of a different way to solve the problem.

A reduced test case is attached, along with a .ll file where the FileCheck checks describe what the pass is doing.

Signed-off-by: Pedro Olsen Ferreira <pedro.olsenferreira@arm.com>
Change-Id: I0abf1eeff2035101f01de0ccfddfbc663e92b998

### DIFF
--- a/include/clspv/Passes.h
+++ b/include/clspv/Passes.h
@@ -20,6 +20,7 @@ namespace llvm {
 // This is against Google C++ style guide.
 class FunctionPass;
 class ModulePass;
+class LoopPass;
 class raw_pwrite_stream;
 class raw_ostream;
 template <typename T> class ArrayRef;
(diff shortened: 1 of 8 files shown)


alan-baker:
I think generally this approach can work, but it feels like there will be some corner cases with multi-dimensional arrays. Not sure how best to proceed with those.

PR #778, opened 2021-10-25
Finish implementation of cl_khr_subgroups and OpenCL 3 core support

kpet:
Passes all the corresponding subgroup CTS tests.

New non-semantic instruction specified in https://github.com/KhronosGroup/SPIRV-Registry/pull/132
SPIR-V Headers changes in https://github.com/KhronosGroup/SPIRV-Headers/pull/250

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -22,7 +22,7 @@
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "branch" : "master",
       "subdir" : "third_party/SPIRV-Tools",
-      "commit" : "7403dfafd8a4b7216343dd9be891e540a5f30240"
+      "commit" : "339d4475c1a806c187c57678af26733575d1cecd"
     }
   ]
 }


alan-baker:
Seems like something isn't generating the expected enums properly.

kpet:
Yeah, it's working fine locally so I suspect it's a build parallelisation issue. I'll try to look into it during the weekend.

kpet:
Looking at the build logs, I _think_ there are two issues:
- SPIRV-Tools needs updating to a version that can handle the updated version of SPIRV-Headers. I've now done that. This should fix all the builds except those using amber.
- It seems amber builds are using the version of SPIRV-Headers amber wants rather than that in clspv's `deps.json`. Assuming this is the case I can see two ways forward: 1) Simply update the version of SPIRV-Headers in amber's dependencies 2) Change the jobs so they use the version of SPIRV-Headers in clspv's `deps.json`.

alan-baker:
If you get the SPIRV-Tools updated that should be enough. Amber should roll the deps at a reasonable frequency.

kpet:
I've already done that. I think you can kick off the bots again then.

kpet:
Thanks for merging this! I'll update SPIRV-Headers in clvk so that'll take care of the clvk failure and hopefully amber will update soon which will take care of the amber failure.

PR #781, opened 2021-11-09
Add pass to transform vec3 to vec4

### LINKED ISSUE #679: Problems with vec3 pointer casts
Consider the following kernel:
```
kernel void test(global out* out, global float3* in) {
  uint gid = get_global_id(0);
  float3 x = vload3(gid, in);
  out[gid] = x[0] + x[1] + x[2]; 
}
```

The replacement of `vload3` is reasonable:
```
...
  %3 = bitcast <3 x float> addrspace(1)* %2 to float addrspace(1)*
  %4 = mul i32 %1, 3
  %5 = add i32 %4, 0
  %6 = getelementptr float, float addrspace(1)* %3, i32 %5
  %7 = load float, float addrspace(1)* %6, align 4
  %8 = insertelement <3 x float> undef, float %7, i64 0
  %9 = add i32 %4, 1
  %10 = getelementptr float, float addrspace(1)* %3, i32 %9
  %11 = load float, float addrspace(1)* %10, align 4
  %12 = insertelement <3 x float> %8, float %11, i64 1
  %13 = add i32 %4, 2
  %14 = getelementptr float, float addrspace(1)* %3, i32 %13
  %15 = load float, float addrspace(1)* %14, align 4
  %16 = insertelement <3 x float> %12, float %15, i64 2
  store <3 x float> %16, <3 x float>* %x, align 16
  %17 = load <3 x float>, <3 x float>* %x, align 16
  %vecext = extractelement <3 x float> %17, i32 0
  %18 = load <3 x float>, <3 x float>* %x, align 16
  %vecext2 = extractelement <3 x float> %18, i32 1
  %add = fadd float %vecext, %vecext2
  %19 = load <3 x float>, <3 x float>* %x, align 16
  %vecext3 = extractelement <3 x float> %19, i32 2
  %add4 = fadd float %add, %vecext3
  %20 = load float addrspace(1)*, float addrspace(1)** %out.addr, align 4
  %21 = load i32, i32* %gid, align 4
  %arrayidx = getelementptr inbounds float, float addrspace(1)* %20, i32 %21
  store float %add4, float addrspace(1)* %arrayidx, align 4
```

The problem exists in replacing the bitcast:
```
  %1 = mul i32 %0, 3
  %2 = add i32 %1, 0
  %3 = lshr i32 %2, 1
  %4 = getelementptr <3 x float>, <3 x float> addrspace(1)* %in, i32 %3
  %src_val7 = load <3 x float>, <3 x float> addrspace(1)* %4, align 16
  %5 = and i32 %2, 2
  %tmp_val8 = extractelement <3 x float> %src_val7, i32 %5
  %6 = insertelement <3 x float> undef, float %tmp_val8, i64 0
  %7 = add i32 %1, 1
  %8 = lshr i32 %7, 1
  %9 = getelementptr <3 x float>, <3 x float> addrspace(1)* %in, i32 %8
  %src_val5 = load <3 x float>, <3 x float> addrspace(1)* %9, align 16
  %10 = and i32 %7, 2
  %tmp_val6 = extractelement <3 x float> %src_val5, i32 %10
  %11 = insertelement <3 x float> %6, float %tmp_val6, i64 1
  %12 = add i32 %1, 2
  %13 = lshr i32 %12, 1
  %14 = getelementptr <3 x float>, <3 x float> addrspace(1)* %in, i32 %13
  %src_val = load <3 x float>, <3 x float> addrspace(1)* %14, align 16
  %15 = and i32 %12, 2
  %tmp_val = extractelement <3 x float> %src_val, i32 %15
  %16 = insertelement <3 x float> %11, float %tmp_val, i64 2
  %vecext = extractelement <3 x float> %16, i32 0
  %vecext2 = extractelement <3 x float> %16, i32 1
  %add = fadd float %vecext, %vecext2
  %vecext3 = extractelement <3 x float> %16, i32 2
  %add4 = fadd float %add, %vecext3
  %arrayidx = getelementptr inbounds float, float addrspace(1)* %out, i32 %0
  store float %add4, float addrspace(1)* %arrayidx, align 4
```

Clspv eventually generates a runtime array of `float3` with an array stride of 16.

The index adjustments to the loads are incorrect as they are relative to a float2 (shift right by 1). The element extraction is also wrong as it can only extract elements 0 and 2. In fact, representing the kernel argument as a float3 with 16 stride makes the bytes [12, 15] inaccessible, so vload3(1, ...) cannot work. We cannot use an array stride of 12 as that is invalid without scalar layout support (and even then replace pointer bitcasts would need to change).

It seems clspv should instead represent the kernel argument as either a runtime array of float or float4. Correct code could be generated in such a case. Float4 is probably preferable since it is closer to float3 already.

rjodinchr:
This is done only if a bitcast of a vec3 is detected.
It can also be forced using the '-vec3-to-vec4' option.

exceptions:
 - internal spirv global variables are often vec3 corresponding to the
 3 dimensions. Those are not transformed.
 - builtins: some of them are safe to use with vec4 instead of vec3,
 but not all of them (OpCopyMemory, OpAny, OpAll, etc)

No regression on the OpenCL CTS + 27 new tests passed

Fixes #679

### DIFF
--- a/lib/ThreeElementVectorLoweringPass.cpp
+++ b/lib/ThreeElementVectorLoweringPass.cpp
@@ -1,0 +1,1419 @@
+// Copyright 2020-2021 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/InstVisitor.h"
+#include "llvm/IR/Intrinsics.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/ValueHandle.h"
+#include "llvm/Pass.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Transforms/Utils/Cloning.h"
+#include "llvm/Transforms/Utils/Local.h"
+
+#include "BuiltinsEnum.h"
+#include "Constants.h"
+#include "clspv/Option.h"
+#include "clspv/Passes.h"
+
+#include "Builtins.h"
+#include "Passes.h"
+
+#include <array>
+#include <functional>
+
+using namespace llvm;
+
+#define DEBUG_TYPE "ThreeElementVectorLowering"
+
+namespace {
+
+class ThreeElementVectorLoweringPass final
+    : public ModulePass,
+      public InstVisitor<ThreeElementVectorLoweringPass, Value *> {
+public:
+  static char ID;
+
+public:
+  ThreeElementVectorLoweringPass() : ModulePass(ID) {}
+
+  /// Lower the content of the given module @p M.
+  bool runOnModule(Module &M) override;
+
+private:
+  // Implementation details for InstVisitor.
+
+  using Visitor = InstVisitor<ThreeElementVectorLoweringPass, Value *>;
+  using Visitor::visit;
+  friend Visitor;
+
+  /// Higher-level dispatcher. This is not provided by InstVisitor.
+  /// Returns nullptr if no lowering is required.
+  Value *visit(Value *V);
+
+  /// Visit Constant. This is not provided by InstVisitor.
+  Value *visitConstant(Constant &Cst);
+
+  /// Visit Unary or Binary Operator. This is not provided by InstVisitor.
+  Value *visitNAryOperator(Instruction &I);
+
+  /// InstVisitor impl, general "catch-all" function.
+  Value *visitInstruction(Instruction &I);
+
+  // InstVisitor impl, specific cases.
+  Value *visitAllocaInst(AllocaInst &I);
+  Value *visitBinaryOperator(BinaryOperator &I);
+  Value *visitCallInst(CallInst &I);
+  Value *visitCastInst(CastInst &I);
+  Value *visitCmpInst(CmpInst &I);
+  Value *visitExtractElementInst(ExtractElementInst &I);
+  Value *visitExtractValueInst(ExtractValueInst &I);
+  Value *visitGetElementPtrInst(GetElementPtrInst &I);
+  Value *visitInsertElementInst(InsertElementInst &I);
+  Value *visitInsertValueInst(InsertValueInst &I);
+  Value *visitLoadInst(LoadInst &I);
+  Value *visitPHINode(PHINode &I);
+  Value *visitSelectInst(SelectInst &I);
+  Value *visitShuffleVectorInst(ShuffleVectorInst &I);
+  Value *visitStoreInst(StoreInst &I);
+  Value *visitUnaryOperator(UnaryOperator &I);
+
+private:
+  // Helpers for lowering values.
+
+  /// Return true if the given @p U needs to be lowered.
+  ///
+  /// This only looks at the types involved, not the opcodes or anything else.
+  bool handlingRequired(User &U);
+
+  /// Return the lowered version of @p U or @p U itself when no lowering is
+  /// required.
+  Value *visitOrSelf(Value *U) {
+    auto *V = visit(U);
+    return V ? V : U;
+  }
+
+  /// Register the replacement of @p U with @p V.
+  ///
+  /// If @p U and @p V have the same type, replace the relevant usages as well
+  /// to ensure the rest of the program is using the new instructions.
+  void registerReplacement(Value &U, Value &V);
+
+private:
+  // Helpers for lowering types.
+
+  /// Get a vector of 4 elements equivalent for this type, if it uses a vector
+  /// of 3 elements. Returns nullptr if no lowering is required.
+  Type *getEquivalentType(Type *Ty);
+
+  /// Implementation details of getEquivalentType.
+  Type *getEquivalentTypeImpl(Type *Ty);
+
+  /// Return the equivalent type for @p Ty or @p Ty if no lowering is needed.
+  Type *getEquivalentTypeOrSelf(Type *Ty) {
+    auto *EquivalentTy = getEquivalentType(Ty);
+    return EquivalentTy ? EquivalentTy : Ty;
+  }
+
+private:
+  // High-level implementation details of runOnModule.
+
+  /// Lower all global variables in the module.
+  bool runOnGlobals(Module &M);
+
+  /// Lower the given function.
+  bool runOnFunction(Function &F);
+
+  /// Map the call @p CI to an OpenCL builtin function or an LLVM intrinsic to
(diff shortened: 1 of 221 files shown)


kpet:
@rjodinchr I don't know if you'd spotted this but Clang can codegen loads/stores from/to 3-element vectors as loads/stores from/to 4-element vectors (disabled at https://github.com/google/clspv/blob/master/lib/Compiler.cpp#L550). It's not as simple as changing the option (dealing with bitcasts between vec3 and vec4 pointers would also be required) but I'm wondering whether this could be a possible alternative to this pass.

rjodinchr:
@kpet thank you, I didn't know about that.
I'm not sure that would solve the issue.
As we are not able to bitcast pointer in spirv, if a argument of the compute shader is declared as a pointer on a vec3, the padding element will not be accessible whatever you do?
What do you think?

kpet:
I'm not sure either. I'd have to dive into it to form an opinion. I just thought I'd highlight this as a potential alternative approach. Maybe @alan-baker or @jrprice have an opinion?

alan-baker:
I'm reviewing this, but a higher level question: why does it completely change vec3 to vec4 (expect builtins)? My first expectation would have been a much tighter target around pointer casts and/or vload3-like behaviour. I'm not sure there's a performance disadvantage, but it ends up being a lot more code than I expected. Would making this pass more focused on the bitcasts and special cases lower the amount of code necessary. It might make sense at that point to use the option @kpet pointed out.

If we do go this route, the code is looking pretty good overall, but there is a lot to review.

rjodinchr:
My understanding is that pointer casting is not possible in SPIRV. Thus if a vec3 compute shader argument is casted into something else at some point cannot stay a vec3, otherwise we would not be able to access the padding element that some programs want to use.
So modifying the prototype for those arguments look like a must for me, but I might be missing something here.
After that we could try to change less thing from vec3 to vec4. But as a lot could happen between to beginning of the kernel and the place where it really needs to be in vec4, it feels that all the code of this PR could be needed to arrive at this point. Trying to change less things would just required more code to be written, and thus be more error prone.
We could also try to modify even less by adding bitcast from vec4 to vec3 in the llvm IR, which could be manageable, but those will end-up to be loading the vector in vec4 to then reconstruct the vec3, which would add a overhead.
All of this relies on the assumption that we need to change vec3 to vec4 in the kernel prototype. From my understanding it is a must, but maybe there is another way I don't get right now.

Another point is that this PR comes from the fact that the CTS is doing something quite unexpected with vec3 I think.
The CTS is keeping the same dataset for every vector types. So for vec3, the padding element also needs to be compute.
At the moment the CTS do it use this pattern:
```
__kernel void test(__global type3 *arg, ... ) {
   int gid = get_global_id(0);
   type3 tmp = vload3(gid, (__global type *)arg);
   ...
}
```
The part that is really testing is the one after the vload3, so changing to the following pattern would test the same thing while not generating this issue for us:
```
__kernel void test(__global type *arg, ... ) {
   int gid = get_global_id(0);
   type3 tmp = vload3(gid, arg);
   ...
}
```
It would mean that try to access the padding element of a vec3 using a cast is not supported in OpenCL-C, but it is not the case.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -0,0 +1,1419 @@
...
+      if (EquivalentTy != nullptr) {
+        Types.push_back(EquivalentTy);
+        RequiredLowering = true;
+      } else {
+        Types.push_back(CTy);
+      }
+    }
+
+    if (RequiredLowering) {
+      return StructType::get(Ctx, Types, StructTy->isPacked());
```
alan-baker, line 1032: If the struct was packed, lowering from vec3 to vec4 might change the layout. I think the following can be packed into 16 bytes:

```
struct S {
  float3 x;
  int y;
};
```

If you have:
```
struct S {
  float4 x;
  int y;
};
```
The offset for y will be at 16 bytes making the total size 20 bytes.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -0,0 +1,1419 @@
...
+    if (Arg->getType()->isVectorTy()) {
+      Value *NewArg = UndefValue::get(
+          VectorType::get(Arg->getType()->getScalarType(), 3, false));
+      Value *tmp = B.CreateExtractElement(Arg, (uint64_t)0);
+      NewArg = B.CreateInsertElement(NewArg, tmp, (uint64_t)0);
+      tmp = B.CreateExtractElement(Arg, (uint64_t)1);
+      NewArg = B.CreateInsertElement(NewArg, tmp, (uint64_t)1);
+      tmp = B.CreateExtractElement(Arg, (uint64_t)2);
+      NewArg = B.CreateInsertElement(NewArg, tmp, (uint64_t)2);
+      Args.push_back(NewArg);
```
alan-baker, line 1163: This seems like it should be a shuffle vector.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -0,0 +1,1419 @@
...
+  }
+
+  CallInst *NewVectorCall = B.CreateCall(VectorFunction, Args);
+  NewVectorCall->copyIRFlags(&VectorCall);
+  NewVectorCall->copyMetadata(VectorCall);
+  NewVectorCall->setCallingConv(VectorCall.getCallingConv());
+
+  Type *RetTy = VectorFunction->getReturnType();
+
+  if (RetTy->isVectorTy()) {
```
alan-baker, line 1176: Same here

test/issue-679.cl
```diff
@@ -0,0 +1,40 @@
+// RUN: clspv %s -o %t2.spv
```
alan-baker, line 1: I'd prefer to have a version of this that is more targeted at the pass. Maybe just run the lowering pass and the function internalizer

alan-baker (changes requested):
The code looks pretty good overall.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -0,0 +1,1419 @@
...
+      NewValue = B.CreateInsertValue(NewValue, NewElement, {i});
+    }
+  } else if (EquivalentTy->isVectorTy()) {
+    assert(Ty->isVectorTy());
+
+    unsigned Arity = dyn_cast<FixedVectorType>(Ty)->getNumElements();
+    for (unsigned i = 0; i < Arity; ++i) {
+      Value *Scalar = B.CreateExtractElement(V, i);
+      NewValue = B.CreateInsertElement(NewValue, Scalar, i);
+    }
```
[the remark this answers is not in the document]
rjodinchr: Indeed.
I modified the code to use shuffle vector, but the generated spirv at the end is not the same.
The `instcombine` pass is capable of getting ride of those `extractelement`/`insertelement` when they are not really useful, while it keeps the `shufflevector`.

lib/ThreeElementVectorLoweringPass.cpp
rjodinchr: I agree.
It is not an easy problem.
I think a proper solution would be to say that float3 inside a packed structure is just not supported by clspv at the moment, and have a llvm_unreachable for that case. What do you think?
I am pushing another "solution" which do not lower packed structure. But I am not very confident about this implementation as it changes quite a lot in the way the pass works, and I did not test it enough.

lib/ThreeElementVectorLoweringPass.cpp
rjodinchr: Done

lib/ThreeElementVectorLoweringPass.cpp
rjodinchr: Done

test/issue-679.cl
rjodinchr: I have added another test that target only the pass.

alan-baker (approved):

PR #783, opened 2021-11-19
longvector: Uses Array instead of Struct

### LINKED ISSUE #780: CTS: basic: vstore_local: char8: issue with GEP and long vector
Some tests of the CTS (`vstore_local`, `vload_local` and others) fail when trying with long vector type (ie. more than 4 elements).

In _Release_ mode, it _segfault_ in `Type.h`, coming from `SimplifyPointerBitcastPass.cpp:295`
In _Debug_, it hits an assert in `Instructions.h` also coming from `SimplifyPointerBitcastPass.cpp:295`

At that point clspv is trying to merge two GEPs (one being the input of the other).
The initial pointer is a pointer on a long vector, which has been lower to a pointer on a structure.
The line in `SimplifyPointerBitcastPass` is trying to create the new GEP which will be the merge of the two GEPs.

In this case, the GEP, is trying to point on a element inside the structure.

When llvm is making this GEP, it needs to know the return type of this GEP. The pointer being a structure, it tries to know which element of the structure is targeted, but the Idx is dynamic in this case, thus, it is impossible to know which element of the structure is targeted. At the end, llvm consider the index as invalid for the structure, and return a null pointer as the return type of the GEP, which produces the errors mentioned above.

rjodinchr:
To support long vectors, clspv use to transform them into struct of
the element type of the vector. But it is not ideal because llvm does
not know that all the elements of the structure are the same.
Let's use array instead of struct to help llvm.

No regression on OpenCL CTS + vload_local is now passing

Fixes #780

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -113,7 +113,7 @@ private:
 private:
   // Helpers for lowering types.
 
-  /// Get a struct equivalent for this type, if it uses a long vector.
+  /// Get a array equivalent for this type, if it uses a long vector.
   /// Returns nullptr if no lowering is required.
   Type *getEquivalentType(Type *Ty);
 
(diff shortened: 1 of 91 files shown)


alan-baker:
@rjodinchr please rebase this PR.

alan-baker:
Builds are failing due to a SPIRV-Headers issue. I think this needs a rebase.

alan-baker (approved):

PR #791, opened 2022-01-05
Add support for image1d buffer

rjodinchr:
This patch is the complement of ongoing work on clvk side.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -2559,7 +2559,7 @@ bool ReplaceOpenCLBuiltinPass::replaceSampledReadImageWithIntCoords(
     // The coordinate (integer type that we can't handle).
     auto Arg2 = CI->getOperand(2);
 
-    uint32_t dim = clspv::ImageDimensionality(Arg0->getType());
+    uint32_t dim = clspv::ImageNumDimensions(Arg0->getType());
     uint32_t components =
         dim + (clspv::IsArrayImageType(Arg0->getType()) ? 1 : 0);
     Type *float_ty = nullptr;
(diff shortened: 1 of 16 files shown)


alan-baker (changes requested):
I think this should not just map to a 1D image, but should instead map to an OpTypeImage with dimensionality of Buffer. That likely also means needing a different descriptor type so that the API can set it as either uniform texel buffer or storage texel buffer.

rjodinchr:
I try to do it, but I end up with something that I don't understand yet.
A kernel using 1D image buffer generates a SPIRV code containing:
```
OpCapability SampledBuffer
%18 = OpTypeImage %int Buffer 0 0 0 1 Unknown
%37 = OpLoad %18 %21
%48 = OpImageFetch %v4int %37 %47 Lod|SignExtend %uint_0
```
and `spirv-val` returns an error:
```
error: line 95: Image Operand Lod requires 'Dim' parameter to be 1D, 2D, 3D or Cube
  %48 = OpImageFetch %v4int %37 %47 Lod|SignExtend %uint_0
```
While the [specification](https://www.khronos.org/registry/SPIR-V/specs/1.2/SPIRV.html#OpImageFetch) only mentions that:
```
Its Dim operand cannot be Cube, and its Sampled operand must be 1.
```
Could you help me on that point?

alan-baker:
I think you may have to translate it as a storage image (i.e. Sampled=2).

rjodinchr:
I am not sure how to translate it as a storage image.
I tried a few things, and always got:
```
error: line 94: Expected Image 'Sampled' parameter to be 1
  %47 = OpImageFetch %v4int %36 %46 Lod|SignExtend %uint_0
```

alan-baker:
Sorry, for not explaining it further. If you generate a storage image, you'd then also generate OpImageRead instead of OpImageFetch.

alan-baker (approved):

Issue #795, opened 2022-01-13
crash on a specific input


FROL256:
Hello. I have updated clspv today and got crash on my project. I tried to isolate bug, but it actually reproduces  only under certain circumstances depend on different places of a code. So, please see archive where i did my best to isolate it )

https://drive.google.com/file/d/13Rw_2NVY5nGfccz_74Q0S0XS9zX-nHEH/view?usp=sharing

Lines 3, 56, 57 and 72 -- al of them influence on crash.

rjodinchr:
I had a look at it.
It is due to the pass transforming vec3 into vec4, where the PHI Nodes are not supported yet.
2 possibilities from there:
- implement PHI nodes support in this pass
- add an option to explicitly disable this pass, which might not be needed in your case.
I will try to create PR for both before the end of the week so that we can at least cherry-pick one of them.

rjodinchr:
I think we can now close this issue

FROL256:
yep, it works for me, thank you very much!

PR #796, opened 2022-01-14
vec3: implement phi nodes support

rjodinchr:
Need more test, but should be enough to fix issue from #795


alan-baker (changes requested):
Needs some tests. Especially some that cover loops formed with the phi node.

alan-baker (approved):

alan-baker:
Looks like the MSVC builds are having some problems with the use of PHINode::removeIncomingValue().

PR #797, opened 2022-01-14
vec3: add option to force disabling of the vec3 pass

rjodinchr:
no-vec3-to-vec4 takes priority over vec3-to-vec4 as it intends to
avoid running into bugs in the vec3 pass implementation.

Ref #795

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -644,13 +644,6 @@ int PopulatePassManager(
     break;
   }
 
-  if (clspv::Option::Vec3ToVec4() ==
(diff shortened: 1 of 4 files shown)


alan-baker (changes requested):
Seems like there should be an error if both options are specified.

lib/Compiler.cpp
```diff
@@ -644,6 +644,13 @@ int PopulatePassManager(
     break;
   }
 
+  if (clspv::Option::Vec3ToVec4() ==
```
alan-baker, line 647: This should be in `ParseOptions()` with the other option error handling.

alan-baker (approved):

PR #804, opened 2022-01-19
OpSClamp: workaround Nvidia GPU driver issue

rjodinchr:
OpSClamp on Nvidia GPU does not behave as expected with types smaller
than 32bit.
To avoid that when doing mad_sat/add_sat/sub_sat, let's extend the
type to 32bit to make sure OpSClamp will behave properly. It does not
change the number of instruction needed.


alan-baker (changes requested):
Generally our approach to these sorts of bugs has been to add a -hack option (e.g. -hack-initializers). I'd suggest doing the same here (e.g. -hack-clamp-width).

alan-baker (approved):

PR #808, opened 2022-01-20
locking clspv invocations

rjodinchr:
llvm and clspv are not thread safe, let's lock the clspv API
invocations to make sure that a multi-threaded program can use it.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -939,7 +939,11 @@ bool LinkBuiltinLibrary(llvm::Module *module) {
 } // namespace
 
 namespace clspv {
+
+static std::mutex g_compiler_lock;
+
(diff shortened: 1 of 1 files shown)


alan-baker:
I can understand wanting to do this, but I'm on the fence about whether or not this is clspv's responsibility or not. I really wish LLVM's options weren't global variables.

rjodinchr:
I agree.
I can also do that in `clvk`.
My thinking was that, if anyone else was using `clspv` in a multi-threaded way, they will potentially have the same issue, so it seems better to be there.

rjodinchr:
@alan-baker How do you want to proceed with this PR?

alan-baker:
I think I'd rather not add this here. To me, it seems more appropriate in clvk.

PR #810, opened 2022-01-28
Update LLVM

dneto0:
Also, strip device-enque features from the header file.
A recent change to Clang accidentally removes the predeclaration
of the queue_t and clk_event_t types for OpenCL 2.0 flows.
A compile error results in opencl-c.h if we don't strip the offending
declarations of the enqueue-related builtins.

### DIFF
--- a/cmake/strip_banned_opencl_features.py
+++ b/cmake/strip_banned_opencl_features.py
@@ -30,7 +30,13 @@ def main():
     args = parser.parse_args()
 
     # Strip invalid features.
-    regex = re.compile('(convert_[a-zA-Z0-9]+(_rt[pn]|_sat))|(reserve_id_t)')
+    unsupported_features = '(convert_[a-zA-Z0-9]+(_rt[pn]|_sat))'
+    unsupported_features += '|(reserve_id_t)'
+    unsupported_features += '|(ndrange_t)'
+    unsupported_features += '|(queue_t)'
+    unsupported_features += '|(clk_event_t)'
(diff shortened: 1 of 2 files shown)


dneto0:
This is the offending upstream change:

https://github.com/llvm/llvm-project/commit/a5de66c4c50b5dec085ffbcd2afefd47f3f18d7f#diff-c538f45bf941ec180ae164adeaa4d749520101359503af9373076feecfb88228R337

The overall change adds OpenCL 3.0 support for feature macro __opencl_c_device_enqueue.
That's fine. But in Sema, it accidentally turns off predeclaration of queue_t and clk_event_t.
But the opencl-c.h header depends on those definitions existing.

So compilation fails with -cl-std=CL2.0  because the __opencl_c_device_enqueue is enabled by opencl-c-base.h but then fails at the first declaration that uses queue_t or clk_event_t.

dneto0:
As luck would have it, I can't reproduce the problem independenltly on upstream clang.

I expected a trivial source file would fail to compile with these options:

    clang -cc1 -cl-std=CL2.0 foo.cl -emit-llvm -o -

But I get no problem.
On the other hand, I'm also unable to compile the 'ctz' function, which was introduced in OpenCL 2.0, I think.

```
$ ./bin/clang -cc1 -cl-std=CL2.0 a.cl
a.cl:2:8: error: implicit declaration of function 'ctz' is invalid in OpenCL
  *A = ctz(*A);
       ^
1 error generated.
```

alan-baker (approved):

PR #814, opened 2022-02-04
popcount: manage non 32-bit types explicitly

rjodinchr:
The Base operand of any OpBitCount, OpBitReverse, OpBitFieldInsert,
OpBitFieldSExtract, or OpBitFieldUExtract instruction must be a 32-bit
integer scalar or a vector of 32-bit integers.
https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/chap46.html#VUID-StandaloneSpirv-Base-04781

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -3832,55 +3832,38 @@ SPIRVID SPIRVProducerPass::GenerateInstructionFromCall(CallInst *Call) {
     auto TyBitWidth = Ty->getScalarSizeInBits();
 
     SPIRVOperandVec Ops;
-    if (TyBitWidth > BitcountSize) {
-      SPIRVID Value;
-      bool bAdd = false;
-      auto Cst32 = ConstantInt::get(Ty, BitcountSize);
+    if (TyBitWidth == BitcountSize * 2) {
+      Ops.clear();
+      Ops << Int32Ty << BaseValue;
+      auto convertLower = addSPIRVInst(spv::OpUConvert, Ops);
 
-      assert(TyBitWidth % BitcountSize == 0);
+      Ops.clear();
+      Ops << Int32Ty << convertLower;
+      auto bitcountLower = addSPIRVInst(spv::OpBitCount, Ops);
 
-      while (TyBitWidth != 0) {
-        Ops.clear();
-        Ops << Int32Ty;
-        if (bAdd)
-          Ops << Value;
-        else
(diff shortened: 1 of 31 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3818,15 +3818,80 @@ SPIRVID SPIRVProducerPass::GenerateInstructionFromCall(CallInst *Call) {
...
+    auto BaseValue = Call->getOperand(0);
+    Type *Int32Ty = Ty->getInt32Ty(Context);
+    if (Ty->isVectorTy())
+      Int32Ty = VectorType::get(Int32Ty, dyn_cast<VectorType>(Ty));
+    auto TyBitWidth = Ty->getScalarSizeInBits();
+
     SPIRVOperandVec Ops;
-    Ops << Call->getType() << Call->getOperand(0);
+    if (TyBitWidth > BitcountSize) {
+      SPIRVID Value;
```
alan-baker, line 3836:
```suggestion
      SPIRVID Value = getSPIRVValue(BaseValue);
```

Then you can drop the ifs to decide between using Value and BaseValue.

lib/SPIRVProducerPass.cpp
```diff
@@ -3818,15 +3818,80 @@ SPIRVID SPIRVProducerPass::GenerateInstructionFromCall(CallInst *Call) {
...
+        else
+          Ops << BaseValue;
+        auto convert = addSPIRVInst(spv::OpUConvert, Ops);
+
+        Ops.clear();
+        Ops << Int32Ty << convert;
+        auto bitcount = addSPIRVInst(spv::OpBitCount, Ops);
 
-    RID = addSPIRVInst(spv::OpBitCount, Ops);
+        if (TyBitWidth != BitcountSize) {
```
alan-baker, line 3855: Nit: it feels like you're mixing multiple control styles. This could be moved first and based on bAdd.

alan-baker (changes requested):

lib/SPIRVProducerPass.cpp
rjodinchr: Yes, I don't know what I was thinking, but I was kind of considering types bigger than 64-bit here, in which case it makes sense to have `bAdd` to differentiate the first loop, and this condition for the last loop.
I will rework it as types above 64-bit does not exist in OpenCL-C.

alan-baker (approved):

PR #818, opened 2022-02-16
SRem: manage signed modulus explicitly

rjodinchr:
For the OpSRem and OpSMod instructions, if either operand is negative
the result is undefined.
https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/chap46.html#spirvenv-precision-operation


alan-baker:
Did you run the conformance tests on this change? I assume they passed.

rjodinchr:
Yes I ran the CTS. With it all integer_ops tests pass (except mad_sat because of a Nvidia bug on OpIAddCarry). Also some tests takes quite a long time to run (~20min). I have a PR in the CTS repo to allow to reduce the dataset size to be able to run them faster,

alan-baker (approved):

PR #823, opened 2022-02-24
Allow for all options in CompileFromSourceString

RSenApps:
Shares core logic between Compile and CompileFromSourceString to
allow for some previously unsupported options in
CompileFromSourceString such as support for input languages
other than OpenCL and outputting to a file.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -1053,7 +1053,7 @@ int Compile(const int argc, const char *const argv[]) {
   if (auto error = ParseOptions(argc, argv))
     return error;
 
-  // if no output file was provided, use a default
+  // if no input file was provided, use a default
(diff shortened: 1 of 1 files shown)


lib/Compiler.cpp
```diff
@@ -1036,30 +1025,63 @@ int Compile(const int argc, const char *const argv[]) {
...
+  }
+
+  return 0;
+}
+
+int Compile(const int argc, const char *const argv[]) {
+  if (auto error = ParseOptions(argc, argv))
+    return error;
+
+  // if no output file was provided, use a default
```
alan-baker, line 1056:
```suggestion
  // if no input file was provided, use a default
```

alan-baker (changes requested):
Looks good overall. Small comment typo only.

alan-baker (approved):

PR #825, opened 2022-02-27
Use Clang's TableGen-based fast builtin function declarations

kpet:
On my machine that gives about a 2x speedup on either clspv's test suite
or a wimpy CTS run.

See https://clang.llvm.org/docs/OpenCLSupport.html#opencl-builtins for
background.


alan-baker:
What kind of errors happen now if you use unsupported features like pipes? Previously those would lead to reasonable missing definition errors. Will it just segfault now?

Also looks, like the Amber tests need updated for rhadd and hadd, but also I'm not sure the behaviour is desirable. Why does the user declaration completely shadow the opencl builtins?

kpet:
>What kind of errors happen now if you use unsupported features like pipes? Previously those would lead to reasonable missing definition errors. Will it just segfault now?

For pipes, clspv now just crashes when building with -cl-std=CL2.0, as for many other unsupported features, but cleanly errors out when building with -cl-std=CL3.0. Given the general state of error handling, I'm not sure it's too concerning TBH but there may be cases that are more problematic.

>Also looks, like the Amber tests need updated for rhadd and hadd, but also I'm not sure the behaviour is desirable. Why does the user declaration completely shadow the opencl builtins?

I've updated the amber tests. This didn't make me pause when I updated the failing bitselect test as I assumed that OpenCL C "naturally" reserved all built-in function names but I can't find any such statement in the specification. OpenCL C doesn't really allow overloading functions but this is a change of behaviour and the error message is not that helpful so all in all this is probably something we want to improve in Clang.

alan-baker:
Looking at the link you provided, it seems like the builtin declaration aren't actually part of the AST. So, with regards to errors, we could modify the frontend action to issue errors for unsupported data types. I think that would catch almost all the cases that should be rejected. That would certainly be better than a crash.

I'm still contemplating shadowing the names. At least we should have a note in OpenCLOnVulkan.md warning people not to create their own overrides of builtin functions.

kpet:
>we could modify the frontend action to issue errors for unsupported data types. I think that would catch almost all the cases that should be rejected. That would certainly be better than a crash.

That would certainly be an improvement or rather not a regression on the current behaviour. I'll have a look during the weekend.

>At least we should have a note in OpenCLOnVulkan.md warning people not to create their own overrides of builtin functions.

We could do but I don't think it is legal to do in OpenCL C anyway. Admittedly this PR changes the behaviour and (illegal) code that was previously working would no longer compile and with a not-so-obvious error message. I'll follow up with the orignal author of the change in Clang to see what we could do there as well.

kpet:
@alan-baker I've added tests for the features that were previously stripped out from the header, namely DSE and pipes. The only case that was crashing and required additional diagnostics in the frontend action was pipe types in programs compiled with `-cl-std=CL2.0`, all other cases I've tested produced error messages. I didn't add tests for the rtp/rtp/sat variants of `convert` because these are currently producing conversions, ignoring the rounding mode or request for saturation. PTAL.

alan-baker (approved):

PR #828, opened 2022-03-31
OpMulExtended: workaround Intel GPU issue

rjodinchr:
OpMulExtended only works on Intel GPU with 32bit types.
Workaround it with a new option 'hack-mul-extended' which prevents
clspv from using OpMulExtended.

Also change the SRem implementation not to use OpMulExtended all the
time.

### DIFF
--- a/lib/ReplaceOpenCLBuiltinPass.cpp
+++ b/lib/ReplaceOpenCLBuiltinPass.cpp
@@ -311,6 +311,9 @@ private:
   // identical contents.
   Type *GetPairStruct(Type *type);
 
+  Value *InsertOpMulExtended(Instruction *InsertPoint, Value *a, Value *b,
+                                   bool IsSigned);
+
   DenseMap<Type *, Type *> PairStructMap;
 };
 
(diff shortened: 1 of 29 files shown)


lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -678,6 +681,161 @@ Type *ReplaceOpenCLBuiltinPass::GetPairStruct(Type *type) {
...
+    return clspv::InsertSPIRVOp(InsertPoint, opcode, {Attribute::ReadNone},
+                                RetTy, {a, b});
+  }
+
+  unsigned int ScalarSizeInBits = Ty->getScalarSizeInBits();
+  bool IsVector = Ty->isVectorTy();
+
+  IRBuilder<> Builder(InsertPoint);
+
+  if (ScalarSizeInBits <= 32) {
```
alan-baker, line 704: Doing this for 32-bit integers means adding the Int64 capability requirement. That seems unfortunate for a workaround.

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -678,6 +681,161 @@ Type *ReplaceOpenCLBuiltinPass::GetPairStruct(Type *type) {
...
+        Builder.CreateAdd(Builder.CreateAdd(a0b0_1, a1b0_0), a0b1_0);
+    auto carry_mul_lo_hi = Builder.CreateLShr(mul_lo_hi, 32);
+    auto mul_hi_lo = Builder.CreateAdd(
+        Builder.CreateAdd(Builder.CreateAdd(a1b1_0, a1b0_1), a0b1_1),
+        carry_mul_lo_hi);
+    auto mul_lo = Builder.CreateAdd(a0b0_0, Builder.CreateShl(mul_lo_hi, 32));
+    auto mul_hi = Builder.CreateAdd(mul_hi_lo, Builder.CreateShl(a1b1_1, 32));
+
+    if (IsSigned) {
+      // Apply the signed that we got from the previous if statement setting res_neg.
```
alan-baker, line 817:
```suggestion
      // Apply the sign that we got from the previous if statement setting res_neg.
```

alan-baker (changes requested):
Not a big fan of adding an Int64 requirement for a workaround. I'd suggest either handling the 32-bit case the same as the 64-bit case or adding a special case that checks if int64 is already used in the module.

rjodinchr:
@alan-baker, I did not make this merge branch commit. Is it something done by kokoro?

lib/ReplaceOpenCLBuiltinPass.cpp
rjodinchr: I have modified it so that 32x32 in 64 is done as 64x64 in 128. But the 64x64 in 128 is still using Int64 as it makes sense to require it in that case.

alan-baker (approved):

PR #830, opened 2022-04-01
fix replaceFshl to avoid spirv undefined behavior

rjodinchr:
Some CTS tests fail on some platforms when doing rotate of 0.
This is because rotate ends up to combination of shifts. One of them
being of type size in bits when rotating of 0.

But spirv says that shifting of the type size in bits results in an
undefined behavior.


rjodinchr:
I also need to update the tests for this one.
I will do that on Monday.

alan-baker (approved):

PR #832, opened 2022-04-01
Use python3 as the executable name.

woachk:
Starting from recent macOS releases, python(2) is no longer shipped
with the operating system at all. The script can already fetch LLVM
fine when using python3, so switching to that.


rjodinchr:
superseded by https://github.com/google/clspv/pull/1290

PR #836, opened 2022-04-04
enable support of non uniform workgroup size by default

rjodinchr:
Support of non uniform workgroup size also allow to split a NDRange in
several smaller regions.
We need that feature available by default to run the CTS on small
embedded GPUs even in CL1.2.
Allow to disable to feature with '-uniform-group-size'.

This is related to the work in kpet/clvk#367


kpet:
I meant we should enable it by default in clvk :). I'm not sure we want it enabled by default in clspv.

PR #839, opened 2022-04-14
Fix image1d buffer support

rjodinchr:
reading a image1d buffer needs to go through OpImageFetch and not
OpImageRead.
This is because OpImageRead requires the capability ReadWithoutFormat
which is not necessarily implemented on all vulkan compliant driver.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -1488,23 +1488,6 @@ SPIRVID SPIRVProducerPass::getSPIRVType(Type *Ty, bool needs_layout) {
   // stores a laid out and non-laid out version of the type.
   const unsigned layout = needs_layout ? 1 : 0;
 
-  if (dyn_cast<StructType>(Ty) && IsImageType(dyn_cast<StructType>(Ty))) {
-    // %opencl.image1d_buffer_ro.t, %opencl.image1d_buffer_wo.t and
-    // %opencl.image1d_buffer_rw.t needs to be rework as they will map on the
-    // same SPIRV type.
-    const char *bufferString = "_buffer_";
-    StringRef CanonicalName = Ty->getStructName();
-    size_t bufferPos = CanonicalName.find(bufferString);
-    if (bufferPos != std::string::npos) {
-      std::string NewName = CanonicalName.str().replace(
-          bufferPos + strlen(bufferString), 2, "rw");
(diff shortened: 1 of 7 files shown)


alan-baker:
The code looks ok, but how is this handled in clvk? How does it know to set the uniform texel buffer or storage texel buffer descriptor usage bits?

alan-baker:
cc @kpet

kpet:
There is currently no support for texel buffers in clvk and we might want to keep it that way :). My original suggestion to @rjodinchr was to treat 1D Buffer images the same as 1D images in clspv and use linearly tiled images aliasing the same memory as the underlying buffer in clvk. This approach can later on be generalised to cl_khr_image2d_from_buffer and cl_ext_image_from_buffer. While we could do things differently for 1D buffer images specifically, it would require special-casing in several code paths. It might buy us a bit more coverage short-term if Vulkan implementations support more formats for texel buffers than for linearly tiled images but that's AFAICT the only reason we'd want to use texel buffers.

alan-baker:
Binding an image with a buffer dimensionality seems to be going against the spec. Footnote 1 of Table 21 in https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-resources says buffer is only allowable for texel buffer bindings.

@kpet if they were emitted as normal 1D images wouldn't we potentially hit a coordinate limit that is not imposed by texel buffers?

alan-baker:
I'm going to merge this, but it seems clear that this will need revisited in the future. Using a sampled image is an improvement over a storage image and both have the same interface problem.

rjodinchr:
Thank you alan,
I will talk with kpet about it to see what to do on the clvk side.

kpet:
@alan-baker Finally managed to spend some time looking at this. I agree that texel buffers guarantee 64K elements are supported which is enough to satisfy the OpenCL requirement for 1D Buffer images. We may want to use linearly-tiled 1D  images in the future but having the option to use texel buffers is good and I'm going to add support to clvk (at least 1D buffer images will no longer be working by happy accident :)). I'll start by changing the descriptor interface: clspv currently report texel buffers as storage/sampled images which isn't quite right since the required descriptors and resource on the runtime side are different. I've drafted a proposed change to the non-semantic instruction set to add support for texel buffers: https://github.com/KhronosGroup/SPIRV-Registry/pull/172.

PR #841, opened 2022-04-25
Switch clspv to use LLVM's new pass manager

alan-baker:
* Update LLVM
* Removed deprecated sampler map, but left API entry points temporarily
  * use will result in an error
* Moved all passes to new pass manager style
  * every pass now has a header
  * TODO: cleanup passes for change in returns. Only top level run
    methods were changed
* Replaced pass registration
  * two new registration methods for mapping class names and parsing
    pipelines
* clspv-opt now uses the --passes option similar to llvm's opt
* updated several tests
  * common changes
    * disabling inlining on functions to target specific
    features
    * switched some tests from end-to-end to pass-specific lit tests


jrprice (approved):
LGTM!

PR #846, opened 2022-05-03
PointerBitcast: rework to support vec3 and array bitcast

rjodinchr:
Full rework of PointerBitcast.
The goal is to support cast to vec3 as well as cast with array as
source or destination.
Store of element bigger than the buffer element size are still not
handle.

This rework changes a few thing for alreay supported cast:
1) Only load the needed bytes.
   We used to load potentially more to only consider the appropriate
   part. But this part was buggy with bitcast feeding user GEPs.
2) Convert by using only bitcast/shuffle/extract/insert
   (except for the conversion involving structures).
   We used to use trunc and zext, but it is more efficient to use only
   the instructions above.


alan-baker (commented):
Is the intent to handle casts like: <3 x i8> to <4 x i32>? I don't see any tests covering vec3 on the source side.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,24 +208,269 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
+
+// 'Values' is expected to contain scalar values.
+// Group those values in vector of size 'NumElePerVec'.
+void GroupScalarValuesIntoVector(IRBuilder<> &Builder,
+                                 SmallVector<Value *, 8> &Values,
+                                 unsigned NumElePerVec) {
+  Type *SrcTy = Values[0]->getType();
+  assert(!SrcTy->isVectorTy() && !SrcTy->isArrayTy());
+  VectorType *DstTy = FixedVectorType::get(SrcTy, NumElePerVec);
+  unsigned int NumVector = Values.size() / NumElePerVec;
```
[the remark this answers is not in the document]
rjodinchr: yes it should.

rjodinchr:
> Is the intent to handle casts like: <3 x i8> to <4 x i32>? I don't see any tests covering vec3 on the source side.

Because the vec3 to vec4 pass is before it, I am not considering the case where vec3 is a source.
The intent is to handle cast with vec3 as an output, or array in either input or output.

The whole point is to support the `test_half` of the CTS. I am getting close with a lot more modification than this PR.
But those function changed during the rework of other casts, so I'll rework it so that you just have to review the end code, and not all the (buggy) intermediate versions.

rjodinchr:
I don't see anything wrong in the `CI-macos-clang-debug` log. Am I missing something?

rjodinchr:
> > I have run the CTS on this implementation:
> > 
> > * no regression
> > * test_half {vload|vloada|vstorea_rtz} are now passing
> > * test_half {vstore|vstorea} are still failing but not for compilation reasons now. They are segfaulting during kernel execution. I have looked at the generated SPIRV kernel, but it looked ok. I need to look at them more to understand what is going on.
> 
> update:
> 
> * test_half vstorea is failing because of an accuracy error (always very small, so should be related to the math floating point accuracy issue of OpenCL over Vulkan).
> * test_half vstore is failing when executing the kernel.

update:

on a Intel GPU, I only have accuracy errors:
```
# ./test_half

test_halfTest binary built May 19 2022 11:24:34
sizeof( void*) = 8
ARCH:   x86_64
system name = Linux
node name   = localhost
release     = 5.4.184-18100-ga66f997890d3
version     = #1 SMP PREEMPT Wed May 18 19:32:53 PDT 2022
machine     = x86_64
        Profile: Full
 Initializing random seed to 0.
Requesting Default device based on command line for platform index 0 and device index 0
Compute Device Name = Intel(R) UHD Graphics (TGL GT2), Compute Device Vendor = FIXME, Compute Device Version = OpenCL 1.2 CLVK on Vulkan v1.2.182 driver 88088581, CL C Version = OpenCL C 1.2 CLVK on Vulkan v1.2.182 driver 88088581
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
1 compute devices at 0.000000 GHz
Max thread group size is 2048.
Using const buffer size 0x10000 (65536)

Compute Device info:
        Device Name: Intel(R) UHD Graphics (TGL GT2)
        Vendor: FIXME
        Device Version: OpenCL 1.2 CLVK on Vulkan v1.2.182 driver 88088581
        OpenCL C Version: OpenCL C 1.2 CLVK on Vulkan v1.2.182 driver 88088581
        Driver Version: 1.2 CLVK on Vulkan v1.2.182 driver 88088581
        Processing with 1 devices
        Device Frequency: 0 MHz
        Has double? NO
        Test double? NO
vload_half...
........
vload_half passed
vloada_half...
........
vloada_half passed
vstore_half...
................vstore_half passed
vstorea_half...
................vstorea_half passed
vstore_half_rte...
................vstore_half_rte passed
vstorea_half_rte...
................vstorea_half_rte passed
vstore_half_rtz...
...
Failure at [393216] with 0x1.0c0000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [262144] with 0x1.080000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [131072] with 0x1.040000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [1] with 0x1.000002p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global
vstore_half_rtz FAILED
vstorea_half_rtz...
...
Failure at [393216] with 0x1.0c0000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [262144] with 0x1.080000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [131072] with 0x1.040000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [1] with 0x1.000002p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global
vstorea_half_rtz FAILED
vstore_half_rtp...

Failure at [393216] with 0x1.0c0000p-126: *0x0001 vs 0x0000,  vector_size = 1, address_space = global

Failure at [262144] with 0x1.080000p-126: *0x0001 vs 0x0000,  vector_size = 1, address_space = global

Failure at [131072] with 0x1.040000p-126: *0x0001 vs 0x0000,  vector_size = 1, address_space = global

Failure at [0] with 0x1.000000p-126: *0x0001 vs 0x0000,  vector_size = 1, address_space = global
vstore_half_rtp FAILED
vstorea_half_rtp...

Failure at [393216] with 0x1.0c0000p-126: *0x0001 vs 0x0000,  vector_size = 2, address_space = global

Failure at [262144] with 0x1.080000p-126: *0x0001 vs 0x0000,  vector_size = 2, address_space = global

Failure at [131072] with 0x1.040000p-126: *0x0001 vs 0x0000,  vector_size = 2, address_space = global

Failure at [0] with 0x1.000000p-126: *0x0001 vs 0x0000,  vector_size = 2, address_space = global
vstorea_half_rtp FAILED
vstore_half_rtn...
...
Failure at [393216] with 0x1.0c0000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [262144] with 0x1.080000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [131072] with 0x1.040000p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global

Failure at [1] with 0x1.000002p-25: *0x0000 vs 0x0001,  vector_size = 1, address_space = global
vstore_half_rtn FAILED
vstorea_half_rtn...
...
Failure at [393216] with 0x1.0c0000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [262144] with 0x1.080000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [131072] with 0x1.040000p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global

Failure at [1] with 0x1.000002p-25: *0x0000 vs 0x0001,  vector_size = 2, address_space = global
vstorea_half_rtn FAILED
roundTrip...
........
roundTrip passed
FAILED 6 of 6 sub-tests.
FAILED 6 of 13 tests.
```

rjodinchr:
`test_basic astype` is also passing using this [branch](https://github.com/rjodinchr/clspv/tree/bitcast-longvector) (which is based on this PR)

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +265,894 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
-            SrcAddrIdx = NewAddrIdx;
-          }
+  // Convert into 'Ty'
+  ConvertInto(Ty, Builder, Values);
+}
 
-          // Load value from src.
-          SmallVector<Value *, 8> LDValues;
+// Reduce SrcTy to do as less load/store operation as possible while not loading
+// unneeded data.
```
alan-baker, line 849:
```suggestion
// Reduce SrcTy to do as few load/store operations as possible while not loading
// unneeded data.
```

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +265,894 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
-            unsigned NumElement = 1;
-            if (SrcTy->isVectorTy()) {
-              NumElement = SrcVecTy->getElementCount().getKnownMinValue() * 2;
-            }
+  // Because the vec3 to vec4 pass is before this one, we should not have a vec3
+  // src. But it seems that some llvm passes after vec3 to vec4 can produce new
+  // vec3. At the moment the only case known is to produce vec3 that will be
+  // bitcast to another vec3 which element has the same time as the src vec3. In
+  // that particular case, just keep the vec3 as we only need to bitcast them,
+  // which will be handled correctly by this pass.
```
alan-baker, line 952: This is not clear. If the only known case of the source being a vec3 is when casting to another vec3 then wouldn't `IsVec3` already be true? Maybe this should be an assert?

alan-baker, line 952:
```suggestion
  // Because the vec3 lowering pass is run before this one, we should not have a vec3
  // src; however, it seems that some llvm passes after vec3 lowering can produce a new
  // vec3. At the moment the only case known is to produce a vec3 that will be
  // bitcast to another vec3 whose elements have the same size as the src vec3. In
  // that particular case, just keep the vec3 as we only need to bitcast them,
  // which will be handled correctly by this pass.
```

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +265,894 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
-              Value *DstAddr = Builder.CreateGEP(
-                  BaseAddr->getType()->getScalarType()->getPointerElementType(),
-                  BaseAddr, Idxs);
-              Type *TmpSrcTy = SrcEleTy;
-              if (auto TmpSrcVecTy = dyn_cast<VectorType>(TmpSrcTy)) {
-                TmpSrcTy = TmpSrcVecTy->getElementType();
-              }
-              Value *TmpVal = Builder.CreateBitCast(STValues[i], TmpSrcTy);
+  // Consider only the index for the size that has been loaded (the rest has
+  // already been consider during the load).
```
alan-baker, line 828:
```suggestion
  // Consider only the index for the size that has been loaded (the rest have
  // already been considered during the load).
```

alan-baker (changes requested):
I've taken a cursory look and the code seems mostly ok, but I'm not very fond of use of so many output parameters especially when combined with a lots of explanatory comments. It makes the code very difficult to follow. 

I still have much more to review, but I'd like to see some changes to make it easier to follow. Maybe leverage multiple return values or return structs, but definitely add more high level comments describing the expected inputs and outputs.

lib/ReplacePointerBitcastPass.cpp
rjodinchr: This condition `IsVec3` is used to avoid expecting a vec3 output from `combineInto`. Instead, we expect a vec4 output, and then we convert the vec4 to vec3.
When llvm creates this vec3 to vec3 bitcast with elements of the same size, we do not want/need to expect a vec4 output from `combineInto`. We can just ask `combineInto` to produce a vec3.

This is just to have a more efficient and understandable  output from this pass. Removing this condition would work as well, but it would produce unneeded transformation that is hard to understand when you just look at the test concerned by it.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -26,8 +26,65 @@ using namespace llvm;
...
+// Returns the size in bits of 'Ty'
+// In the case of a vector of 3 elements, return the size of the same vector of
+// 4 elements as vec3 has padding between 2 vectors.
+size_t SizeInBits(const DataLayout &DL, Type *Ty) {
+  if (auto VecTy = dyn_cast<FixedVectorType>(Ty)) {
+    if (VecTy->getNumElements() == 3) {
+      return 4 * SizeInBits(DL, VecTy->getElementType());
+    }
+  }
+  return DL.getTypeStoreSizeInBits(Ty);
```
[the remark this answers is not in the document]
rjodinchr: Good point, I will try it first to make sure.
Thanks

rjodinchr:
> I've taken a cursory look and the code seems mostly ok, but I'm not very fond of use of so many output parameters especially when combined with a lots of explanatory comments. It makes the code very difficult to follow.
> 
> I still have much more to review, but I'd like to see some changes to make it easier to follow. Maybe leverage multiple return values or return structs, but definitely add more high level comments describing the expected inputs and outputs.

Could you be more specific about which functions this comment applies to? Is it just the `ReduceType` function, or are all the functions working on the `SmallVector<Value *, 8>` concerned?

alan-baker:
I would say most of the helper functions. Especially those that take references to pointers, generally though at least document the output parameters.

alan-baker (approved):

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +258,941 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
+  }
+}
 
-            // Create store values.
-            Value *STVal = ST->getValueOperand();
+// 'Val' is expected to be a vector.
+// 'Idx' is the index where to extract the subvector, but in the casted type
+// coordinate. If null, just extract from the origin of the vector.
+// At the end of the function, 'Idx' has been updated with the potential
+// reminder of the index to get to the expected element.
```
alan-baker, line 755:
```suggestion
// remainder of the index to get to the expected element.
```

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +258,941 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
+}
 
-            if (DstTy->isVectorTy() && (SrcEleTyBitWidth != DstTyBitWidth)) {
-              VectorType *TmpVecTy = FixedVectorType::get(
-                  SrcEleTy, DstTyBitWidth / SrcEleTyBitWidth);
-              STVal = Builder.CreateBitCast(STVal, TmpVecTy);
-            }
+// 'Values' is expected to contain either vectors or scalars.
+// At the end of the function, 'Idx' has been updated with the potential
+// reminder of the index to get to the expected element.
```
alan-baker, line 787:
```suggestion
// remainder of the index to get to the expected element.
```

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -208,838 +258,941 @@ Value *ConvertValue(Value *src, Type *dst_type, IRBuilder<> &builder) {
...
-              DstNumElement = DstVecTy->getElementCount().getKnownMinValue();
-              for (unsigned i = 0; i < DstNumElement; i++) {
-                Value *Idx = Builder.getInt32(i);
-                Value *TmpVal = Builder.CreateExtractElement(STVal, Idx);
-                STValues.push_back(TmpVal);
-              }
-            }
+// 'Val' is expected to be a vector.
+// At the end of the function, 'Idx' has been updated with the potential
+// reminder of the index to get to the expected element.
```
alan-baker, line 829:
```suggestion
// remainder of the index to get to the expected element.
```

alan-baker (commented):
A few typos.

PR #851, opened 2022-05-12
Transition to opaque pointers

alan-baker:
* Remove pointer element type uses in cluster pod kernel arguments
* Make normalize global variables work with both types of pointers
  * duplicate IR tests as opaque versions


dneto0 (approved):
Nice!

PR #852, opened 2022-05-13
Handle llvm.{u,s}{add,sub}_sat intrinsics

### LINKED ISSUE #760: Weird compiler crash triggered when using `uint` as a parameter
Hi! Recently I encountered a weird crash of clspv, the simplified code is
```OpenCL
__kernel void shift(__global int *in, __global int *out, uint t0) {
    uint t = get_global_id(0);
    uint x;
    if(t > t0){
        x = t - t0;
    } else {
        x = 0;
    }
    out[t] = in[x];
}
```
clspv crashes when compiling this code using `clspv temp.cl -o temp.spv -O1` (or `-O2`, `-O3`), the output is
```
; Function Attrs: nofree nosync nounwind readnone speculatable willreturn
declare i32 @llvm.usub.sat.i32(i32, i32) #1
[1]    14310 segmentation fault (core dumped)  clspv temp.cl -o temp.spv -O1
```
exit code is 139

Here's what I've tried:
* When using`-O0` the error disappears
* If I change `uint t0` to `int t0` the error disappears
* the type of `t`, `x`, `in`, `out` seems not affecting the result (but not sure)
* If comment out `out[t] = in[x]` the error disappears
* moving `if` statment into a separate function doesn't work
  * ```OpenCL
    inline uint get_index(uint t, uint t0) {
        uint x;
        if(t > t0){
            x = t - t0;
        } else {
            x = 0;
        }
        return x;
    }

    __kernel void shift(__global uint *in, __global uint *out, uint t0) {
        uint t = get_global_id(0);
        uint x = get_index(t, t0);
        out[t] = in[x];
    }
    ```

I don't know whether it's a limitation of clspv or there's some bug in it. I'm not familiar with llvm either, thus don't know how to debug this, could you help me?

gnl21:
The existing handling for OpenCL builtins add_sat and sub_sat is used to lower the LLVM saturating intrinsics. Instead of lowering the OpenCL functions add_sat, sub_sat directly during the replace-opencl-builtins pass, they are converted to the LLVM intrinsics and then the existing lowering is moved to the replace-llvm-intrinsics pass. This gives correct handling for the intrinsics, which are sometimes generated during optimisation.

Fixes #760.

### DIFF
--- a/lib/ReplaceLLVMIntrinsicsPass.cpp
+++ b/lib/ReplaceLLVMIntrinsicsPass.cpp
@@ -24,8 +24,10 @@
 #include "clspv/Option.h"
 #include "spirv/unified1/spirv.hpp"
 
+#include "Builtins.h"
 #include "Constants.h"
 #include "ReplaceLLVMIntrinsicsPass.h"
+#include "SPIRVOp.h"
 
 using namespace llvm;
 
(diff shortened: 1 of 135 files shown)


alan-baker (approved):
Thanks very much!

PR #853, opened 2022-05-14
Support llvm.{s|u}{add|sub}_sat in long vector lowering

alan-baker:
* Recent refactoring (#852) requires handling these llvm intrinsics
* Fixes add_sat and sub_sat OpenCL CTS tests


dneto0 (commented):
LGTM

PR #856, opened 2022-05-31
Convert AllocateDescriptors to handle opaque pointers

alan-baker:
Contributes to #836

* Introduces type inference for opaque pointers to determine a
  reasonable data type to use in the translation
  * Examines the value and it's uses to determine the right type
  * Considers load/store/gep and many builtin functions
  * Will need improvement as more passes are converted, in particular
    image specialization will affect how images are identified
* Handle more cases in type mangling
  * atomics
  * reused structs in pointer parameters
* Identify mangled struct names as image and sampler types
* GetArgKind now accepts a data type parameter to handle opaque pointers


lib/Types.h
```diff
@@ -15,13 +15,21 @@
...
 #include "llvm/IR/Type.h"
 
 #include "spirv/unified1/spirv.hpp"
 
 namespace clspv {
 
+// Returns the inferred type of |v|.
+//
+// If the type of |v\ is an opaque pointer, this function traverses the uses of
+// |v| to determine the appropriate type.
```
dneto0, line 29: nit: Say this can return nullptr if no type can be inferred.

lib/Types.cpp
```diff
@@ -12,18 +12,176 @@
...
+        if (auto vec_ty = dyn_cast<VectorType>(data_ty))
+          return CacheType(VectorType::get(int32Ty, vec_ty));
+        else
+          return CacheType(int32Ty);
+      }
+      case clspv::Builtins::kBuiltinNone:
+        if (!call->getCalledFunction()->isDeclaration()) {
+          // See if the type can be inferred from the use in the called
+          // function.
+          auto *ty = InferType(call->getCalledFunction()->getArg(operand),
```
dneto0, line 139: It's good OpenCL C doesn't allow recursion.

lib/Types.cpp
```diff
@@ -64,7 +223,25 @@ bool clspv::IsImageType(llvm::StructType *STy) {
         STy->getName().startswith("opencl.image2d_array_wo_t") ||
         STy->getName().startswith("opencl.image3d_ro_t") ||
         STy->getName().startswith("opencl.image3d_rw_t") ||
-        STy->getName().startswith("opencl.image3d_wo_t")) {
+        STy->getName().startswith("opencl.image3d_wo_t") ||
```
dneto0, line 226: nit: I guess these are coming from a different frontend?

dneto0 (approved):
Looks good. Thanks!

Issue #816, opened 2022-02-14
Switch to opaque pointers


alan-baker:
LLVM is moving towards using opaque pointers generally. The information that used to be in the pointer element type is available on GEPs and load/store instructions. Clspv needs to switch uses of `getNonOpaquePointerElementType()` to more robust ways of accessing the desired type.

Conversions needed:

- [x] AllocateDescriptors (bidirectional)
- [x] AutoPodArgs
- [x] ClusterPodKernelArguments
- [x] Builtins (only used in type mangling)
- [x] DefineOpenCLWorkItemBuiltins
- [x] DirectResourceAccess
- [x] LongVectorLowering (partial, just missing memcpy)
- [x] MultiVersionUBOFunctions
- [x] NormalizeGlobalVariables (bidirectional)
- [x] PushConstants
- [x] ReplaceLLVMIntrinsics (memset and memcpy)
- [x] ReplaceOpenCLBuiltin
- [x] ReplacePointerBitcast (bidirectional)
- [x] SPIRVProducer (partial)
- [x] SimplifyPointerBitcast
- [x] SpecializeImageTypes (bidirectional)
- [x] ThreeElementVectorLowering
- [x] UBOTypeTransform (partial)
- [x] UndoByval
- [x] UndoInstCombine
- [x] ZeroInitializeAllocas
- [x] UndoSRet
- [x] Types
- [x] Builtin library from libclc
- [x] InlineFuncWithPointerBitCastArg (bidirectional)
- [x] llvm.memset and llvm.memcpy (@rjodinchr)

The overall strategy for the transition is:
1. Support both modes concurrently.
2. Get CTS and important benchmarks passing with opaque pointers.
3. Switch the default to opaque.
4. Remove support for transparent pointers after some period of stability.

alan-baker:
For the SPIRVProducer, something along the lines of splitting getSPIRVType into two methods: getSPIRVType and getSPIRVPointerType should work. The new method would look something like:
```c++
SPIRVID SPIRVProducerPass::getSPIRVPointerType(Type *PtrTy, Type* EleTy) {
  if (!isa<PointerType>(PtrTy)) {
    llvm_unreachable("Use getSPIRVType instead");
  }

  const unsigned addr_space = PtrTy->getPointerAddressSpace();
  const bool needs_layout = PointerRequiresLayout(addr_space);
  const uint32_t layout_idx = needs_layout ? 1 : 0;

  if (addr_space != AddressSpace::UniformConstant) {
[...]

PR #862, opened 2022-06-03
DRA: Replace use of getPointerElemenType

dneto0:
Track the pointee type of the base object.
That pointee type is used as the GEP pointer interpretation type.

Contributes to #816

### DIFF
--- a/lib/DirectResourceAccessPass.cpp
+++ b/lib/DirectResourceAccessPass.cpp
@@ -93,6 +93,11 @@ bool clspv::DirectResourceAccessPass::RewriteResourceAccesses(Function *fn) {
   return Changed;
 }
 
+// Rewrites, if possible, all the arg_index'th arg to all callers to 'fn',
+// to replace passing the resource by pointer argument to directly accessing
+// the underlying global variable for the resource.  This is only possible
+// when all callers call 'fn' with the same underlying resource variable.
+// Returns true if a change was made.
 bool clspv::DirectResourceAccessPass::RewriteAccessesForArg(Function *fn,
                                                             int arg_index,
                                                             Argument &arg) {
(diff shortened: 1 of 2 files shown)


lib/DirectResourceAccessPass.cpp
```diff
@@ -106,22 +111,30 @@ bool clspv::DirectResourceAccessPass::RewriteAccessesForArg(Function *fn,
   // such a thing (where the GEP can only have zero indices).
   struct ParamInfo {
     // The base value. It is either a global variable or a resource-access
-    // builtin function. (@clspv.resource.var.* or @clspv.local.var.*)
+    // builtin function. In the global variable case it is either
+    // @clspv.resource.var.* or @clspv.local.var.*
```
alan-baker, line 115: This comment doesn't make sense to me. It seems like it should either be a resource/local variable or a global variable.

lib/DirectResourceAccessPass.cpp
```diff
@@ -106,22 +111,30 @@ bool clspv::DirectResourceAccessPass::RewriteAccessesForArg(Function *fn,
...
     unsigned num_gep_zeroes;
-    // An example call fitting
+    // A sample call using this function argument as the resource described
+    // above.
     CallInst *sample_call;
   };
-  // The common valid parameter info across all the callers seen soo far.
 
-  bool seen_one = false;
+  // The common valid parameter info across all the callers seen soo far.
```
alan-baker, line 134:
```suggestion
  // The common valid parameter info across all the callers seen so far.
```

lib/DirectResourceAccessPass.cpp
```diff
@@ -167,19 +182,26 @@ bool clspv::DirectResourceAccessPass::RewriteAccessesForArg(Function *fn,
               dyn_cast<ConstantInt>(call->getOperand(0))->getZExtValue());
           const auto binding = uint32_t(
               dyn_cast<ConstantInt>(call->getOperand(1))->getZExtValue());
-          if (!merge_param_info({callee, set, binding, num_gep_zeroes, call}))
+          auto *resource_type = call->getOperand(6)->getType();
```
alan-baker, line 185: Not on you, but maybe lib/Constants.h should get enums for the operands of the resource functions (and other special clspv builtins). The raw `6` isn't great for maintainability.

alan-baker (changes requested):
Some small issues with the comments.

lib/DirectResourceAccessPass.cpp
dneto0: yeah. For another CL.

PR #863, opened 2022-06-06
Handle opaque pointers in image specialization

alan-baker:
Contributes to #816

* Image specialization now handles both transparent and opaque pointers
  * Much of the code can be removed after moving fully to opaque
    pointers
  * Doesn't use InferTypes because the pass's traversal is necessary and would
    simply be repeated
  * Specialized builtin functions are now remangled from an updated type instead of having a
    suffix tacked on
* Fixed some image query functions to support mangled struct names
* Fixed some builtin demangling
* Changed how opaque pointers is disabled
  * added a new command line option to control it
  * sets the code gen option
  * remove the command line parsing


lib/Builtins.cpp
```diff
@@ -233,11 +233,14 @@ bool Builtins::FunctionInfo::GetFromMangledNameCheck(
   while (pos < mangled_name_len) {
     ParamTypeInfo type_info;
     if (mangled_name[pos] == 'S') {
-      // handle duplicate param_type
-      if (mangled_name[pos + 1] != '_') {
+      // handle duplicate param_type. Comes in two flavours:
+      // S_ and S#_.
+      char p1 = mangled_name[pos + 1];
+      char p2 = mangled_name[pos + 2];
```
dneto0, line 239: In isolation, I'm concerned about buffer overflow
Suggest guarding the access by pos+2 with the first test.


```
     if (p1 != '_' && (mangled_name[pos+2] != '_')) {
                return false;
     }
```

lib/SpecializeImageTypes.cpp
```diff
@@ -139,47 +198,46 @@ Type *SpecializeImageTypesPass::RemapUse(Value *value, unsigned operand_no) {
...
-          if (auto new_type = RemapUse(U.getUser(), U.getOperandNo())) {
-            return new_type;
-          }
-        }
-      }
+      return std::make_pair(ResultType::kNotSpecialized, struct_ty);
       break;
     }
-  } else if (IsImageType(value->getType())) {
+  } else if (value->getType()->isPointerTy()) {
```
dneto0, line 224: nit: When does this case occur?

Is it analogous to OpCopyObject of what looks like a pointer argument?  But I'm not sure if that happens in the LLVM IR.  A comment here would be helpful.

dneto0 (changes requested):
Minor change requested.

Does this test the cases where type remapping recurses into user helper functions?

lib/SpecializeImageTypes.cpp
alan-baker: I know it happens for address space casts. I'll add a comment.

alan-baker:
Added an extra test for the subfunctions. It was covered by the transparent pointers, but not opaque pointers.

PR #865, opened 2022-06-08
Transition SPIRVProducer to opaque pointers P1

alan-baker:
Contributes to #816

* Convert simple cases of getPointerElementType() in SPIRVProducerPass
  to finding the type in other ways
  * generally uses the data type in the resource functions
* Added an enum for resource function operands
* Fixed a bug in UBOTypeTransformPass that was not properly visiting
  constant operands
  * This is now frequently covered by resource variable functions

### DIFF
--- a/lib/Constants.h
+++ b/lib/Constants.h
@@ -23,6 +23,27 @@ namespace clspv {
 // Name for module level metadata storing workgroup argument spec ids.
 const std::string &LocalSpecIdMetadataName();
 
+enum ClspvOperand {
+  // Operands for workgroup variables.
+  kWorkgroupSpecId = 0,
+  kWorkgroupDataType = 1,
+
+  // Operands for resource variables.
+  kResourceDescriptorSet = 0,
+  kResourceBinding = 1,
+  kResourceArgKind = 2,
+  kResourceArgIndex = 3,
+  kResourceDiscriminantIndex = 4,
+  kResourceCoherent = 5,
+  kResourceDataType = 6,
+
+  // Operands for literal samplers.
+  kSamplerDescriptorSet = 0,
+  kSamplerBinding = 1,
+  kSamplerParams = 2,
+  kSamplerDataType = 3,
+};
+
 // Base name for workgroup variable accessor function.
 const std::string &WorkgroupAccessorFunction();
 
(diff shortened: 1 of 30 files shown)


lib/UBOTypeTransformPass.cpp
```diff
@@ -332,7 +329,7 @@ bool clspv::UBOTypeTransformPass::RemapUser(User *user, Module &M) {
   bool changed = RemapValue(user, M);
 
   for (Use &use : user->operands()) {
-    User *operand_user = use.getUser();
+    User *operand_user = dyn_cast<User>(use.get());
```
alan-baker, line 332: Previously this just returned User the function was called with! Didn't cause problems because it was generally never acted on due to the filter below. Now it properly traverses constant operands. This gets covered by many existing tests using the new data type in the resource function. Wasn't a problem until the SPIRVProducerPass starting using those types.

lib/UBOTypeTransformPass.cpp
```diff
@@ -295,18 +295,15 @@ bool clspv::UBOTypeTransformPass::RemapGlobalVariables(
     SmallVectorImpl<GlobalVariable *> *variables_to_modify, Module &M) {
   bool changed = false;
   for (auto &GV : M.globals()) {
-    if (auto *ptr_ty = dyn_cast<PointerType>(GV.getType())) {
-      auto *remapped = RebuildType(ptr_ty->getNonOpaquePointerElementType(), M);
-      if (ptr_ty->getNonOpaquePointerElementType() != remapped) {
-        changed = true;
-        variables_to_modify->push_back(&GV);
-      }
+    auto *remapped = RebuildType(GV.getValueType(), M);
```
dneto0, line 298: Nice that this is simpler, clearer code.

lib/Constants.h
```diff
@@ -23,6 +23,27 @@ namespace clspv {
 // Name for module level metadata storing workgroup argument spec ids.
 const std::string &LocalSpecIdMetadataName();
 
+enum ClspvOperand {
```
dneto0, line 26: Thanks for this cleanup

lib/SPIRVProducerPass.cpp
```diff
@@ -4383,16 +4397,10 @@ void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
     // Find SPIRV instruction for parameter type.
     auto Ty = I.getType();
     if (Ty->isPointerTy()) {
-      auto PointeeTy = Ty->getPointerElementType();
-      if (PointeeTy->isStructTy() &&
```
dneto0, line 4387: Interesting we can just delete this clause. But I think we we can.

dneto0 (approved):
Nice work.

lib/SPIRVProducerPass.cpp
alan-baker: Right, this would be invalid SPIR-V anyways so why special case for it? If this ever came up as being generated, it should be fixed earlier in the flow.

PR #879, opened 2022-06-28
add support for get_image_array_size

rjodinchr:


alan-baker (changes requested):
Please add more test coverage. Prefer to cover all dimensionalities and sample types (e.g. write_only images too). It's also preferable to make the input IR and run the producer through clspv-opt.

alan-baker (approved):
Thanks very much.

PR #882, opened 2022-07-08
add support for get image channel order and data_type

rjodinchr:
Add a pass to inline function containing either
function (get_channel_order or get_channel_data_type) to be able to
know which image is in input.

Add a pass to set metadatas on each kernel containing either function,
as well as metadata on each call to those functions.
Those metadatas are used in SPIRVProducer to generate to appropriate
offset and nonsemantic instructions.

Regroup RedeclareGlobalPushConstants under PushConstants.{cpp,h} to be
able to use them in SetImageChannelMetadataPass.

Do not use per-kernel push constant interfaces if the kernel call
either of the channel functions. We need the pod kernel arguments when
we redeclare globalpushconstant to add info about image channel.


lib/SetImageChannelMetadataPass.cpp
```diff
@@ -0,0 +1,153 @@
...
+PreservedAnalyses
+clspv::SetImageChannelMetadataPass::run(Module &M, ModuleAnalysisManager &) {
+  PreservedAnalyses PA;
+
+  ImageGetterMap MapOrder, MapDataType;
+  OffsetMap off_map;
+
+  // Go through function and instruction to look for image metadata getter
+  // function
+  for (Function &F : M) {
```
alan-baker, line 126: Would It be simpler to handle each kernel at a time and have that handling return the max offset needed for that kernel? There doesn't really seem to be an advantage to mapping the pair of kernel and image argument.

Also it might be easier to combine the channel order and formats into a single map. You could use a tombstone key for an unused entry.

lib/SetImageChannelMetadataPass.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  // Create and return the structure that will contains the needed values
+  std::vector<Type *> orderTypes(max_elements,
+                                 IntegerType::get(M.getContext(), 32));
+  StructType *Ty = StructType::get(M.getContext(), orderTypes);
+
+  // All of them are not ChannelOrder, but it does not matter, as it will just
+  // be used to be skipped in spirvproducer.
+  clspv::RedeclareGlobalPushConstants(M, Ty,
+                                      (int)clspv::PushConstant::ChannelOrder);
```
alan-baker, line 113: This makes me a bit nervous. It feels a bit kludgy. I agree we should minimize how many times we redeclare the variable, but is there some way not to abuse the data?

alan-baker: Maybe its as simple as using a different enum (e.g. PushConstant::ImageMetadata).

lib/Compiler.cpp
```diff
@@ -626,6 +627,7 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     // This pass mucks with types to point where you shouldn't rely on
     // DataLayout anymore so leave this right before SPIR-V generation.
     pm.addPass(clspv::UBOTypeTransformPass());
+    pm.addPass(clspv::SetImageChannelMetadataPass());
```
alan-baker, line 630: This probably deserves a comment on the dependency with inlining.

lib/AutoPodArgsPass.cpp
```diff
@@ -91,6 +126,7 @@ void clspv::AutoPodArgsPass::runOnFunction(Function &F) {
   // 4. No arrays.
   // 5. If 16-bit types are used, 16-bit push constants are supported.
   // 6. If 8-bit types are used, 8-bit push constants are supported.
+  // 7. Not to have a image channel getter function call.
```
alan-baker, line 129: The documentation about this function support should be updated. That documentation should also describe the current limitations. For example, clspv does not account for the right push constant size yet when channel order or channel formats are used.

test/ImageBuiltins/get_image_channel_multiple_kernel.cl
```diff
@@ -0,0 +1,52 @@
...
+// CHECK:  [[ptr_pc_struct:%[^ ]+]] = OpTypePointer PushConstant [[pc_struct]]
+// CHECK:  [[ptr_pc_int:%[^ ]+]] = OpTypePointer PushConstant %uint
+
+// CHECK:  [[pc_var:%[^ ]+]] = OpVariable [[ptr_pc_struct]] PushConstant
+
+// CHECK:  [[fct0:%[^ ]+]] = OpFunction
+// CHECK:  OpAccessChain [[ptr_pc_int]] [[pc_var]] %uint_1 %uint_0
+// CHECK:  [[fct1:%[^ ]+]] = OpFunction
+// CHECK:  OpAccessChain [[ptr_pc_int]] [[pc_var]] %uint_1 %uint_0
+// CHECK:  [[fct2:%[^ ]+]] = OpFunction
```
alan-baker, line 38: Better would be to get the id for each kernel from the OpEntryPoint instruction based on the name and then reuse that id here and elsewhere.

test/InlineFuncWithImageMetadataGetter/get_image_channel_data_type.ll
```diff
@@ -0,0 +1,36 @@
...
+define dso_local spir_kernel void @foo(i32 addrspace(1)* align 4 %dst, %opencl.image2d_ro_t addrspace(1)* %image) {
+entry:
+  %call = call spir_func i32 @get(%opencl.image2d_ro_t addrspace(1)* %image)
+  store i32 %call, i32 addrspace(1)* %dst, align 4
+  ret void
+}
+
+; CHECK:  define dso_local spir_func i32 @bar
+; CHECK:  call spir_func i32 @_Z27get_image_channel_data_type14ocl_image2d_ro
+; CHECK:  ret i32
```
alan-baker, line 32: The important part is inlining into the kernel function so I'd probably drop this.

test/InlineFuncWithImageMetadataGetter/get_image_channel_data_type.ll
```diff
@@ -0,0 +1,36 @@
...
+
+; CHECK:  define dso_local spir_kernel void @foo
+; CHECK:  call spir_func i32 @_Z27get_image_channel_data_type14ocl_image2d_ro
```
alan-baker, line 35: Can this be strengthened to CHECK-NEXT?

test/InlineFuncWithImageMetadataGetter/get_image_channel_order.ll
```diff
@@ -0,0 +1,36 @@
...
+  ret void
+}
+
+; CHECK:  define dso_local spir_func i32 @bar
+; CHECK:  call spir_func i32 @_Z23get_image_channel_order14ocl_image2d_ro
+; CHECK:  ret i32
+
+; CHECK:  define dso_local spir_kernel void @foo
+; CHECK:  call spir_func i32 @_Z23get_image_channel_order14ocl_image2d_ro
+; CHECK:  ret void
```
alan-baker, line 36: Same as the previous test

alan-baker (changes requested):
Seems reasonable overall for an initial implementation. Definitely needs some documentation on the current limitations. Please also file an issue to track the remaining work.

lib/SetImageChannelMetadataPass.h
```diff
@@ -0,0 +1,28 @@
...
+#include "llvm/IR/Module.h"
+#include "llvm/IR/PassManager.h"
+
+#ifndef _CLSPV_LIB_SET_IMAGE_CHANNEL_METADATA_GETTER_H
+#define _CLSPV_LIB_SET_IMAGE_CHANNEL_METADATA_GETTER_H
+
+namespace clspv {
+struct SetImageChannelMetadataPass
+    : llvm::PassInfoMixin<SetImageChannelMetadataPass> {
+  llvm::PreservedAnalyses run(llvm::Module &M, llvm::ModuleAnalysisManager &);
```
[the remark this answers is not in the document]
rjodinchr: Done

lib/SetImageChannelMetadataPass.cpp
```diff
@@ -0,0 +1,153 @@
...
+    for (unsigned i = 0; i < fct_md->getNumOperands(); i++) {
+      MDs.push_back(fct_md->getOperand(i));
+    }
+  }
+}
+
+void setMetadatas(Module &M, ImageGetterMap &map, OffsetMap &off_map, int pc) {
+  auto i32 = IntegerType::get(M.getContext(), 32);
+
+  for (auto order : map) {
```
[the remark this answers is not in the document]
rjodinchr: I used `std::map` instead of `DenseMap`.

lib/SetImageChannelMetadataPass.cpp
rjodinchr: I went for the unique pushconstant enum, and I used another enum to differentiate order and data_type

lib/Compiler.cpp
rjodinchr: Done

test/InlineFuncWithImageMetadataGetter/get_image_channel_data_type.ll
rjodinchr: I have drop it

test/InlineFuncWithImageMetadataGetter/get_image_channel_data_type.ll
rjodinchr: Done

test/InlineFuncWithImageMetadataGetter/get_image_channel_order.ll
rjodinchr: Done

lib/AutoPodArgsPass.cpp
rjodinchr: I am not sure to understand what documentation I should update, could you precise please?
Also, I don't get where clspv tells the right push constant size?

alan-baker: docs/OpenCLOnVulkan.md describes the state of support that clspv has for opencl.

There is a section that lists get_image_channel_order() and get_image_channel_data_type() as unsupported. 

There is an option in clspv, `-max-pushconstant-size`, that specifies the max size of push constants. It defaults to the minimum requirement in Vulkan, but clvk will pass the right value for the implementation. Clspv uses this option to determine if POD args will fit in the push constants (also accounting for the other members, e.g. region offset). If they can't fit clspv has to choose another implementation (UBO). The size of the image metadata is unknown at this point in the flow and there is currently no fall back implementation for image metadata. If the POD args (plus earlier elements) fit in the limit, but the image metadata pushes that size over the limit the SPIR-V will be invalid for the implementation. This will lead to incorrect results or a crash in clvk. I'm ok with limiting the initial implementation, but we need to advertise that limitation until the fallback is implemented.

rjodinchr: I have updated the documentation

lib/SetImageChannelMetadataPass.cpp
```diff
@@ -57,14 +51,15 @@ void concatWithFunctionMetadata(Function *F, MetadataVector &MDs) {
...
 
-void setMetadata(Module &M, ImageGetterMap &map, OffsetMap &off_map, int pc) {
+unsigned setMetadata(Module &M, Function *F, ImageGetterMap &map) {
   auto i32 = IntegerType::get(M.getContext(), 32);
 
-  for (auto order : map) {
-    auto F = order.first.first;
-    auto image = order.first.second;
+  unsigned int count = 0;
+  for (auto elem : map) {
```
alan-baker, line 58: This is still non-deterministic. Pointers are effectively random values. Hash maps are always non-deterministic no matter the key, but ordered maps are non-deterministic for pointer keys.

alan-baker (changes requested):
Still need a better solution than iterating over the ordered map.

alan-baker (approved):
I think the Amber bot is failing because it hasn't rolled a new SPIRV-Headers in some time.

lib/AutoPodArgsPass.cpp
```diff
@@ -52,6 +52,39 @@ PreservedAnalyses clspv::AutoPodArgsPass::run(Module &M,
...
+              if (visited_fct.count(f) == 0) {
+                next_fcts_to_visit.push_back(f);
+              }
+            }
+          }
+        }
+      }
+    }
+    fcts_to_visit = std::move(next_fcts_to_visit);
+    next_fcts_to_visit.clear();
```
dneto0, line 82: This is triggering the use-after-move Clang check.

https://clang.llvm.org/extra/clang-tidy/checks/bugprone/use-after-move.html 

Seems the declaration of next_fcts_to_visit should be moved inside the loop.  I don't think this is performance critical to warrant tolerating the bug.

lib/SetImageChannelMetadataPass.cpp
```diff
@@ -0,0 +1,136 @@
...
+      MDs.push_back(fct_md->getOperand(i));
+    }
+  }
+}
+
+unsigned setMetadata(Module &M, Function *F, ImageGetterMap &map) {
+  auto i32 = IntegerType::get(M.getContext(), 32);
+
+  unsigned int count = 0;
+  for (auto elem : map) {
```
dneto0, line 58: This is triggering: https://clang.llvm.org/extra/clang-tidy/checks/performance/for-range-copy.html 
'elem' can be const reference.

PR #884, opened 2022-07-12
roundup integer type to be a multiple of 8

rjodinchr:
Most vulkan drivers do not support integer type not being a multiple
of 8.
Make sure we always produce integer type with a size multiple of 8.

This will fix a compilation issue with libclc lgamma and lgamma_r
functions.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -1468,7 +1468,14 @@ SPIRVID SPIRVProducerPassImpl::addSPIRVGlobalVariable(const SPIRVID &TypeID,
 }
 
 Type *SPIRVProducerPassImpl::CanonicalType(Type *type) {
-  if (type->getNumContainedTypes() != 0) {
+  if (type->isIntegerTy()) {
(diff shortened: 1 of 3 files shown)


rjodinchr:
Yes they do on Intel, AMD and ARM GPUs. I have not tested it with Nvidia.
I have run the CTS test without any flag.

lib/SPIRVProducerPass.cpp
```diff
@@ -1927,6 +1927,11 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVType(Type *Ty, bool needs_layout) {
     uint32_t bit_width =
         static_cast<uint32_t>(Canonical->getPrimitiveSizeInBits());
 
+    if (bit_width > 1) {
```
alan-baker, line 1930: I think this might need to be moved to `CanonicalType()`. As is, clspv might generate duplicate integer types which is invalid in SPIR-V.  For example if the kernel generated the i2 and also used i8, clspv would cache two different `OpTypeInt 8 0` instructions. Make sure to add a test that covers that situation.

alan-baker (changes requested):

alan-baker (approved):

PR #886, opened 2022-07-15
Fix incorrect POD arg struct layout for kernels with long-vector POD args

callumfare:

Long-vector POD args are lowered to arrays. When these args are clustered in a struct, changing these member types can change the layout of the struct. This can result in the clustered POD arg struct not having the same layout as the corresponding reflection metadata.

Fix this behavior by using the same remapped type offset/size metadata as the UBO Type Transform Pass to preserve the original layout.

The test case here comes from the basic/parameter_types test from the OpenCL CTS.

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -437,6 +437,49 @@ Value *convertVectorOperation(Instruction &I, Type *EquivalentReturnTy,
   return ReturnValue;
 }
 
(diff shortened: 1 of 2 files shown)


callumfare:
> @callumfare, I'm going to wait to review until the CLA gets signed.

Hi @alan-baker - the CLA is signed now, sorry for the delay, we had to clarify some things on our side. A review of this would be appreciated.

alan-baker (approved):

PR #889, opened 2022-07-15
Support opaque pointers in ReplacePointerBitcast and InlineFuncWithPointerBitCastArg

alan-baker:
Contributes to #816

* Inline functions where there is an implicit pointer cast in the
  arguments
* Support implicit pointer casts in replace pointer bitcasts
* Cover alloca and global variables for inferring types from the result
* Remove some unnecessary pointer casts in ReplaceOpenCLBuiltin when
  using opaque pointers

### DIFF
--- a/lib/InlineFuncWithPointerBitCastArgPass.cpp
+++ b/lib/InlineFuncWithPointerBitCastArgPass.cpp
@@ -157,8 +157,8 @@ bool clspv::InlineFuncWithPointerBitCastArgPass::InlineFunctions(Module &M) {
               auto *next_gep = cast<GetElementPtrInst>(user);
               if (result_ele_ty &&
                   next_gep->getResultElementType() != result_ele_ty) {
-                // TODO: this is an over-approximation, but it's easier to not
-                // worry about phis.
+                // TODO: this is an over-approximation, but it is necessary
+                // unless we want to traverse phis multiple times.
                 add_all = true;
               }
               for (auto &use : user->uses()) {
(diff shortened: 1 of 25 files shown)


alan-baker:
Worth noting that both major modifications are basically just identifying implicit pointer casts because the bitcast instructions should no longer appear with opaque pointers.

lib/InlineFuncWithPointerBitCastArgPass.cpp
```diff
@@ -93,6 +94,86 @@ bool clspv::InlineFuncWithPointerBitCastArgPass::InlineFunctions(Module &M) {
...
+                to_check.push_back(
+                    std::make_pair(use.getUser(), use.getOperandNo()));
+              }
+              break;
+            case Instruction::GetElementPtr: {
+              auto *next_gep = cast<GetElementPtrInst>(user);
+              if (result_ele_ty &&
+                  next_gep->getResultElementType() != result_ele_ty) {
+                // TODO: this is an over-approximation, but it's easier to not
+                // worry about phis.
```
dneto0, line 161: nit: This is a funny comment.  It talks about Phis, but this is the GEP case.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -522,8 +594,16 @@ clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
...
+    // TODO(#816): remove after final transition.
+    Value *start = isa<BitCastInst>(Inst) ? Inst : Src;
+    for (User *BitCastUser : start->users()) {
+      if (ImplicitCasts.count(BitCastUser)) {
+        // If this user was queued on the worklist as an implicit cast
+        // separately, don't handle it now.
+        continue;
+      }
+
+     LLVM_DEBUG(dbgs() << "#### BitCastUser: "; BitCastUser->dump());
```
dneto0, line 606: nit: indent one space :-)

dneto0 (approved):
Thanks!

PR #891, opened 2022-07-17
add support for opencl compliant division

rjodinchr:
This is an implementation based on OpFDiv to avoid loosing too much
performance.
When possible it can be reduced to a single OpFDiv if we are divided
by a constant value.

This implementation passes the CTS for divide/half_divide on Intel,
AMD, ARM and Nvidia.

Also remove kHalfRecip and kHalfDivide from ReplaceOpenCLBuiltinPass
as they are in fact managed through clpsv--.bc.

Apply NoContraction Decoration to a bit more fp operands as GPU like
Intel are doing optimization with some of them (CTS contractions test
are failing without it on OpFNegate for example).


rjodinchr:
> When you say "passes CTS" is that wimpy more or full mode?

I mean full mode with all the vector size.

lib/SPIRVProducerPass.cpp
```diff
@@ -4434,6 +4461,158 @@ void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
         Ops.clear();
         Ops << Ty << Cmp << Mod << Modm1;
         RID = addSPIRVInst(spv::OpSelect, Ops);
+      } else if (I.getOpcode() == Instruction::FDiv && !Option::UnsafeMath()) {
+        // float div(float a, float b) {
```
alan-baker, line 4465: What happens with other types? Double for sure should be handled, but I'm not certain that half divide could appear.

rjodinchr: I agree, devices supporting those types we will need to handle them differently.
Let's reduce this implementation only for fp32 to start with.

lib/SPIRVProducerPass.cpp
```diff
@@ -4434,6 +4461,159 @@ void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
...
+
+
+        const float fmax_val = 0x1.0p+126f;
+        const float fmin_val = 0x1.0p-126f;
+        const float c_max_val = 0x1.0p-4f;
+        const float c_min_val = 0x1.0p+24f;
+        SPIRVID divisor, c;
+
+        // Try to optimize it if b is constant.
+        if (auto b_cst = dyn_cast<ConstantFP>(I.getOperand(1))) {
```
alan-baker, line 4494: There's a lot of repeated code here. What about something along the lines of:
```
const bool is_vector = FPTy->isVectorTy();
const auto num_eles = is_vector ? cast<VectorType>(FPTy)->getNumElements() : 1;
if (auto *cst = dyn_cast<Constant>(I.getOperand(1))) {
  SmallVector<Constant *, 2> constants;
  SmallVector<Constant *, 2> divisor_vec;
  SmallVector<Constant *, 2> c_vec;
  if (is_vector) {
    for (int i = 0; i < num_eles; ++i) {
      constants.push_back(cst->getAggregateElement(i));
    }
  } else {
    constants.push_back(cst);
  }

  bool c_vec_needed = false;
  for (auto *ele : constants) {
    // Process like vector case
  }

  // Construct final values
}
```

If this is less code it's probably worthwhile for readability.

alan-baker (changes requested):
I think this works, but I wonder if the code can be simplified.

lib/SPIRVProducerPass.cpp
rjodinchr: indeed, it works well, thanks

alan-baker (approved):

PR #897, opened 2022-07-26
Avoid performance-for-range-copy clang-tidy warning

dneto0:


dneto0:
"loop variable is copied but only used as const reference; consider making it a const reference"

https://clang.llvm.org/extra/clang-tidy/checks/performance/for-range-copy.html

rjodinchr (approved):

alan-baker (approved):

Issue #826, opened 2022-03-06
Support for physical addressing


kpet:
## Motivation

- Full support for OpenCL C requires physical addressing. While clspv has done a great job of enabling many OpenCL C programs without the use of physical addressing, there will always be programs that cannot be compiled without physical addressing.
- clspv's pointer bitcast rewriting passes are rather complex and tend to require changes when large new code bases are compiled.
- Some OpenCL CTS tests would be easy to pass with physical addressing (e.g. support for `NULL` kernel arguments).
- Physical addressing opens the door to supporting OpenCL SVM or USM which in addition to being desirable features in their own right are also very useful to provide a bridge to enable other programming languages to target Vulkan.

## Proposed design outline

Here's a rough outline of what I prototyped:

- Use the SPV_KHR_physical_storage_buffer extension to enable the use of the **PhysicalStorageBufferEXT** storage class where **StorageBuffer** is currently used.
- Disable pointer transformation passes (pointer bitcast replacement and simplification passes).
- Introduce new runtime interfaces in the clspv non-semantic instruction set to allow passing pointers obtained by the runtime using `VK_KHR_buffer_device_address` via either push constants or uniform buffers. The interface uses a structure of 64-bit integers (or 2-element vectors of 32-bit integers), one for each pointer argument, that is part of the global push constant structure or stored in a dedicated uniform buffer. Pointer arguments to kernel functions are rewritten to load the required integers from this structure and convert them to pointers. A new pass is introduced to declare the interface structure and perform the necessary rewriting.

## Proposed staging

- [ ] Support 64-bit addressing model (hard requirement of the **PhysicalStorageBuffer64EXT** addressing model)
    - [x] Add libclc target (https://reviews.llvm.org/D116668)
    - [ ] Build libclc as part of clspv's build system
    - [ ] Add support for 64-bit addressing model
- [ ] Update clspv non-semantic instruction set with new runtime interface instructions
    - [ ] Specification
    - [ ] Headers
    - [ ] Use in clspv and other dependent projects (e.g. SPIR-V Tools)
- [ ] Introduce utilities to manage global push constants
- [ ] Add support for physical addressing (using 64-bit integers)
- [ ] Add support for physical addressing (using 2-element vectors of 32-bit integers)

alan-baker:
Seems reasonable overall. A few points to consider:
1. You can only disable pointer transforms for global (and sometimes constant) address space.
[...]

PR #903, opened 2022-08-02
Build libclc as part of the build system

callumfare:
Build libclc as part of the build system rather than relying on a checked-in
bitcode file. This somewhat increases the duration of the build as clang, opt,
etc are built as dependencies.

This is one of the tasks mentioned in #826

Also work around a bug in the libclc cmake by adding some extra LLVM library
dependencies.


alan-baker:
The failed builds are complaining about a CMake problem.

callumfare:
I can see that most jobs are failing because the library is being built with opaque pointers by default. This can in theory be fixed by passing `-DCMAKE_CLC_FLAGS=-Xclang\ -no-opaque-pointers` but there's another mistake in the libclc cmake that means it can't parse the extra flags properly.

I can put up a patch to fix this, unless it's not worth it if non-opaque pointer support will be dropped imminently. On that note we might have capacity to help out with the transition to opaque pointers (#816) if there are specific areas/passes that aren't being worked on right now.

callumfare:
Fixes for the two libclc cmake issues: https://reviews.llvm.org/D131266

alan-baker:
Thanks for looking into this. I admit that this PR is not the easiest thing for me to integrate into our internal builds. I do think it is the right long term direction compared to keeping a pre-built library checked in, but I need to work through the internal issues before closing on this.

rjodinchr:
superseeded by #1316

Issue #712, opened 2021-02-24
clspv crashes when code contains function declaration


mantognini:
The following OpenCL C code results in a crash because `foo` is not defined.

```
void foo(void);

kernel void bar() {
  foo();
}
```

I'm not sure whether this should be supported by clspv or not.

Here is the error:
```
; Function Attrs: convergent
declare spir_func void @foo() local_unnamed_addr #1
Unsupported Constant???
UNREACHABLE executed at /work/checkouts/downstreams/clspv/clspv/lib/SPIRVProducerPass.cpp:2190!
```
and the backtrace:
```
(gdb) bt
#0  0x00007ffff705ffb7 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff7061921 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007fffe809eb56 in llvm::llvm_unreachable_internal (msg=0x7ffff79817e7 "Unsupported Constant???", file=0x7ffff7981220 "clspv/lib/SPIRVProducerPass.cpp", line=2190)
    at clspv/third_party/llvm/llvm/lib/Support/ErrorHandling.cpp:213
#3  0x00007ffff77b759b in (anonymous namespace)::SPIRVProducerPass::getSPIRVConstant (this=0x5555557aeca0, Cst=0x55555586b128) at clspv/lib/SPIRVProducerPass.cpp:2190
#4  0x00007ffff77b77a8 in (anonymous namespace)::SPIRVProducerPass::getSPIRVValue (this=0x5555557aeca0, V=0x55555586b128) at clspv/lib/SPIRVProducerPass.cpp:2214
#5  0x00007ffff77c0ac0 in (anonymous namespace)::SPIRVProducerPass::HandleDeferredInstruction (this=0x5555557aeca0) at clspv/lib/SPIRVProducerPass.cpp:4782
#6  0x00007ffff77b13b7 in (anonymous namespace)::SPIRVProducerPass::runOnModule (this=0x5555557aeca0, M=...) at clspv/lib/SPIRVProducerPass.cpp:788
#7  0x00007ffff150a4e7 in (anonymous namespace)::MPPassManager::runOnModule (this=0x555555d49290, M=...) at clspv/third_party/llvm/llvm/lib/IR/LegacyPassManager.cpp:1549
[...]

PR #904, opened 2022-08-03
delete llvm.assume calls 

omarahmed1111:
# PR changes:
* **Delete llvm.assume declaration and its calls**: as llvm.assume will require SPV_KHR_expect_assume extension which is not supported by Vulkan at the moment. So, we delete it for now. llvm.assume is a compiler hint so should be safe to ignore. One of the CTS tests that was affected: basic/test_basic kernel_call_kernel_function.
* **Print a meaningful error when there is unsupported llvm intrinsic call or unhandled function declaration/definition**: the current code will handle unsupported llvm intrinsic calls or unlinked function declarations by treating them as llvm::Constants and eventually will fail with ("unsupported Constant???") error. That should be handled so any future call for llvm intrinsic that is not supported or unlinked function declarations is identified.
* **Fix make_test.py for passing ir module to it**
* **Catch the error when kernel argument have no meta data**: This is to prevent segmentation fault when we use `-cl-kernel-info` option and kernel argument have no meta data.

## Related issues
* https://github.com/google/clspv/issues/712 : Not an example for llvm.assume but the failed function declaration in the issue example should give a more concise error message with this PR.

### DIFF
--- a/lib/ReplaceLLVMIntrinsicsPass.cpp
+++ b/lib/ReplaceLLVMIntrinsicsPass.cpp
@@ -37,15 +37,14 @@ PreservedAnalyses
 clspv::ReplaceLLVMIntrinsicsPass::run(Module &M, ModuleAnalysisManager &) {
   PreservedAnalyses PA;
 
-  // Remove lifetime annotations first.  They could be using memset
-  // and memcpy calls.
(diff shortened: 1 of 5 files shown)


alan-baker:
Looks good, but please look into the Windows build failures.

PR #905, opened 2022-08-04
add support for load/store from gep cst expr

rjodinchr:
also fix gep from bitcast cst expr

### DIFF
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -702,4 +702,52 @@ void ConvertInto(Type *Ty, IRBuilder<> &Builder,
   }
 }
 
+bool RemovedCstExprFromFunction(Function *F) {
+  SmallVector<std::pair<Instruction *, unsigned>, 16> WorkList;
+
+  auto CheckInstruction = [&WorkList](Instruction *I) {
+    for (unsigned OperandId = 0; OperandId < I->getNumOperands(); OperandId++) {
+      if (auto CE = dyn_cast<ConstantExpr>(I->getOperand(OperandId))) {
+
+        // TODO(#816): remove after final transition.
+        // This case happen in some particular scenario, where it is not
+        // needed to simplify it. Mostly when using global array
+        // variables. In the end llvm managed to deal with it without us
+        // having to simplify it. Trying to simplify would make it very
+        // complicated for the ReplacePointerBitcast pass.
+        if (CE->getOpcode() == Instruction::BitCast &&
+            !CE->getOperand(0)->getType()->isOpaquePointerTy() &&
+            CE->getOperand(0)
+                ->getType()
+                ->getNonOpaquePointerElementType()
+                ->isStructTy())
+          continue;
+
+        WorkList.push_back(std::make_pair(I, OperandId));
+      }
+    }
+  };
+
+  for (BasicBlock &BB : *F) {
+    for (Instruction &I : BB) {
+      CheckInstruction(&I);
+    }
+  }
+
+  bool Changed = !WorkList.empty();
+
+  while (!WorkList.empty()) {
+    auto *I = WorkList.back().first;
+    auto OperandId = WorkList.back().second;
+    WorkList.pop_back();
+
+    auto Operand =
+        dyn_cast<ConstantExpr>(I->getOperand(OperandId))->getAsInstruction(I);
+    CheckInstruction(Operand);
+    I->setOperand(OperandId, Operand);
+  }
+
+  return Changed;
+}
+
 } // namespace BitcastUtils
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -18,6 +18,7 @@
 #include "llvm/Pass.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "BitcastUtils.h"
 #include "SimplifyPointerBitcastPass.h"
 
 using namespace llvm;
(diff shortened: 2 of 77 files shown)


alan-baker:
I wonder whether we should simply lower all ConstantExpr to instructions (using ConstantExpr::getAsInstruction). That pass could get run (probably twice) and it would prevent patching up these in an ad hoc manner.

rjodinchr:
> I wonder whether we should simply lower all ConstantExpr to instructions (using ConstantExpr::getAsInstruction). That pass could get run (probably twice) and it would prevent patching up these in an ad hoc manner.

Sure, sounds like a good idea, I'll do that.

lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -44,9 +44,11 @@ clspv::SimplifyPointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
   return PA;
 }
 
-bool clspv::SimplifyPointerBitcastPass::runOnGEPFromBitcastCstExpr(
+bool clspv::SimplifyPointerBitcastPass::runOnInstFromCstExpr(
```
alan-baker, line 47: I guess I would have expected something more generic here to just flatten out all constant expressions. Along the lines of:
```
for all i in instructions {
  for (auto o : i.operands()) {
    if (isa<ConstantExpr>(o)) {
      worklist.push_back(std::make_pair(o, i));
    }
  }
}

while !worklist.empty() {
  ConstantExpr *ce = worklist.back().first;
  Instruction *insert = worklist.back().second;
  worklist.pop_back();

  auto new_i = ce->getAsInstruction(insert);
  for (auto o : new_i->operands()) {
    if (isa<ConstantExpr>(o)) {
      worklist.push_back(std::make_pair(o, new_i));
    }
  }
}
```

Is there something I'm missing? I could also see this moving to a utility or separate pass if it's more appropriate.

alan-baker (changes requested):
It would be good to add some chained constant expressions in the tests too.

lib/SimplifyPointerBitcastPass.cpp
rjodinchr: I have created a utility function in BitcastUtils that is now used in SimplifyPointerBitcast as well as LongVectorLowering and ThreeElementVectorLowering.

lib/BitcastUtils.cpp
```diff
@@ -702,4 +702,33 @@ void ConvertInto(Type *Ty, IRBuilder<> &Builder,
   }
 }
 
+bool RemovedCstExprFromFunction(Function *F) {
```
alan-baker, line 705: Almost every call to this function does it in a loop. I'd suggest making this worklist driven and hitting all the constant expressions at once to simplify the callers.

lib/BitcastUtils.cpp
```diff
@@ -702,4 +702,33 @@ void ConvertInto(Type *Ty, IRBuilder<> &Builder,
...
+          // needed to simplify it. Mostly when using global array
+          // variables. In the end llvm managed to deal with it without us
+          // having to simplify it. Trying to simplify would make it very
+          // complicated for the ReplacePointerBitcast pass.
+          if (CE->getOpcode() == Instruction::BitCast &&
+              CE->getOperand(0)
+                  ->getType()
+                  ->getNonOpaquePointerElementType()
+                  ->isStructTy())
+            continue;
```
alan-baker, line 723: Is this necessary? Does LLVM only handle this oddball case because it is a constantexpr or will it handle the case if there are instructions instead? Ideally this could be removed.

If it has to stay, please a `TODO(#816): remove after final transition` and also guard that the  operand type is a transparent pointer.

lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -44,50 +45,13 @@ clspv::SimplifyPointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
...
-                       ->isStructTy()) {
-                // ... record the GEP as something we need to process.
-                WorkList.push_back(GEP);
-              }
-            }
-          }
-        }
-      }
-    }
+    Changed |= BitcastUtils::RemovedCstExprFromFunction(&F);
```
alan-baker, line 51: Refactor this to the run method when changing the function to hit every constantexpr.

alan-baker (changes requested):

lib/BitcastUtils.cpp
rjodinchr: Done

lib/BitcastUtils.cpp
rjodinchr: It is unfortunately still necessary.
here is an example of the problematic input in ReplacePointerBitCast (`NormalizeGlobalVariables/constant_bitcast.cl`):
```
 %1 = bitcast <{ <{ [9 x i32], [8 x i32] }>, [17 x i32], [17 x i32], [17 x i32] }> addrspace(2)* @data to [4 x [17 x i32]] addrspace(2)*
 %3 = getelementptr inbounds [4 x [17 x i32]], [4 x [17 x i32]] addrspace(2)* %1, i32 0, i32 %2, i32 %y
 %4 = load i32, i32 addrspace(2)* %3, align 4
 ```
 At the moment the pass does not support it. But llvm seems to end up simplifying it by itself.

lib/SimplifyPointerBitcastPass.cpp
rjodinchr: Not sure what you expected here

alan-baker (approved):

PR #906, opened 2022-08-08
fix alignment of structure when lowering long vectors

rjodinchr:
vectors are aligned on their size. But when lowering them to arrays,
it changes their alignment to be on the alignment of the element.

When lowering a structure make sure that we are keeping the offset
right. If not add a padding element.

Also make sure that the structure has the same size before and after
the lowering. If not add a padding element at the end.

As the lowered type might have more element than the original one,
rework the indices when lowering gep extractvalue or insertvalue

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -313,7 +313,8 @@ Function *getBIFScalarVersion(Function &Builtin) {
 /// @return an equivalent pointer when both @p V and @p newType are pointers.
 /// @return an equivalent vector when @p V is an aggregate.
 /// @return an equivalent aggregate when @p V is a vector.
-Value *convertEquivalentValue(IRBuilder<> &B, Value *V, Type *EquivalentTy) {
+Value *convertEquivalentValue(IRBuilder<> &B, Value *V, Type *EquivalentTy,
+                              const DataLayout *DL) {
   if (V->getType() == EquivalentTy) {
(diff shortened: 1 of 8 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -1382,19 +1419,36 @@ Type *clspv::LongVectorLoweringPass::getEquivalentTypeImpl(Type *Ty) {
...
       }
     }
 
     if (RequiredLowering) {
-      return StructType::get(Ctx, Types, false);
+      auto InitialSize = DL->getTypeAllocSize(StructTy);
+      auto NewSize = DL->getTypeAllocSize(StructType::get(Ctx, Types, Packed));
+      if (InitialSize != NewSize) {
+        Types.push_back(
+            ArrayType::get(Ty->getInt8Ty(Ctx), InitialSize - NewSize));
```
alan-baker, line 1449: This seems to add an Int8 requirement that might not otherwise exist. I suggest using an i32 array if the size is evenly divisible by 4.

alan-baker (changes requested):
Seems good overall, but I'd like to avoid introducing smaller data types whenever possible.

lib/LongVectorLoweringPass.cpp
rjodinchr: Done

alan-baker (approved):

PR #909, opened 2022-08-12
Transition SimplifyPointerBitcastPass to opaque pointers

omarahmed1111:
# PR changes
* **Modify `SimplifyPointerBitcastPass::runOnBitcastFromGEP`**: This function should simplify cases when we bitcast a pointer that is generated from gep instruction. Before opaque pointers that case was valid as we could transform a pointer type to another pointer type like (i64* to i32*). After opaque pointer changes that case can't happen and will always be a trivial bitcast as we don't have any pointer type other than "ptr". Trivial bitcasts will be handled with `runOnTrivialBitcast`. So, we should delete this function after full transformation to opaque pointers.  
Example: `runOnTrivialBitcast` will delete the bitcast instruction. 
```
%gep = getelementptr i32, ptr %ptr, i32 0     
%cast = bitcast ptr %gep to ptr
```

* **Modify `SimplifyPointerBitcastPass::runOnGEPFromBitcastCstExpr`**: This function should simplify cases when we have gep instruction with its pointer operand as a bitcast constant expression. Which will have a similar trivial case as `SimplifyPointerBitcastPass::runOnBitcastFromGEP`. So, we should delete this function after full transformation to opaque pointers. 
Example:  
`%gep = getelementptr i32, ptr bitcast (ptr %ptr to ptr), i32 0`  
LLVM passes will transform it to `getelementptr i32, ptr %ptr, i32 0`

* **Modify `SimplifyPointerBitcastPass::runOnGEPFromGEP**: This function handles cases when a gep instruction pointer operand is another gep instruction. Implicit casts could occur after opaque pointer transition. So, we make sure that there is no implicit cast before optimizing this cases.

* **Add some trivial and simple bitcast test cases**

### Related issues
https://github.com/google/clspv/issues/816

### DIFF
--- a/test/PointerCasts/trivial_casts.ll
+++ b/test/PointerCasts/trivial_casts.ll
@@ -1,0 +1,81 @@
+; RUN: clspv-opt %s -o %t --passes=simplify-pointer-bitcast
+; RUN: FileCheck %s < %t
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK-LABEL: define void @test1() {
+; CHECK: entry:
+; CHECK:   %0 = add i32 0, 4
+; CHECK:   %gep = getelementptr i32, i32* null, i32 %0
+; CHECK:   ret void
+; CHECK: }
+
+define void @test1() {
+entry:
+  %0 = add i32 0, 4
+  %cast = bitcast i32 %0 to i32
+  %gep = getelementptr i32, i32* bitcast (float* null to i32*), i32 %0
+  ret void
+}
+
+; CHECK-LABEL: define void @test2() {
+; CHECK: entry:
+; CHECK:   %gep = getelementptr i64, i64* null, i32 250
+; CHECK:   ret void
+; CHECK: }
+
+define void @test2() {
+entry:
+  %cast1 = bitcast i32 bitcast ( i32 250 to i32 ) to float
+  %cast2 = bitcast float %cast1 to i32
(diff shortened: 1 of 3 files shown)


test/PointerCasts/trivial_casts.ll
```diff
@@ -0,0 +1,68 @@
...
+; CHECK: }
+
+define void @test4(i32* %in, float* %out) {
+entry:
+  %gep = getelementptr i32, i32* %in, i32 1     
+  %cast = bitcast i32* %gep to float*
+  %val = load float, float* %cast
+  store float %val, float* %out
+  ret void
+}
```
omarahmed1111, line 68: Not a trivial bitcast but thought adding it as a prove of concept for opaque pointer transformation. as `opaque_trivial_casts.ll` have its similar version but with opaque pointer (which will be trivial). It also should cover a case that is not currently tested for `runOnBitcastFromGEP` (this function should be deleted after full transformation anyway).

alan-baker (changes requested):
Mostly there. I think the GEP from GEP case needs additional checking now in an opaque pointer world though. There might be an implicit pointer cast between them so I think we need a check that the result element type of the first GEP is the same as the source element type of the second GEP. I believe that also works with transparent pointers too. It would be good to make sure there is coverage of those cases for opaque pointers too in the tests.

omarahmed1111:
> Mostly there. I think the GEP from GEP case needs additional checking now in an opaque pointer world though. There might be an implicit pointer cast between them so I think we need a check that the result element type of the first GEP is the same as the source element type of the second GEP. I believe that also works with transparent pointers too. It would be good to make sure there is coverage of those cases for opaque pointers too in the tests.

Fixed and added small simple tests for that.

alan-baker (approved):

PR #910, opened 2022-08-16
Improve support for elided GetElementPtr instructions

callumfare:
LLVM will optimize out the GEP to the first element of a vector or aggregate type when opaque pointers are enabled, as there is no real pointer arithmetic to be done. This needs explicitly handled to prevent the correct access chain being omitted from the SPIR-V.

The ReplacePointerBitcastPass will reinsert elided GEPs where possible.

This fixes some work item builtin calls like `get_global_id(0)`.

### DIFF
--- a/lib/ReplacePointerBitcastPass.cpp
+++ b/lib/ReplacePointerBitcastPass.cpp
@@ -103,6 +103,38 @@ Value *ExtractSubVector(IRBuilder<> &Builder, Value *&Idx, Value *Val,
   return Val;
 }
 
+// Check whether a pointer to the type `ContainingTy` is also usable as a
+// pointer to the type `TargetTy` due to the layout of the vector or aggregrate
+// type. Descends through the first element of the `ContainingTy` until it is
+// found or it cannot descend any further. Writes out levels of indirection to
+// `Steps`.
+bool FindAliasingContainedType(Type *ContainingTy, Type *TargetTy, int &Steps) {
+  int StepCount = 0;
+
+  do {
+    if (ContainingTy == TargetTy) {
+      Steps = StepCount;
+      return true;
+    }
+    StepCount++;
+
+    if (auto *VectorTy = dyn_cast<VectorType>(ContainingTy)) {
+      ContainingTy = VectorTy->getElementType();
+    } else if (auto *ArrayTy = dyn_cast<ArrayType>(ContainingTy)) {
+      ContainingTy = ArrayTy->getArrayElementType();
+    } else if (auto *StructTy = dyn_cast<StructType>(ContainingTy)) {
+      if (StructTy->isOpaque())
+        break;
+      ContainingTy = StructTy->getStructElementType(0);
+    } else {
+      break;
+    }
+  } while (ContainingTy->isAggregateType() || ContainingTy->isVectorTy() ||
+           ContainingTy == TargetTy);
+
+  return false;
+}
+
 // 'Values' is expected to contain either vectors or scalars.
 // At the end of the function, 'Idx' has been updated with the potential
 // remainder of the index to get to the expected element.
@@ -449,6 +481,7 @@ clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
   SmallVector<User *, 16> UserWorkList;
   DenseMap<Value *, Type *> type_cache;
   DenseSet<Value *> ImplicitCasts;
+  DenseMap<Instruction *, int> ImplicitGEPs;
   for (auto &F : M) {
     for (auto &BB : F) {
       for (auto &I : BB) {
@@ -474,6 +507,15 @@ clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
         }
 
         if (source_ty && dest_ty && source_ty != dest_ty) {
+          int Steps = 0;
+          if (!isa<GetElementPtrInst>(I) &&
+              FindAliasingContainedType(source_ty, dest_ty, Steps)) {
+            if (Steps > 0) {
+              ImplicitGEPs.insert({&I, Steps});
+              continue;
+            }
+          }
+
           bool ok = true;
           UserWorkList.push_back(&I);
           while (!UserWorkList.empty()) {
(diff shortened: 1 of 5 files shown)


callumfare:
There are couple of things I wasn't sure about for fixing this:
* The changes to ReplacePointerBitcastPass aren't needed if the UndoInstCombine pass is scheduled before them but I was hesitant to add it to the pipeline another 2 times
* Instead of changing the UndoInstCombine pass the SPIRVProducerPass could catch the case of mismatched load/store types and insert the OpAccessChain directly when handling Load and Store instructions. It seemed like the UndoInstCombine changes matched more closely with existing fixes but I'm happy to do it this way instead

lib/UndoInstCombinePass.cpp
```diff
@@ -224,3 +225,30 @@ bool clspv::UndoInstCombinePass::UndoWideVectorShuffleCast(Instruction *inst) {
...
+  auto load = dyn_cast<LoadInst>(inst);
+  if (!load)
+    return false;
+
+  auto gv = dyn_cast<GlobalVariable>(load->getPointerOperand());
+  if (!gv)
+    return false;
+
+  auto value_ty = gv->getValueType();
+  if (auto* value_vec_ty = dyn_cast<VectorType>(value_ty)) {
```
alan-baker, line 239: My understanding is that LLVM will elide any gep with all 0 indices. The vector case is definitely the most common for clspv.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -474,6 +474,14 @@ clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
         }
 
         if (source_ty && dest_ty && source_ty != dest_ty) {
+          // LLVM can elide the GEP of the first element of a vector when using
+          // opaque pointers, skip this case
+          if (auto *source_vec_ty = dyn_cast<VectorType>(source_ty)) {
```
alan-baker, line 479: There seems to be a mismatch between this change and the one in UndoInstCombine that only looks for global variables.

lib/UndoInstCombinePass.cpp
```diff
@@ -224,3 +225,30 @@ bool clspv::UndoInstCombinePass::UndoWideVectorShuffleCast(Instruction *inst) {
 
   return true;
 }
+
+bool clspv::UndoInstCombinePass::UndoGEPRemove(Instruction *inst) {
+  auto load = dyn_cast<LoadInst>(inst);
+  if (!load)
+    return false;
+
+  auto gv = dyn_cast<GlobalVariable>(load->getPointerOperand());
```
alan-baker, line 234: This seems more restrictive than necessary. I wonder whether we should leverage `InferType()` here.

I was playing around with a toy kernel:
```
kernel void zero_copy(global int4* in, global int* out) {
  *out = in[0].x;
}
```

The IR prior to the SPIRVProducer looks like:
```
target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
target triple = "spir-unknown-unknown"

define dso_local spir_kernel void @zero_copy(ptr addrspace(1) nocapture readonly align 16 %in, ptr addrspace(1) nocapture writeonly align 4 %out) local_unnamed_addr #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !6 !kernel_arg_type !7 !kernel_arg_base_type !8 !kernel_arg_type_qual !9 !clspv.pod_args_impl !10 {
entry:
  %0 = call ptr addrspace(1) @_Z14clspv.resource.0(i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, { [0 x <4 x i32>] } zeroinitializer)
  %1 = getelementptr { [0 x <4 x i32>] }, ptr addrspace(1) %0, i32 0, i32 0, i32 0
  %2 = call ptr addrspace(1) @_Z14clspv.resource.1(i32 0, i32 1, i32 0, i32 1, i32 1, i32 0, { [0 x i32] } zeroinitializer)
  %3 = getelementptr { [0 x i32] }, ptr addrspace(1) %2, i32 0, i32 0, i32 0
  %4 = load <4 x i32>, ptr addrspace(1) %1, align 16
  %5 = extractelement <4 x i32> %4, i64 0
  store i32 %5, ptr addrspace(1) %3, align 4
  ret void
}
```

If I run instcombine on this IR it will drop the geps. ReplacePointerBitcasts will choke on the result because it can't calculate the bitwidth of the runtime array properly to figure out the number of iterations properly. So if we can identify this case generally it would be very helpful.

alan-baker (changes requested):
I like the idea behind this. Given that replace pointer bitcasts has trouble with some casts, it would be good to generalize the solution here. Can we identify a simple missing 0 gep in the general case and either handle it in UndoInstCombine or ReplacePointerBitcasts? I like the simplicity of the UndoInstCombine addition, but it might be more robust to handle this in ReplacePointerBitcasts. It might depend on the ease of identifying the general form of this elided gep.

callumfare:
Thanks for the review. I did also wonder about using `InferType` and being able to catch more cases, I think something more robust is possible so I'll give that a go

callumfare:
@alan-baker 

I've updated this PR based on your feedback. The missing GEPs are now handled by the ReplacePointerBitcastPass, and it supports the general case of implicit GEPs with arrays, vectors and structs (including nested types).

Currently it supports Load and Store instructions but skips GEPs. I made the change for GEPs as well, but a few lit tests failed. For example, the test PointerCasts/opaque_load_cast_of_array.ll fails with the following output:
```llvm
...
entry:
  %cast = getelementptr [4 x [8 x i32]], ptr addrspace(3) %b, i32 0
  %0 = getelementptr [4 x [8 x i32]], ptr addrspace(3) %cast, i32 0, i32 0, i32 0
  %gep = getelementptr i32, ptr addrspace(3) %0, i32 %n
  %ld = load i32, ptr addrspace(3) %gep, align 4
...
```

You can see that it uses another GEP instead of the expected lshr/and instructions to recalculate the address. This seems like simpler IR but I'm not sure if I'm missing something that means this pattern isn't ok for the actual SPIR-V generation. What do you think? If this is desirable I can allow GEPs as well and updated the failing tests.

alan-baker:
> @alan-baker
> 
> I've updated this PR based on your feedback. The missing GEPs are now handled by the ReplacePointerBitcastPass, and it supports the general case of implicit GEPs with arrays, vectors and structs (including nested types).
> 
> Currently it supports Load and Store instructions but skips GEPs. I made the change for GEPs as well, but a few lit tests failed. For example, the test PointerCasts/opaque_load_cast_of_array.ll fails with the following output:
> 
> ```llvm
> ...
> entry:
>   %cast = getelementptr [4 x [8 x i32]], ptr addrspace(3) %b, i32 0
>   %0 = getelementptr [4 x [8 x i32]], ptr addrspace(3) %cast, i32 0, i32 0, i32 0
>   %gep = getelementptr i32, ptr addrspace(3) %0, i32 %n
>   %ld = load i32, ptr addrspace(3) %gep, align 4
> ...
> ```
> 
> You can see that it uses another GEP instead of the expected lshr/and instructions to recalculate the address. This seems like simpler IR but I'm not sure if I'm missing something that means this pattern isn't ok for the actual SPIR-V generation. What do you think? If this is desirable I can allow GEPs as well and updated the failing tests.

Unfortunately, that gep would result in an OpPtrAccessChain being generated. That is worse for two reasons:
1. It requires variable pointers
2. It would be UB if %n were greater or equal to 8

Combining the two geps into `getelementptr i32, ptr addrspace(3) %cast, i32 0, i32 0, i32 %n` is also incorrect. The strategy that is currently used (using div and rem) is the only one I'm aware that produces correct code.

alan-baker (approved):
Looks good, thanks!

PR #913, opened 2022-08-19
UBOTypeTransformPass: Handle opaque pointers

FMarno:
# Changes
UBOTypeTransform will now read the underlying type of pointer arguments from the clspv resource builtins.
I also removed an unneeded copy of the offsets when building the new struct type and ensured attributes were copied when created new global variables and functions.

# Related Issue
#816


alan-baker (approved):
Thanks for the contribution!

PR #917, opened 2022-08-25
Add FixupBuiltinsPass to ensure NaN check is performed on sqrt and rsqrt

aarongreig:
This is the first step towards removing these builtins from our libclc
SOURCES and relying entirely on the GLSL instructions.

FixupBuiltinsPass is intended to be generic, and an alternative to
ReplaceOpenCLBuiltinsPass that doesn't necessitate erasing the original
function from the module.


alan-baker:
Have you tested the fixed up functions against CTS? If so, which implementations have you tested?

aarongreig:
I've tested it successfully against swift shader and intel with sqrt and rsqrt removed from my builtins library so that the fixed up functions are generated by default. I also tested whether the CTS would reject results other than nan for the root of a negative and it does seem to, so I think retaining the check for that is worth it.

alan-baker (approved):

aarongreig:
https://reviews.llvm.org/D134040 LLVM revision to remove sqrt and rsqrt from our libclc SOURCES

alan-baker:
Thanks for pushing that change.

kpet:
https://reviews.llvm.org/D134040 now landed in LLVM main.

PR #920, opened 2022-09-01
Transition three element vector lowering pass to opaque pointers

omarahmed1111:
# PR Describtion
Transform ThreeElementVectorLoweringPass to support both opaque and non-opaque pointers:
* Change the main logic of the pass to save all instructions that need to be changed, and change them together at the end. The motivation for that is that opaque pointers types need to be inferred from other instructions so if we change them directly that will add an overhead of analysing the original types before instructions change.
* Add support for opaque pointers in `visitGetElementPtrInst`, `visitLoadInst`, `visitCallInst`, `convertOpCopyMemoryOperation`, `visitStoreInst`, `HandlingRequired` and `runOnGlobals`.
* Add support for recursive type inferring for loads/stores with pointer operands only.
* Migrate `vec3-to-vec4` tests to have an opaque version.
* Add `simple-vec3-to-vec4-opaque.ll` test which should be a simple test for different parts that had support for opaque pointers.

### [Related issue](https://github.com/google/clspv/issues/816)

### DIFF
--- a/lib/ThreeElementVectorLoweringPass.cpp
+++ b/lib/ThreeElementVectorLoweringPass.cpp
@@ -34,6 +34,7 @@
 #include "BitcastUtils.h"
 #include "Builtins.h"
 #include "ThreeElementVectorLoweringPass.h"
+#include "Types.h"
 
 #include <array>
 #include <functional>
@@ -217,36 +218,6 @@ std::string getVec4Name(const clspv::Builtins::FunctionInfo &IInfo) {
   return clspv::Builtins::GetMangledFunctionName(Info);
 }
 
-/// In order not to overflow, we need to copy elements one by one when
-/// CopyMemory arguments are transformed from vec3 to vec4.
-Value *convertOpCopyMemoryOperation(CallInst &VectorCall,
-                                    ArrayRef<Value *> EquivalentArgs) {
-  auto *DstOperand = EquivalentArgs[1];
-  auto *SrcOperand = EquivalentArgs[2];
-  assert(DstOperand->getType()->getPointerElementType()->isVectorTy());
-  assert(dyn_cast<VectorType>(DstOperand->getType()->getPointerElementType())
-             ->getElementCount()
-             .getKnownMinValue() == 4);
-
-  IRBuilder<> B(&VectorCall);
-  Value *ReturnValue = nullptr;
-
-  // for each element
-  for (unsigned eachElem = 0; eachElem < 3; eachElem++) {
-    auto *SrcGEP = B.CreateGEP(
-        SrcOperand->getType()->getScalarType()->getPointerElementType(),
-        SrcOperand, {B.getInt32(0), B.getInt32(eachElem)});
-    auto *Val =
-        B.CreateLoad(SrcGEP->getType()->getPointerElementType(), SrcGEP);
-    auto *DstGEP = B.CreateGEP(
-        DstOperand->getType()->getScalarType()->getPointerElementType(),
-        DstOperand, {B.getInt32(0), B.getInt32(eachElem)});
-    ReturnValue = B.CreateStore(Val, DstGEP);
-  }
-
-  return ReturnValue;
-}
-
 /// SIMD Builtin are builtin where the instruction uses only 1 data element
 bool isBuiltinSIMD(clspv::Builtins::BuiltinType Builtin) {
   if (Builtin > clspv::Builtins::kType_Math_Start &&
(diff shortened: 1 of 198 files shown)


lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -294,24 +275,50 @@ bool vec3ShouldBeLowered(Module &M) {
...
 
-} // namespace
+bool clspv::ThreeElementVectorLoweringPass::vec3BitcastInFunction(Function &F) {
+  for (Instruction &I : instructions(F)) {
+    if (auto *Inst = dyn_cast<CastInst>(&I)) {
+      auto *Type = Inst->getSrcTy();
+      if (Type->isPointerTy() && !Type->isOpaquePointerTy()) {
+        auto *PointeeType = Type->getNonOpaquePointerElementType();
+        if (auto *VectorType = dyn_cast<FixedVectorType>(PointeeType)) {
+          if (VectorType->getElementCount().getKnownMinValue() == 3)
```
omarahmed1111, line 285: From my understanding this should be fixing bug with no effect. As the previous logic when it sees a vector type element, it will process the pass even when the vector type is not of size 3 which will do nothing in this case. This should fix that and only run the pass when it finds vectors of type 3.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -804,13 +837,17 @@ clspv::ThreeElementVectorLoweringPass::visitUnaryOperator(UnaryOperator &I) {
...
     return true;
   }
 
   for (auto &Operand : U.operands()) {
     auto *OperandTy = Operand.get()->getType();
-    if (getEquivalentType(OperandTy) != nullptr) {
+    if (OperandTy->isOpaquePointerTy()) {
+      OperandTy = clspv::InferType(Operand, U.getContext(), &type_cache_);
+    }
+    if (OperandTy != nullptr && getEquivalentType(OperandTy) != nullptr) {
```
omarahmed1111, line 850: `OperandTy != nullptr` is a guard as InferType could return nullptr like when it cannot guess the opaque pointer type.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -294,24 +275,50 @@ bool vec3ShouldBeLowered(Module &M) {
...
+        }
 
-  return PA;
+        // .. or have a mismatched vector size...
+        auto *VectorType = dyn_cast<FixedVectorType>(ArgInferredType);
+        if (VectorType->getElementCount().getKnownMinValue() != 3) {
+          return true;
+        }
+      }
+    }
```
omarahmed1111, line 317: Do we better have this like that or just checking if the metadata or inferred type is a vector of size 3 would be simpler and would still work?

alan-baker: I'd prefer not to rely on metadata. We want to identify cases where a vec3 buffer is being used as a non-vec3. If InferType returns returns a vec3 that is subsequently treated as a different type that's what we need to catch. So probably we want to catch where the buffers are inferred as vec3, but then there is an implicit cast to a non-vec3 elsewhere in the flow.

For example the following is perfectly ok:
```
kernel void foo(global float3* data) {
  global float* tmp = (global float*)data;
  *data = 0;
}
```
clspv will just treat this as a float buffer, but:
```
kernel void bar(global float3* data) {
  global float3* tmp1 = data[3];
  global float4* tmp2 = (global float4*)tmp1;
  *tmp2 = (float4)(0.0f,0.0f,0.0f,0.0f);
}
```
This won't be caught (in the naive translation at least).

omarahmed1111: Updated, I have used the stores to detect implicit casts.

lib/ThreeElementVectorLoweringPass.cpp
```diff
@@ -294,24 +275,61 @@ bool vec3ShouldBeLowered(Module &M) {
...
 
-  cleanDeadFunctions();
-  cleanDeadGlobals();
-
-  return PA;
+  // This method only works for opaque pointers as it will fail with non opaque
+  // pointer to pointer.
+  for (Instruction &I : instructions(F)) {
+    // If we have a store instruction...
+    if (auto *store = dyn_cast<StoreInst>(&I)) {
```
alan-baker, line 297: What happens with the following testcase:
```
define void @test(ptr addrspace(1) %data) {
entry:
  %gep1 = getelementptr <3 x i32>, ptr addrspace(1) %data, i32 0, i32 2
  %gep2 = getelementptr i32, ptr addrspace(1) %gep1, i32 3
  store i32 0, ptr addrspace(1) %gep2
  ret void
}
```

It seems like this code wouldn't identify this case as problematic, but clspv would want to infer the argument type as `<3 x i32>`. To be safe I think the code needs to handle more cases where  pointer type is inferred. I would think a load case would also be problematic:
```
define i32 @test(ptr addrspace(1) %data) {
entry:
  %gep1 = getelementptr <3 x i32>, ptr addrspace(1) %data, i32 0, i32 2
  %gep2 = getelementptr i32, ptr addrspace(1) %gep1, i32 3
  %ld = load i32, ptr addrspace(1) %gep2
  ret i32 %ld
}
```

Both examples access the "4th" element of three-element vector.

omarahmed1111: I tried to generalize it more by detecting if there is an arg or instruction that was inferred of a type and used as another type. This should handle these cases and the previous store case. store instruction inside `InferType` function changes will look throw the store operands to infer the type if it have only pointer operands so no need to handle it explicitly.

alan-baker (approved):
This looks great! Thanks for the changes.

PR #921, opened 2022-09-02
Add first support for debug information

rjodinchr:
This is part of the effort to integrate clspv into [compiler explorer](https://github.com/compiler-explorer/compiler-explorer)

Screenshot_2022-09-02_15 31 34

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -298,6 +298,7 @@ struct SPIRVProducerPassImpl {
   typedef UniqueVector<StrideType> StrideTypeList;
   typedef DenseMap<Value *, SPIRVID> ValueMapType;
   typedef DenseMap<DIFile *, SPIRVID> DIFileMap;
+  typedef DenseMap<BasicBlock *, std::pair<uint32_t, uint32_t>> BBDILocMap;
   typedef std::list<SPIRVID> SPIRVIDListType;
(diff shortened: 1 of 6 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -4324,10 +4334,26 @@ SPIRVID SPIRVProducerPassImpl::GenerateInstructionFromCall(CallInst *Call) {
 
 void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
   ValueMapType &VMap = getValueMap();
+  DIFileMap &DbgDIFileMap = getDebugDIFileMap();
   LLVMContext &Context = module->getContext();
 
   SPIRVID RID;
 
+  if (auto md = I.getMetadata("dbg")) {
```
alan-baker, line 4342: Ideally we would only generate one OpLine instruction for a chain of instructions that share a location. Can we save the debug info (reset every basic block) and check if it differs?

Separately, I think this needs an else that if we want debug info and there is no info (and we're in a block) we should insert an OpNoLine instruction to prevent the previous OpLine from affecting this instruction.

alan-baker (changes requested):
Please also add a test with some control flow. There some layout requirements in SPIR-V that I think are implicitly ok, but should double check with a test.

alan-baker (approved):

PR #923, opened 2022-09-07
Support variable indices in insertelement/extractelement for LongVectorLoweringPass 

omarahmed1111:
# PR Description
* **Support variable indices in insertelement/extractelement instructions for longVectorLowering pass**: Add 2 cases for lowering vectors inside insertelement/extractelement with variable indices:
  * First case when we have the vector operand instruction as `load`.
  * Second case is the general case where we add `alloca` instruction to extract the array pointer.
* Add tests for the 2 cases.

**Note**: This changes fixes CTS tests: spirv_new/op_vector_char16_extract, op_vector_char16_insert

### DIFF
--- a/lib/LongVectorLoweringPass.cpp
+++ b/lib/LongVectorLoweringPass.cpp
@@ -968,16 +968,33 @@ Value *
 clspv::LongVectorLoweringPass::visitExtractElementInst(ExtractElementInst &I) {
   Value *EquivalentValue = visit(I.getOperand(0));
   assert(EquivalentValue && "value not lowered");
-
   ConstantInt *CI = dyn_cast<ConstantInt>(I.getOperand(1));
-  // Non-constant indices could be done with a gep + load pair.
-  assert(CI && "Dynamic indices not supported yet");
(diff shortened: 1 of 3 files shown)


lib/LongVectorLoweringPass.cpp
```diff
@@ -1100,14 +1115,30 @@ clspv::LongVectorLoweringPass::visitInsertElementInst(InsertElementInst &I) {
...
+    registerReplacement(I, *V);
+    return V;
+  }
+
+  // Extract array pointer
+  Value *PtrOperand = nullptr;
+  if (auto *Load = dyn_cast<LoadInst>(EquivalentValue)) {
+    PtrOperand = Load->getPointerOperand();
+  } else {
+    PtrOperand = B.CreateAlloca(EquivalentValue->getType());
```
alan-baker, line 1133: This always makes me a bit nervous. Is it not a requirement that allocas are at the entry block of the function?

omarahmed1111: Fixed, sorry for that.

PR #925, opened 2022-09-12
Implement printf

callumfare:
Add support for printf. The compiler will generate definitions that store a unique identifier for the given format string and the value of each argument.

A storage buffer of a configurable size is used to provide the printf buffer. Information on the buffer and the individual printf calls is embedded in the reflection instructions.

Also add a few misc fixes to make this work:
* Allow lowering of ConstantExprs assuming there is only 1 user
* Add support for atomic instruction implicit GEPs to ReplacePointerBitcastPass

Related Pull Requests (the SPIRV-Headers change is required):
https://github.com/KhronosGroup/SPIRV-Headers/pull/290
https://github.com/KhronosGroup/SPIRV-Registry/pull/168


lib/Option.cpp
```diff
@@ -355,6 +355,14 @@ static llvm::cl::opt<bool>
     debug_info("g", llvm::cl::init(false),
                llvm::cl::desc("Produce debug information."));
 
+static llvm::cl::opt<bool>
+    printf_support("enable-printf", llvm::cl::desc("Enable support for printf"),
```
callumfare, line 359: Is this an option we can have enabled by default?

alan-baker:
Can you describe the overall design succinctly in one of these pull requests? My understanding is the following:
1. clspv creates a descriptor for a storage buffer for printf and uses index 0 in the buffer as an atomic counter for reservations
2. Every printf call is translated as a reservation of space for an id of the call and the size of arguments. The id is unique to the format string.
3. The format strings are reflected by the PrintInfo reflection instruction. Clvk is expected map these ids and strings per kernel/program to translate the recorded data into actual printf statements on the host.
4. Clvk needs to manage a buffer for the data. It is not clear how many buffers are needed. Maybe just one per queue, but reuse of the buffer might be problematic.

@kpet any thoughts? The clspv side of things seems reasonable if I'm understanding correctly, but the buffer management seems like it might be a pain.

callumfare:
> Can you describe the overall design succinctly in one of these pull requests? My understanding is the following:

I've added some documentation on the design to [docs/OpenCLCOnVulkan.md](https://github.com/google/clspv/pull/925/commits/b5a74d37b3d9b47c953545e4d023713347f9e2b3#diff-47c5a05a43b74b1fed8a92222ebd6c977061cc809bbba7324ed7cec56be6fc23) - I'm happy to add to this if it's still not clear.

Your understanding is basically correct, although the printf ID is unique to the printf call site rather the format string. Multiple printf calls with the same format string (but possibly different arguments) will have unique IDs.

There's some ongoing discussion on the clvk pull request on the best way to handle the buffer.

alan-baker:
Please rebase this and resolve the conflicts.

lib/Option.cpp
alan-baker: Probably in the future yes, but let's leave it off for the time being.

callumfare:
I've rebased and updated this. Unfortunately there's now a regression with the `printf_physical_buffer.cl` lit test - it seems like some recent changes mean that with physical addressing enabled, the printf buffer in the final SPIR-V has the type `i32` instead of `{ [0 x i32] }`, so the index operands don't match. I'm guessing this is related to opaque pointer support.

Unfortunately we no longer have time to work on this, so it would be great if someone else could pick this up. I suspect it would be a fairly quick/small fix for anyone who is more familiar with the recent changes for opaque pointers etc. I believe everything else is still working correctly, and the clvk changes are also ready to go.

lpavank:
Thanks @callumfare. Samsung side will pick it up and work with anyone to get this in.

rjodinchr:
sure, that's ok, I'll leave it to you then

lpavank:
Since, we took it over and/or dont have permissions to codeplaysoftware fork, we went ahead with our own fork PR. And its here.
https://github.com/google/clspv/pull/1117

This Pr (https://github.com/google/clspv/pull/925) is referenced there. So, I think we can close this PR and continue any discussions on the new one.

PR #932, opened 2022-09-29
compiler: allow compilation of several inputs file together

rjodinchr:
This is a first step to fix OpenCL-CTS 'compiler' tests in clvk.

The idea is to use llvm IR as the intermediate language to link
program together.
To do that, this PR is now exposing one new function:
- CompileFromSourcesString: which will allow to take several input
files, compile and link them together into 1 output buffer.

Major changes:
- 'InputFilename' (now 'InputsFilename') allows for multiple inputs.
- New option 'OutputFormat' allows to choose the output format among those 4 possibilities:
  - `ll` Readable LLVM IR
  - `bc` Binary LLVM IR
  - `spv` Vulkan SPIR-V
  - `c` C initializer list of Vulkan SPIR-V
  Also clean now unused options `-emit-ir` and `-mfmt`.
- 'SetCompilerInstanceOptions' (and new function 'ProgramToModule')
expects 'program' to never be empty. The frontend 'Compile' function
is now making sure of it.
- Logs are now accumulated as many files can be parsed in one call to
clspv.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -887,6 +887,7 @@ int CompilePrograms(const std::vector<std::string> &programs,
                     std::vector<uint32_t> *output_buffer,
                     std::string *output_log) {
   std::vector<std::unique_ptr<llvm::Module>> modules;
+  modules.reserve(programs.size());
 
   llvm::LLVMContext context;
   for (auto program : programs) {
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -155,11 +155,16 @@ enum OutputFormat {
 static llvm::cl::opt<enum OutputFormat> OutputFormat(
     "output-format", llvm::cl::desc("Select output format (ll|bc|spv|c)"),
     llvm::cl::init(OutputFormatSPIRV),
-    llvm::cl::values(clEnumValN(OutputFormatSPIRV, "spv", "SPIRV"),
-                     clEnumValN(OutputFormatLLVMIR, "ll", "Readable LLVM IR"),
-                     clEnumValN(OutputFormatLLVMIRBinary, "bc",
-                                "Binary LLVM IR"),
-                     clEnumValN(OutputFormatC, "c", "C initializer list")));
+    llvm::cl::values(
+        clEnumValN(OutputFormatSPIRV, "spv", "Vulkan SPIR-V"),
+        clEnumValN(
+            OutputFormatLLVMIR, "ll",
+            "Readable LLVM IR (Stop compilation before clspv specific passes)"),
+        clEnumValN(
+            OutputFormatLLVMIRBinary, "bc",
+            "Binary LLVM IR (Stop compilation before clspv specific passes)"),
+        clEnumValN(OutputFormatC, "c",
+                   "C initializer list (of Vulkan SPIR-V)")));
 
 namespace {
 struct OpenCLBuiltinMemoryBuffer final : public llvm::MemoryBuffer {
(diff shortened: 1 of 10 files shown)


rjodinchr:
@kpet  kpet/clvk#391

lib/Compiler.cpp
```diff
@@ -852,63 +875,122 @@ int Compile(const llvm::StringRef &input_filename, const std::string &program,
...
+  llvm::LLVMContext context;
+  for (auto program : programs) {
+    int error;
+    modules.push_back(
+        ProgramToModule(context, "source", program, output_log, &error));
+    if (error != 0)
+      return error;
+  }
+  if (modules.size() == 0 || modules.back() == nullptr)
+    return -1;
```
FMarno, line 897: `modules.size() == 0` should only be possible when `programs` is empty, which is likely a misuse of the function so maybe that should be an assert.
If any item in `modules` is `nullptr` and the error value wasn't set to a non-zero value that is also a bug so I think that should be an assert too.

lib/Compiler.cpp
```diff
@@ -852,63 +875,122 @@ int Compile(const llvm::StringRef &input_filename, const std::string &program,
...
+                        std::istreambuf_iterator<char>());
+    return CompileProgram(inputFilename, program, nullptr, nullptr);
+  } else if (InputsFilename.size() == 1) {
+    llvm::StringRef inputFilename = InputsFilename[0];
+    std::ifstream stream(inputFilename.str());
+    std::string program((std::istreambuf_iterator<char>(stream)),
+                        std::istreambuf_iterator<char>());
+    return CompileProgram(inputFilename, program, nullptr, nullptr);
+  } else {
+    std::vector<std::string> programs;
```
FMarno, line 967:
```suggestion
    std::vector<std::string> programs;
    programs.reserve(InputsFilename.size());
```

lib/Compiler.cpp
```diff
@@ -852,63 +875,122 @@ int Compile(const llvm::StringRef &input_filename, const std::string &program,
...
+    for (auto InputFilename : InputsFilename) {
+      std::ifstream stream(InputFilename);
+      std::string program((std::istreambuf_iterator<char>(stream)),
+                          std::istreambuf_iterator<char>());
+      programs.push_back(std::move(program));
```
FMarno, line 972:
```suggestion   
       programs.emplace_back(std::istreambuf_iterator<char>(stream),
                             std::istreambuf_iterator<char>());

```

FMarno (approved):
I'm interested in linking so I thought I would check out your PRs. LGMT :smile:

FMarno:
> Can you please sign the CLA so that this PR could be merged once Alan would have reviewed it. Thanks

Sorted, I had the CLA linked to the wrong account.

alan-baker:

lib/Compiler.cpp
```diff
@@ -153,6 +156,10 @@ static llvm::cl::opt<std::string> IROutputFile(
         "Emit LLVM IR to the given file after parsing and stop compilation."),
     llvm::cl::value_desc("filename"));
 
+static llvm::cl::opt<bool>
```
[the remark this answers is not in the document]
rjodinchr: I agree.
I have also added `c` as an output format.
And I have removed `emit-ir` and `mfmt` as they are overlapping with `output-format`. I am not sure of that as it breaks `clspv` backward compatibility.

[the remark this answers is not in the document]
alan-baker: Strictly speaking it does, but the transition should be straightforward.

alan-baker (approved):

PR #933, opened 2022-09-29
Defined get_global_linear_id and get_local_linear_id

FMarno:
Here I've added definitions for the OpenCL C 2.0+ builtins get_global_linear_id and get_local_linear_id.

### DIFF
--- a/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
+++ b/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
@@ -28,44 +28,23 @@
 using namespace llvm;
 using namespace clspv;
 
-PreservedAnalyses
-DefineOpenCLWorkItemBuiltinsPass::run(Module &M, ModuleAnalysisManager &) {
-  PreservedAnalyses PA;
-
-  defineGlobalOffsetBuiltin(M);
-  defineGlobalIDBuiltin(M);
-
-  defineMappedBuiltin(M, "_Z14get_local_sizej",
-                      clspv::WorkgroupSizeVariableName(), 1,
-                      clspv::WorkgroupSizeAddressSpace());
-  defineMappedBuiltin(M, "_Z12get_local_idj",
-                      clspv::LocalInvocationIdVariableName(), 0,
-                      clspv::LocalInvocationIdAddressSpace());
-  defineNumGroupsBuiltin(M);
-  defineGroupIDBuiltin(M);
-  defineGlobalSizeBuiltin(M);
-  defineWorkDimBuiltin(M);
-  defineEnqueuedLocalSizeBuiltin(M);
-  defineMaxSubGroupSizeBuiltin(M);
-  defineEnqueuedNumSubGroupsBuiltin(M);
-  addWorkgroupSizeIfRequired(M);
-
-  return PA;
-}
-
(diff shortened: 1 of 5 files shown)


lib/DefineOpenCLWorkItemBuiltinsPass.cpp
```diff
@@ -28,19 +28,77 @@
...
+constexpr auto local_linear_id_mangled_name = "_Z19get_local_linear_idv";
+constexpr auto local_size_mangled_name = "_Z14get_local_sizej";
+constexpr auto max_sub_group_size_mangled_name = "_Z22get_max_sub_group_sizev";
+constexpr auto num_groups_mangled_name = "_Z14get_num_groupsj";
+constexpr auto work_dim_mangled_name = "_Z12get_work_dimv";
+
+Value *inBoundsDimensionCondition(IRBuilder<> &Builder, Value *Dim) {
+  // Vulkan has 3 dimensions for work-items, but the OpenCL API is written
+  // such that it could have more. We have to check whether the value provided
+  // was less than 3...
```
jlewis-austin, line 51: Do you have a reference for this? I wasn't aware of OpenCL supporting more than 3 dimensions

FMarno: This code was copy-and-pasted from below but I found this reference.
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS

rjodinchr: Interesting! I'm wondering how many OpenCL implementations support more than 3 dimensions 🤔 Not much I guess.
Thank you @FMarno

FMarno: I would also guess not many! You're welcome :smile:

jlewis-austin: Thanks! I was looking at the description of NDRange: _An NDRange is an N-dimensional index space, where N is one, two or three_, and the C functions that are defined for any value of dim: _Valid values of dimindx are 0 to get_work_dim() - 1. For other values of dimindx, get_global_id() returns 0._
I wonder if it's an oversight where they just neglected to mention a max value of CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS? Either way, I guess the check is harmless

FMarno: It could potentially be an out of date description? This webpage for OpenCL API 1.0 also mentions a max dimension of 3 https://registry.khronos.org/OpenCL/sdk/1.0/docs/man/xhtml/clEnqueueNDRangeKernel.html

lpavank: Actually I am wondering if PyOpenCL (with nparray) or something along those lines, are one of those "custom" devices to support more than 3 dimensions.  https://documen.tician.de/pyopencl/array.html#
TBH, I have not seen it, but want to throw it out there.

alan-baker (approved):

alan-baker:

PR #936, opened 2022-10-12
Add support for spir64 target

callumfare:
Add support for `spir64-unknown-unknown` target
- Add a copy of the builtins library targeting `clspv64--`
- Add `-spir64` boolean option to choose between targets. This will select the relevant builtins library and Clang frontend options.
- Various pass changes necessary to generate correct SPIR-V
- Add lit testing infrastructure for multiple targets, `check-spirv` now sets `-spir64=0` and `check-spirv-64` sets `-spir64=1`.
- Update some (approx. 80) lit tests to have different Filecheck runs for `spir` and `spir64` where the output SPIR-V that needs checked is different

This is a prerequisite for physical addressing (#826)


callumfare:
The first commit contains all the 'real' changes including changes to lit tests that have differing output for the spir and spir64 targets. The second commit contains the changes that add the 'check-spirv-64' target and edits each lit test to add the new `%target` variable, hence the giant diff.

alan-baker:

alan-baker:
I assume this needs updated to account for #933 being merged.

callumfare:
> I assume this needs updated to account for #933 being merged.

Yep, I've just rebased this on those changes and updated the new linear builtins

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -990,6 +992,15 @@ Value *ReplaceOpenCLBuiltinPass::replaceAsyncWorkGroupCopies(
...
   auto LocalSize2 = Builder.CreateExtractElement(LocalSize, Cst2);
 
+  if (clspv::PointersAre64Bit(M)) {
+    LocalId0 = Builder.CreateZExt(LocalId0, Builder.getInt64Ty());
+    LocalId1 = Builder.CreateZExt(LocalId1, Builder.getInt64Ty());
+    LocalId2 = Builder.CreateZExt(LocalId2, Builder.getInt64Ty());
+    LocalSize0 = Builder.CreateZExt(LocalSize0, Builder.getInt64Ty());
+    LocalSize1 = Builder.CreateZExt(LocalSize1, Builder.getInt64Ty());
+    LocalSize2 = Builder.CreateZExt(LocalSize2, Builder.getInt64Ty());
+  }
```
alan-baker, line 1002: I'm a bit confused as to why these are necessary.

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -2028,6 +2039,10 @@ bool ReplaceOpenCLBuiltinPass::replaceVstore(Function &F) {
     // Avoid pointer casts. Instead generate the correct number of stores
     // and rely on drivers to coalesce appropriately.
     IRBuilder<> builder(CI);
+    if (clspv::PointersAre64Bit(*F.getParent())) {
```
alan-baker, line 2042: It seems like some of the functions are treating indices as 64-bit and some as 32-bit. Why aren't the consistently based on the size_t arguments? Here, for example, the offset parameter could be examined. If the primitive size is 64-bits then use 64-bits throughout.

lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -275,8 +279,19 @@ bool clspv::SimplifyPointerBitcastPass::runOnGEPFromGEP(Module &M) const {
     SmallVector<Value *, 8> Idxs;
 
     Value *SrcLastIdxOp = OtherGEP->getOperand(OtherGEP->getNumOperands() - 1);
-
     Value *GEPIdxOp = GEP->getOperand(1);
+
+    if (clspv::PointersAre64Bit(M) && OtherGEP->getType()->isStructTy()) {
```
alan-baker, line 284: Another instance I wonder why we aren't just working in the index type throughout. Am I missing an assumption why we would never overflow?

alan-baker (changes requested):

lib/Compiler.cpp
```diff
@@ -154,6 +156,10 @@ static llvm::cl::opt<std::string> IROutputFile(
         "Emit LLVM IR to the given file after parsing and stop compilation."),
     llvm::cl::value_desc("filename"));
 
+static llvm::cl::opt<bool>
```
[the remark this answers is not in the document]
callumfare: Agreed it looks better, I've updated this

lib/ReplaceOpenCLBuiltinPass.cpp
callumfare: Here (and other places) the local id and local size are loaded directly from the global variables that contain them (`__spirv_WorkgroupSize` etc). Since these variables end up with the `WorkgroupSize` etc builtin decorations, they're always i32 even when targeting spir64. Generally I've tried to promote these to the index type as early as possible to avoid losing precision.

This could be replaced with calls to the builtins directly but after inlining the ZExts will be there anyway.

lib/ReplaceOpenCLBuiltinPass.cpp
callumfare: Generally I've tried to keep them all treating the indices as 64-bit where possible, but it does seem like there are a few inconsistent ones so I'll revisit those. `vstore_half` is a weird one because in the path where it uses an atomic_xor it uses the offset to create a mask for an i32 value, but I think it can be changed to avoid truncation

lib/SimplifyPointerBitcastPass.cpp
callumfare: We don't control all the GEPs coming into the pass, so quite frequently there are a mix of i32 and i64 indices. Even with the spir64 target, small constant indices are usually i32. So here the pass will:

- Promote all i32 indices to i64. This means we can do arithmetic on the different indices in a GEP without the types mismatching.
- At the end, demote i64 indices back down i32 for struct member indices, since LLVM requires this. There is a check for constant values that they still fit in an i32, but in theory this is only a problem for structs that have more than UINT_MAX members.

I'm reasonably confident that this is ok, but definitely open to other ideas of what to do here.

callumfare: Thinking about this a bit more, the type being a StructTy doesn't guarantee that the current GEP index operand being looked at is the actual index to a struct member, it could be an index into a buffer of structs, which might then overflow when truncated. I think this needs to be more precise to only truncate certain indices when it find a struct, rather than all of them.

alan-baker:
Not sure why some of the bots are aborting, but there seem to be some lit test failures.

lib/DefineOpenCLWorkItemBuiltinsPass.cpp
```diff
@@ -122,6 +123,11 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineMappedBuiltin(
 
   Value *Select2 =
       inBoundsDimensionOrDefaultValue(Builder, Dim, Result, DefaultValue);
+
+  if (clspv::PointersAre64Bit(M)) {
```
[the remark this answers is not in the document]
callumfare: This has been tidied up now, there was a mistake here which was causing the lit test failure, not sure why I didn't see that locally.

lib/ReplaceOpenCLBuiltinPass.cpp
callumfare: I've updated these so it should be consistently using size_t now

lib/SimplifyPointerBitcastPass.cpp
callumfare: It turns out this entire block of code was redundant. I've tidied it up now so the only new changes are:

- Extending indices to i64 when needed to make arithmetic work
- Truncating stray i64 struct member indices back down to i32, but only when we actually know the constant value will fit

alan-baker:
Debug builds are hitting an assertion in the get_global_linear_id test.
```
clspv: ../third_party/llvm/llvm/lib/IR/Instructions.cpp:2850: static llvm::BinaryOperator *llvm::BinaryOperator::Create(llvm::Instruction::BinaryOps, llvm::Value *, llvm::Value *, const llvm::Twine &, llvm::Instruction *): Assertion `S1->getType() == S2->getType() && "Cannot create binary operator with two operands of differing type!"' failed.
```

alan-baker (approved):

rjodinchr:
It feels that we are not running the `check-spirv-64` target in our test.
When I run it locally I get:
```
Failed Tests (2):
  clspv :: PointerCasts/deref_load_cast_float2_to_int4.cl
  clspv :: PointerCasts/deref_store_cast_int4_to_float2.cl


Testing Time: 52.17s
  Passed           : 3264
  Expectedly Failed:    1
  Failed           :    2
```

callumfare:
> It feels that we are not running the `check-spirv-64` target in our tests. When I run it locally I get:
> 
> ```
> Failed Tests (2):
>   clspv :: PointerCasts/deref_load_cast_float2_to_int4.cl
>   clspv :: PointerCasts/deref_store_cast_int4_to_float2.cl
> 
> 
> Testing Time: 52.17s
>   Passed           : 3264
>   Expectedly Failed:    1
>   Failed           :    2
> ```

I'm able to reproduce these fails after doing a clean build, I think I might have been on an older LLVM commit and didn't see these locally. I'll take a look at these failures next week. It would be good to get the new tests running in CI

PR #937, opened 2022-10-12
fix clvk CI by compiling in release with assertions ON

rjodinchr:
let clvk test in its CI the release modes


kokoro/clvk-linux-clang-release/presubmit.cfg
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-build_file: "clspv/kokoro/clvk-linux-clang-debug/build.sh"
+build_file: "clspv/kokoro/clvk-linux-clang-release/build.sh"
```
alan-baker, line 15: Same here.

alan-baker (changes requested):
Please revert the build file changes. That should be sufficient. To change bot names requires some associated changes on internal repos. If necessary that can be a separate step.

alan-baker (approved):

PR #941, opened 2022-10-24
Added command line option to specify CL3.0 feature macros

FMarno:
This adds the ability to specify OpenCL3.0 feature macros that should be enabled by the clang frontend.

One thing I'm not sure about in this PR is the overlap of some options, specifically FP64 and ImageSupport. I think it would be best for the program to error out when one of these is disabled but the feature macro is enabled, but I'm not sure how CLSPV usually manages that.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -741,16 +741,14 @@ int ParseOptions(const int argc, const char *const argv[]) {
 
   const auto enabled_feature_macros = clspv::Option::EnabledFeatureMacros();
   if (!clspv::Option::FP64() &&
-      enabled_feature_macros.find(clspv::FeatureMacro::__opencl_c_fp64) !=
-          enabled_feature_macros.end()) {
+      enabled_feature_macros.count(clspv::FeatureMacro::__opencl_c_fp64)) {
     llvm::errs() << "error: Cannot enabled feature macro __opencl_c_fp64 while "
                     "-fp64 is disabled!\n";
     return -1;
   }
 
   if (!clspv::Option::ImageSupport() &&
-      enabled_feature_macros.find(clspv::FeatureMacro::__opencl_c_images) !=
-          enabled_feature_macros.end()) {
+      enabled_feature_macros.count(clspv::FeatureMacro::__opencl_c_images)) {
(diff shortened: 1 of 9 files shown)


alan-baker:
Take a look at ParseOptions in lib/Compiler.cpp to see how we handle incompatible options specifications.

I think longer term we may want to deprecate (and remove) some options if we want to specify feature macros.

FMarno:
Thanks @alan-baker, that worked well for me.

FMarno:
I think I've fixed the MSVC compile error, unfortunately I don't have a machine to test it on.

lib/Compiler.cpp
```diff
@@ -738,6 +739,24 @@ int ParseOptions(const int argc, const char *const argv[]) {
     return -1;
   }
 
+  const auto enabled_feature_macros = clspv::Option::EnabledFeatureMacros();
+  if (not clspv::Option::FP64() &&
+      enabled_feature_macros.find(clspv::FeatureMacro::__opencl_c_fp64) !=
+          enabled_feature_macros.end()) {
```
alan-baker, line 745:
```suggestion
  if (!clspv::Option::FP64() &&
      enabled_feature_macros.count(clspv::FeatureMacro::__opencl_c_fp64)) {
```

lib/Compiler.cpp
```diff
@@ -738,6 +739,24 @@ int ParseOptions(const int argc, const char *const argv[]) {
...
+    llvm::errs() << "error: Cannot enabled feature macro __opencl_c_fp64 while "
+                    "-fp64 is disabled!\n";
+    return -1;
+  }
+
+  if (not clspv::Option::ImageSupport() &&
+      enabled_feature_macros.find(clspv::FeatureMacro::__opencl_c_images) !=
+          enabled_feature_macros.end()) {
```
alan-baker, line 753:
```suggestion
  if (!clspv::Option::ImageSupport() &&
      enabled_feature_macros.count(clspv::FeatureMacro::__opencl_c_images)) {
```

alan-baker (changes requested):
Just nits.

alan-baker (approved):

PR #943, opened 2022-10-31
fix reading binary input + add test

rjodinchr:


rjodinchr:
This is blocking the update of clspv in clvk: 
https://github.com/kpet/clvk/pull/447
https://github.com/kpet/clvk/pull/429

alan-baker (approved):

PR #946, opened 2022-11-04
Updates and fixes

alan-baker:
* Update LLVM
  * switch uses of removed function attributes to memory effects
* Fix tests broken by #936
  * fixes sizes in replace pointer bitcasts
* Update bots to run check-spirv-64


alan-baker:
CC @callumfare

rjodinchr (approved):

callumfare:
@alan-baker Thanks for fixing the broken tests! LGTM

Issue #945, opened 2022-11-03
AMD devices require decorating NonUniform objects


rjodinchr:
At the moment clspv is not decorating any object with the non uniform decorator.

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Uniformity

On AMD devices, it has an impact, producing wrong behavior on some tests of the CTS. Especially around Load/Store.

I have made a PoC by adding the non uniform decorator on all pointers going into a Load or Store, which fixes the tests. But this is a bit too much, and I think we will want to have a more precise detection of which object we need to decorate.

PR #951, opened 2022-11-08
add option to decorate nonuniform pointers

rjodinchr:
Look for possible decoration for every pointers going to a load or a store.
Visit the origin of the pointer by going through phis, selects and geps.

Ref #945

### DIFF
--- a/lib/Option.cpp
+++ b/lib/Option.cpp
@@ -385,6 +385,11 @@ static llvm::cl::opt<bool>
     debug_info("g", llvm::cl::init(false),
                llvm::cl::desc("Produce debug information."));
 
+static llvm::cl::opt<bool> decorate_non_uniform(
+    "decorate-nonuniform", llvm::cl::init(false),
+    llvm::cl::desc(
+        "Decorate NonUniform Pointers with the NonUniform decoration."));
+
 } // namespace
 
 namespace clspv {
@@ -494,6 +499,8 @@ bool OpaquePointers() { return opaque_pointers; }
 bool DebugInfo() { return debug_info; }
 
 std::set<FeatureMacro> EnabledFeatureMacros() { return enabled_feature_macros; }
+
+bool DecorateNonUniform() { return decorate_non_uniform; }
 
 } // namespace Option
 } // namespace clspv
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -318,8 +318,9 @@ struct SPIRVProducerPassImpl {
         binaryTempOut(binaryTempUnderlyingVector), binaryOut(out),
         outputCInitList(outputCInitList), patchBoundOffset(0), nextID(1),
         OpExtInstImportID(0), HasVariablePointersStorageBuffer(false),
-        HasVariablePointers(false), SamplerPointerTy(nullptr),
-        SamplerDataTy(nullptr), WorkgroupSizeValueID(0), WorkgroupSizeVarID(0),
+        HasVariablePointers(false), HasNonUniformPointers(false),
+        SamplerPointerTy(nullptr), SamplerDataTy(nullptr),
+        WorkgroupSizeValueID(0), WorkgroupSizeVarID(0),
         TestOutput(out == nullptr) {
     addCapability(spv::CapabilityShader);
     Ptr = this;
@@ -330,9 +331,9 @@ struct SPIRVProducerPassImpl {
         binaryTempOut(binaryTempUnderlyingVector), binaryOut(nullptr),
         outputCInitList(false), patchBoundOffset(0), nextID(1),
         OpExtInstImportID(0), HasVariablePointersStorageBuffer(false),
-        HasVariablePointers(false), SamplerPointerTy(nullptr),
-        SamplerDataTy(nullptr), WorkgroupSizeValueID(0), WorkgroupSizeVarID(0),
-        TestOutput(true) {
+        HasVariablePointers(false), HasNonUniformPointers(false),
+        SamplerPointerTy(nullptr), SamplerDataTy(nullptr),
+        WorkgroupSizeValueID(0), WorkgroupSizeVarID(0), TestOutput(true) {
     addCapability(spv::CapabilityShader);
     Ptr = this;
   }
@@ -375,6 +376,13 @@ struct SPIRVProducerPassImpl {
     if (!HasVariablePointers) {
       addCapability(spv::CapabilityVariablePointers);
       HasVariablePointers = true;
+    }
+  }
+  bool hasNonUniformPointers() { return HasNonUniformPointers; }
+  void setNonUniformPointers() {
+    if (!HasNonUniformPointers) {
+      addCapability(spv::CapabilityShaderNonUniform);
+      HasNonUniformPointers = true;
     }
   }
   GlobalConstFuncMapType &getGlobalConstFuncTypeMap() {
@@ -427,6 +435,8 @@ struct SPIRVProducerPassImpl {
   SPIRVID getSPIRVValue(Value *V);
 
   bool PointerRequiresLayout(unsigned aspace);
+
+  void DecoratePtrIfNonUniform(Value *ptr);
 
   SPIRVID getSPIRVBuiltin(spv::BuiltIn BID, spv::Capability Cap);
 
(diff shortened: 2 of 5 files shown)


lib/Compiler.cpp
```diff
@@ -757,6 +757,13 @@ int ParseOptions(const int argc, const char *const argv[]) {
     return -1;
   }
 
+  if (clspv::Option::DecorateNonUniform() &&
+      clspv::Option::SpvVersion() < clspv::Option::SPIRVVersion::SPIRV_1_5) {
+    llvm::errs()
+        << "error: NonUniform decoration is missing before SPIR-V version 1.5.";
+    return -1;
+  }
```
alan-baker, line 765: The decoration is available from SPV_EXT_descriptor_indexing in earlier versions.

lib/SPIRVProducerPass.cpp
```diff
@@ -3375,6 +3448,15 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVBuiltin(spv::BuiltIn BID,
   return RID;
 }
 
+void SPIRVProducerPassImpl::DecoratePtrIfNonUniform(Value *ptr) {
+  SPIRVOperandVec Ops;
+  if (clspv::Option::DecorateNonUniform() && !isPointerUniform(ptr)) {
+    addCapability(spv::CapabilityShaderNonUniform);
```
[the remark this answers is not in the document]
alan-baker: This should also add the SPV_EXT_descriptor_indexing extension based on the version.

alan-baker:
You'll need to rebase this. We removed gcc7 from the kokoro VM image and updated the scripts to use gcc9 instead.

lib/Compiler.cpp
rjodinchr: Does it mean that we should enable it by default always, whatever the version. Users could disable it explicitly with a flag.
So if someone were to try to run a kernel requiring this features with an old version that does not support the extension, they would "just" have to use the flag and then hope for it to work as expected?
I would be completely ok with that, just want to make sure I understand.

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
rjodinchr: About the assertion, I have also seen the `Input` address space.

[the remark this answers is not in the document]
alan-baker: That would result in invalid SPIR-V being generated. It is something we'd have to prevent.

lib/Option.cpp
```diff
@@ -379,6 +379,11 @@ static llvm::cl::opt<bool>
     debug_info("g", llvm::cl::init(false),
                llvm::cl::desc("Produce debug information."));
 
+static llvm::cl::opt<bool> decorate_non_uniform(
+    "decorate-nonuniform", llvm::cl::init(false),
+    llvm::cl::desc(
+        "Decorate NonUniform Pointers with the NonUniform decoration."));
+
```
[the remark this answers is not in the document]
alan-baker: Sorry to flip flop on this, but until the internal spec issue is resolved let's leave it off by default.

lib/Compiler.cpp
alan-baker: Clvk would have to check to see if it's supported, but it could reach more implementations this way. The alternative if they need the decoration and it's not supported is the kernel not working. So catching it as unsupported it probably better.

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
rjodinchr: That's an issue, this is typically what is happening in the CTS.

test/nonuniform.cl
```diff
@@ -0,0 +1,23 @@
+// RUN: clspv %s -o %t.spv --spv-version=1.5 --decorate-nonuniform
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: spirv-val --target-env vulkan1.2spv1.5 %t.spv
+
+
+__kernel void test(__global char2 *srcA, __global char2 *srcB, __global char *srcC, __global char2 *dst)
```
alan-baker, line 7: This seems to fail a couple debug builds. Is the IR consumed by the producer the same in those cases?

I'd suggest turning this in a couple .ll tests that just run the SPIR-V producer.

lib/SPIRVProducerPass.cpp
```diff
@@ -377,6 +377,13 @@ struct SPIRVProducerPassImpl {
       HasVariablePointers = true;
     }
   }
+  bool hasNonUniformPointers() { return HasNonUniformPointers; }
+  void setNonUniformPointers() {
+    if (HasNonUniformPointers) {
+      addCapability(spv::CapabilityShaderNonUniform);
+      HasNonUniformPointers = true;
+    }
+  }
```
alan-baker, line 386: This looks off. The control is set to true only if it is already true. Also `HasNonUniformPointers` doesn't seem to get initialized by the constructors.

lib/SPIRVProducerPass.cpp
```diff
@@ -5292,6 +5321,9 @@ void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
       Ops << ST->getValueOperand();
     }
     RID = addSPIRVInst(spv::OpStore, Ops);
+
+    DecoratePtrIfNonUniform(ptr);
```
alan-baker, line 5325: I think this needs to be added for any memory operation (e.g. atomics, OpCopyMemory, etc).

alan-baker: Maybe it's possible to refactor to the end of GenerateInstruction.

lib/SPIRVProducerPass.cpp
```diff
@@ -6192,6 +6224,22 @@ bool SPIRVProducerPassImpl::selectFromSameObject(Instruction *inst) {
   return false;
 }
 
+bool SPIRVProducerPassImpl::isPointerUniform(Value *ptr) {
```
alan-baker, line 6227: I think there should be an address space check here still. We don't need to decorate workgroup pointers for example and cannot decorate input or uniform constant.

alan-baker (changes requested):

lib/SPIRVProducerPass.cpp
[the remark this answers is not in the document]
alan-baker: Let's file another issue for those cases (especially please list the CTS test that requires it).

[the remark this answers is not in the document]
rjodinchr: I made a mistake. No `Input` address space is needed to be decorated to fix the issue I am seeing on AMD devices.

lib/SPIRVProducerPass.cpp
rjodinchr: Fixed

lib/SPIRVProducerPass.cpp
rjodinchr: I have refactored it to work on the operands of each instruction at the beginning of GenerateInstruction.

lib/SPIRVProducerPass.cpp
rjodinchr: Done

lib/SPIRVProducerPass.cpp
```diff
@@ -6192,6 +6225,35 @@ bool SPIRVProducerPassImpl::selectFromSameObject(Instruction *inst) {
...
+  if (!ptr->getType()->isPointerTy()) {
+    return true;
+  }
+  if (GetStorageClass(ptr->getType()->getPointerAddressSpace()) !=
+      spv::StorageClassStorageBuffer) {
+    return true;
+  }
+
+  static std::set<Value *> nonUniformPointer;
```
alan-baker, line 6237: I didn't notice this before, but I'd prefer not using static here. Instead just move this into the SPIRVProducerPassImpl and clear at the beginning of the pass.

alan-baker (changes requested):
Nearly there, thanks for your patience.

lib/SPIRVProducerPass.cpp
rjodinchr: Done.
Just out of curiosity, why do you prefer not to use static variables in this case?

alan-baker: I think it makes it harder to understand the state of the overall pass.

alan-baker (approved):

PR #954, opened 2022-11-10
Add support for physical addressing of global memory using PhysicalStorageBuffers

callumfare:
This change adds an option which will use PhysicalStorageBuffers instead of regular StorageBuffers to represent global memory. Instead of turning into a resource function, global pointer arguments are passed as integer POD types which are then bitcast to pointers. These 'pointer POD' types are treated as regular PODs, and take part in POD clustering etc. This means they can be passed as push constants or in a uniform buffer.

This uses the existing `ArgumentPointerPushConstant` and `ArgumentPointerUniform` reflection instructions, which were previously unused.

Note that with the new option enabled, `ReplacePointerBitcasts` etc don't need to run on pointers in the global address space, but by default they still will as simplifying the casts helps avoid issues we've seen with multiple drivers.

It's also possible to use PhysicalStorageBuffers for the constant address space, but this will be done as a separate PR as it requires changes to the ClspvReflection extension.


callumfare:
The corresponding clvk change: https://github.com/kpet/clvk/pull/453

alan-baker (commented):
Haven't looked at everything, but why not support the 32-bit version of physical storage buffers based on the SPIR target? I think we'd simply need either a special rule to collapse bitcasts or just use a spirv.op for the special uint2 -> ptr bitcast.

callumfare:
> Haven't looked at everything, but why not support the 32-bit version of physical storage buffers based on the SPIR target? I think we'd simply need either a special rule to collapse bitcasts or just use a spirv.op for the special uint2 -> ptr bitcast.

I might be missing something, but I didn't think there was a 32-bit version of physical storage buffers. The only addressing mode the extension adds is `PhysicalStorageBuffer64` where the physical pointers are 64-bit. We could represent pointers with uint2 instead of long but I think the spir64 target is still needed to actually get pointers of that size.

alan-baker:
> > Haven't looked at everything, but why not support the 32-bit version of physical storage buffers based on the SPIR target? I think we'd simply need either a special rule to collapse bitcasts or just use a spirv.op for the special uint2 -> ptr bitcast.
> 
> I might be missing something, but I didn't think there was a 32-bit version of physical storage buffers. The only addressing model the extension adds is `PhysicalStorageBuffer64` where the physical pointers are 64-bit. We could represent pointers with uint2 instead of long but I think the spir64 target is still needed to actually get pointers of that size.

Yeah sorry, maybe I was misrepresenting. Does this handle the case that Int64 is not supported on the device?

callumfare:
> 
> Yeah sorry, maybe I was misrepresenting. Does this handle the case that Int64 is not supported on the device?

It doesn't explicitly handle it, I haven't tried it but I'd be surprised if it worked. I'm not actually sure if the spir64 target itself works in that case. 

I could look into handling it, but we don't need it for our use case so it would be useful for us if we could get this in without support for devices that lack Int64 and maybe come back to that at a later point.

alan-baker:
> It doesn't explicitly handle it, I haven't tried it but I'd be surprised if it worked. I'm not actually sure if the spir64 target itself works in that case.
> 
> I could look into handling it, but we don't need it for our use case so it would be useful for us if we could get this in without support for devices that lack Int64 and maybe come back to that at a later point.

Filing an issue would be enough. My memory of the standardization process was that this was import for several vendors. It would be good long term to support it.

callumfare:
I've created #956 to cover that. Also tagging #826 so this PR is visible there.

alan-baker:
This will need rebased to properly test the gcc bots.

alan-baker:
Do you mind rebasing and adding a test that the non-uniform decorations are not added to physical storage buffers? I think they won't be, but it's good to have the coverage.

rjodinchr:
Beware that the feature is disabled by default. It will need to be enabled with the `-decorate-nonuniform` option to make sure it is not used when enabled.

alan-baker:
> @alan-baker @rjodinchr I've added a test which is passing. Out of interest, I couldn't find wording in the spec that says that non-uniform physical storage buffers don't get the NonUniform decoration, is it explicitly mentioned anywhere?

The requirement is about accessing descriptors so it doesn't seem applicable here. It probably wouldn't hurt, but it seems more like noise. The addressing model is different at that point so a bunch of other considerations need to be made. I'd rather start from here and see if there are problems.

callumfare:
> 
> The requirement is about accessing descriptors so it doesn't seem applicable here. It probably wouldn't hurt, but it seems more like noise. The addressing model is different at that point so a bunch of other considerations need to be made. I'd rather start from here and see if there are problems.

Ok, that makes sense, thanks!

alan-baker (approved):

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -34,6 +35,16 @@ using namespace BitcastUtils;
 
 #define DEBUG_TYPE "replacepointerbitcast"
 
+namespace {
+
+// TODO: It should be safe to leave these bitcasts as-is, but replacing them
+// anyway helps work around Vulkan driver bugs
+cl::opt<bool> ReplacePhysicalPointerBitcasts(
```
dneto0, line 42: We should file a Vulkan CTS bug to cover this.

PR #958, opened 2022-11-23
Hack preprocessor to allow disabling all feature macros

FMarno:

This will allow CLSPV to turn off all the feature macros.

Due to opencl-c-base.h in the clang project [always defining](https://github.com/llvm/llvm-project/blob/1215e86a0e4703d4c6657564e1a758d95ea59c56/clang/lib/Headers/opencl-c-base.h#L68) certain preprocessor macros, I've added an `#undefine` for `__SPIR__`  and `__SPIRV__`. These defines should not be necessary since the [`BaseSPIRTargetInfo`](https://github.com/llvm/llvm-project/blob/1215e86a0e4703d4c6657564e1a758d95ea59c56/clang/lib/Basic/Targets/SPIR.h#L161) already defines all these feature macros, but I think it's not been fixed properly since a few are missing from [OpenCLExtensions.def](https://github.com/llvm/llvm-project/blob/1215e86a0e4703d4c6657564e1a758d95ea59c56/clang/include/clang/Basic/OpenCLExtensions.def#L119). Due to the missing feature macros in OpenCLExtension.def, I've also had to add those symbols in as a special case.

I've also created a PR to fix this upstream but it's not got any interest yet https://reviews.llvm.org/D137652.

### DIFF
--- a/include/clspv/FeatureMacro.h
+++ b/include/clspv/FeatureMacro.h
@@ -16,8 +16,8 @@
 #define CLSPV_LIB_FEATUREMACRO_H
 
 #include <array>
-#include <utility>
 #include <string>
+#include <utility>
 
 namespace clspv {
 
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -51,8 +51,8 @@
 #include "clspv/Option.h"
 #include "clspv/Passes.h"
 #include "clspv/Sampler.h"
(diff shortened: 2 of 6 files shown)


alan-baker (approved):

rjodinchr:
Both failures are related to the infra network, no need to worry on your side I believe.

PR #963, opened 2022-11-27
Bump version of non-semantic instruction set to 4

kpet:
We're already using instructions from version 3.

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -14,7 +14,7 @@
       "subrepo" : "KhronosGroup/SPIRV-Headers",
       "branch" : "master",
       "subdir" : "third_party/SPIRV-Headers",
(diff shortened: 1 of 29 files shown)


alan-baker (approved):

alan-baker:
I think you might need to bump the deps for SPIRV-Tools and maybe SPIRV-Headers.

kpet:
@alan-baker I've updated them both to the versions tracked by clvk that I've used for testing.

alan-baker:
The amber tests are failing when trying to strip the non-semantic info instructions.

alan-baker:
Finally got around to looking at this. Amber's deps were out of date so the validator was unhappy with the reflection version.

PR #967, opened 2022-11-29
fix decorate non uniform

rjodinchr:
PHI node's operands may have not been process yet by `GenerateInstruction`, which can lead to `GetSPIRVValue` to fail when trying to decorate an operand that has not been processed.


alan-baker (approved):
I think this is ok because the spec requires that the input to the memory instruction is the annotated instruction, so the operands of the phi shouldn't need to be annotated, but the phi itself might.

PR #971, opened 2022-12-06
Update LLVM

alan-baker:
* Codegen changed to generate shufflevector of <4 x i8>
* SPIRVProducer was producing invalid code in this case
  * bitcasting i32 to i32
  * vector shuffle on i32
* Fix code and test expectations


dneto0 (approved):
Thanks!

PR #974, opened 2022-12-09
Support PhysicalStorageBuffers for constant address space, including module-scope constants

callumfare:
This depends on an update to the ClspvReflection instructions, to allow the device address for module-scope constants to be passed:
https://github.com/KhronosGroup/SPIRV-Headers/pull/308
https://github.com/KhronosGroup/SPIRV-Registry/pull/178

This also fixes a problem with ClusterPodKernelArgumentsPass, where it used the size of the global push constant struct so far as the offset for the kernel args, but this wasn't taking into account padding elements. Now it gets the offset from the struct type directly.

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -14,7 +14,7 @@
       "subrepo" : "KhronosGroup/SPIRV-Headers",
       "branch" : "master",
       "subdir" : "third_party/SPIRV-Headers",
-      "commit" : "c214f6f2d1a7253bb0e9f195c2dc5b0659dc99ef"
+      "commit" : "d13b52222c39a7e9a401b44646f0ca3a640fbd47"
     },
     {
       "name" : "SPIRV-Tools",
--- a/include/clspv/PushConstant.h
+++ b/include/clspv/PushConstant.h
@@ -27,6 +27,7 @@ enum class PushConstant : int {
   RegionGroupOffset,
   KernelArgument,
   ImageMetadata,
+  ModuleConstantsPointer,
 };
 
 enum class ImageMetadata : int {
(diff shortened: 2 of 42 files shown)


alan-baker (approved):

alan-baker:
Can you update the deps to the new headers?

alan-baker:
> Can you update the deps to the new headers?

Also, I had to make https://github.com/KhronosGroup/SPIRV-Headers/pull/312 to fix the headers to support older versions of the instruction set so please ensure you update to include it.

alan-baker (changes requested):
Very close, but after seeing this the reflection version updates in a couple PRs now I'd rather pull the specific version checks from the files. Please change the checks for:
>OpExtInstImport "NonSemantic.ClspvReflection.5"
to:
>OpExtInstImport "NonSemantic.ClspvReflection.

I plan to add validation to SPIRV-Tools to catch all the version issues so this can be caught in the tests using the validator.

alan-baker (approved):

callumfare:
@alan-baker 

Thanks! The CI hit an assert in a couple of the tests, I didn't realise I was testing without asserts. I've updated the assert on `ClusterPodKernelArgumentsPass.cpp:91` to check the size is aligned to 8 instead of 16. I think this is safe to do but honestly I'm not 100% sure what the point of this check is, it doesn't really match the assert message, and the type in question gets updated immediately afterwards anyway. Maybe I'm missing some context.

alan-baker:
> @alan-baker
> 
> Thanks! The CI hit an assert in a couple of the tests, I didn't realise I was testing without asserts. I've updated the assert on `ClusterPodKernelArgumentsPass.cpp:91` to check the size is aligned to 8 instead of 16. I think this is safe to do but honestly I'm not 100% sure what the point of this check is, it doesn't really match the assert message, and the type in question gets updated immediately afterwards anyway. Maybe I'm missing some context.

I'm not certain anymore. Probably all the original push constants were vec3 data so the compiler didn't have to worry about alignments, but I'm not certain it does the wrong thing now.

PR #976, opened 2022-12-09
Fix opaque pointer alloca codegen

alan-baker:
Contributes to #816

* Use getSPIRVPointerType in alloca codegen instead of getSPIRVType


rjodinchr (changes requested):
`getSPIRVType` does not support opaque pointers, but it is called by `getSPIRVPointerType`.
Thus, `opaque_image_metadata.ll` is failing.

alan-baker:
That was the test for #975. I've rebased so it should have that fix now too.

rjodinchr (approved):

PR #977, opened 2022-12-12
Update LLVM

vettoreldaniele:
Fix two constructors to accommodate a few changes in the LLVM API. Update test to take into account a different way to construct a value using masks. From what I can tell, the output is still correct.


vettoreldaniele:
The CI failed due to clvk and the SPIRV-to-LLVM translator that also needs changes. I will submit patches for those.

PR #978, opened 2022-12-12
Better handling for opaque pointers in replace bitcasts

alan-baker:
Contributes to #816

* Further generalizes the fix from #973 to only operate on a single implicit bitcast at a time instead of all bitcasts from the same source
  * properly handles both non-casts cases (previous PR) and when there are multiple different types of casts from the same source

### DIFF
--- a/lib/ReplacePointerBitcastPass.cpp
+++ b/lib/ReplacePointerBitcastPass.cpp
@@ -535,8 +535,6 @@ PreservedAnalyses
 clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
   PreservedAnalyses PA;
 
-  errs() << M << "\n";
-
   WeakInstructions ToBeDeleted;
   SmallVector<Instruction *, 16> WorkList;
   SmallVector<User *, 16> UserWorkList;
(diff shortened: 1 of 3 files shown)


alan-baker:
This fixes the roundTrip half CTS test.

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -534,6 +535,8 @@ PreservedAnalyses
 clspv::ReplacePointerBitcastPass::run(Module &M, ModuleAnalysisManager &) {
   PreservedAnalyses PA;
 
+  errs() << M << "\n";
```
rjodinchr, line 538: encapsulate in `LLVM_DEBUG`?

rjodinchr (changes requested):

rjodinchr:
> This fixes the roundTrip half CTS test.

This test is passing in g3 as well as in chromeos. Do you mean it fixes it with the opaque pointers enabled?

alan-baker:
> > This fixes the roundTrip half CTS test.
> 
> This test is passing in g3 as well as in chromeos. Do you mean it fixes it with the opaque pointers enabled?

Yes, it is failing with opaque pointers on main.

lib/ReplacePointerBitcastPass.cpp
alan-baker: I forgot to remove it.

rjodinchr (approved):

PR #987, opened 2022-12-21
Support limited uses of ptrtoint on logical address spaces

callumfare:
This effectively creates a compile-time virtual address space for function and workgroup storage. The only practical use of this support is for inspecting the layout of composite types in memory. In anything but this narrow scenario the behavior will be unchanged. This is enough for the CTS vector alignment tests to pass.

This is a less-than-ideal hack but I think it's the best solution to unblock these CTS tests with the current state of SPIR-V and Vulkan.

This also fixes a bug in LongVectorLowering where updated GEP indices were being used to refer to the original type, rather than the cumulative old indices up to the original index being looked at.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -82,6 +82,10 @@ bool HackClampWidth();
 // around a driver bug.
 bool HackMulExtended();
 
+// Returns true if ptrtoint on logical address spaces should be emulated using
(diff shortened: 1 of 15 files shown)


alan-baker (approved):
Looks ok, but conflicts need resolved.

alan-baker:
This is hitting an assertion in the new tests.

PR #990, opened 2022-12-22
Read kernel attributes and add them to non-semantic reflection information

FMarno:
This is to make it possible to implement clGetKernelInfo with the [CL_KERNEL_ATTRIBUTES](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_ATTRIBUTES) flag. The returned string depends on what the implementation supports, so I think it's best that this goes through the compiler.

Since the string is expected to contain attributes as they were declared in the qualifier, I thought it was best that the frontend reads the data so any formatting can be maintained. Adding annotations to the function was the best way I could find of passing the data from the frontend plugin to the passes. AnnotationToMetadataPass was added because the annotation was referencing the entry point functions so it because more difficult to delete and replace them. 

This depends on this change to SPIRV-Headers - https://github.com/KhronosGroup/SPIRV-Headers/pull/308

### DIFF
--- a/lib/AnnotationToMetadataPass.cpp
+++ b/lib/AnnotationToMetadataPass.cpp
@@ -1,0 +1,76 @@
+// Copyright 2022 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


FMarno:
This isn't in a completed state, in particular I don't believe that the produced string is correct, but I'm going on holiday for two weeks. One of my colleagues may finish the work while I'm away.

FMarno:
Unfortunately I no longer have time to complete this work but it would be a shame if this work was lost since these addition will help toward 100% CTS conformance with CLVK. If someone would like to pick up this work and finish it then please do.
I think you'll need to:
- adjust the string generated by stripping the attributes qualifier
- adjust the non-semantic instruction to account for it being optional
- add a couple tests
- tidy things up a little.

rjodinchr:
I am taking over this PR

PR #993, opened 2022-12-29
Draft: Update builtins modules with incoming raft of libclc patches.

aarongreig:
The modules were built with said patches pulled onto the llvm commit currently in deps.json.

Also fixes the lit tests broken by this change.

Since the -no-opaque-pointer flag seems to be broken (and we're moving towards opaque pointers on by default anyway) the new modules were generated with opaque pointers and this will need to wait for that transition to be fully complete before merging. Also some of the lit tests changed are due to the switch to opaque pointers rather than the builtins modules (I was testing and fixing tests with our opaque option hacked to be `true` by default).

Notable changes (listed here to save scrolling a massive diff) to the lit tests include:
* -cl-native-math has been removed from some vec3-to-vec4 tests, I could revert this but it means those lit tests are now testing the builtins link correctly - @rjodinchr do you have an opinion on this since it seems like you added it?
* the additional -enable-opaque flag runs are gone for obvious reasons
* many tests for builtins using scalar types have been drastically simplified as scalar builtin calls get inlined and the alternative was attempting to encode the actual implementation of the builtin - they basically just CHECK-NOT the glsl builtin while the vector variants check that there's a function call happening
* convert_float4_relational.cl and convert_half4_relational.cl have been deleted, these were created to check for an optimization on certain llvm conversion instructions that don't seem possible to generate any more, although I'm looking to see if there's an alternative

With these changes there are still 20 MathBuiltin lit test fails, these all hit an assert in ThreeElementVectorLoweringPass caused by opaque pointers rather than the new builtins modules. I had a bash at debugging but I didn't get to the bottom of it.


rjodinchr:
About the `-cl-native-math` flag in some `vec3-to-vec4` tests, I think it was there by mistake. It can be removed.

While I completely agree that we need to add all those functions into the builtin library, I think it should not change the default behaviour of `clvk`. Which would need another CL to make sure of it. Do we have a list of everything added in those new libaries?

Also, because we are adding the conversion in https://reviews.llvm.org/D136772, we have what looks like a perf regression on all the conversions we used to support already. Maybe those should be either removed from the library (sounds complicated), or filtered out in a new pass similar to the `NativeMathPass` (except that it would be done all the time?).

aarongreig:
Here's the llvm diff, should have included that to start with:

```
diff --git a/libclc/CMakeLists.txt b/libclc/CMakeLists.txt
index 0a23471537df..e00ec62f4b33 100644
--- a/libclc/CMakeLists.txt
+++ b/libclc/CMakeLists.txt
@@ -213,11 +213,11 @@ foreach( t ${LIBCLC_TARGETS_TO_BUILD} )
 
 	# Add the generated convert.cl here to prevent adding
 	# the one listed in SOURCES
-	if( NOT ${ARCH} STREQUAL "spirv" AND NOT ${ARCH} STREQUAL "spirv64" AND
-			NOT ${ARCH} STREQUAL "clspv" AND NOT ${ARCH} STREQUAL "clspv64" )
+	if( NOT ${ARCH} STREQUAL "spirv" AND NOT ${ARCH} STREQUAL "spirv64")
 		set( rel_files convert.cl )
 		set( objects convert.cl )
-		if( NOT ENABLE_RUNTIME_SUBNORMAL )
+		if( NOT ENABLE_RUNTIME_SUBNORMAL AND NOT ${ARCH} STREQUAL "clspv" AND
+		    NOT ${ARCH} STREQUAL "clspv64")
 			list( APPEND rel_files generic/lib/subnormal_use_default.ll )
 		endif()
 	else()
diff --git a/libclc/clspv/lib/SOURCES b/libclc/clspv/lib/SOURCES
index 0466345cee02..cc4e6c66cff2 100644
--- a/libclc/clspv/lib/SOURCES
+++ b/libclc/clspv/lib/SOURCES
@@ -1,48 +1,84 @@
+math/fma.cl
+math/nextafter.cl
 subnormal_config.cl
 ../../generic/lib/geometric/distance.cl
 ../../generic/lib/geometric/length.cl
-math/fma.cl
-math/nextafter.cl
+../../generic/lib/math/acos.cl
 ../../generic/lib/math/acosh.cl
-../../generic/lib/math/asinh.cl
+../../generic/lib/math/acospi.cl
+../../generic/lib/math/asin.cl
 ../../generic/lib/math/atan.cl
+../../generic/lib/math/asinh.cl
+../../generic/lib/math/asinpi.cl
 ../../generic/lib/math/atan2.cl
 ../../generic/lib/math/atan2pi.cl
 ../../generic/lib/math/atanh.cl
 ../../generic/lib/math/atanpi.cl
 ../../generic/lib/math/cbrt.cl
+../../generic/lib/math/clc_exp10.cl
 ../../generic/lib/math/clc_fmod.cl
 ../../generic/lib/math/clc_hypot.cl
 ../../generic/lib/math/clc_ldexp.cl
 ../../generic/lib/math/clc_nextafter.cl
+../../generic/lib/math/clc_pow.cl
+../../generic/lib/math/clc_pown.cl
+../../generic/lib/math/clc_powr.cl
 ../../generic/lib/math/clc_remainder.cl
 ../../generic/lib/math/clc_remquo.cl
 ../../generic/lib/math/clc_rootn.cl
 ../../generic/lib/math/clc_sqrt.cl
 ../../generic/lib/math/clc_tan.cl
+../../generic/lib/math/clc_tanpi.cl
+../../generic/lib/math/cos.cl
+../../generic/lib/math/cosh.cl
+../../generic/lib/math/cospi.cl
 ../../generic/lib/math/erf.cl
 ../../generic/lib/math/erfc.cl
+../../generic/lib/math/exp.cl
+../../generic/lib/math/exp10.cl
+../../generic/lib/math/exp2.cl
+../../generic/lib/math/exp_helper.cl
+../../generic/lib/math/expm1.cl
+../../generic/lib/math/fdim.cl
 ../../generic/lib/math/fmod.cl
 ../../generic/lib/math/fract.cl
 ../../generic/lib/math/frexp.cl
+../../generic/lib/math/half_cos.cl
 ../../generic/lib/math/half_divide.cl
+../../generic/lib/math/half_powr.cl
 ../../generic/lib/math/half_recip.cl
+../../generic/lib/math/half_sin.cl
 ../../generic/lib/math/half_sqrt.cl
+../../generic/lib/math/half_tan.cl
 ../../generic/lib/math/hypot.cl
 ../../generic/lib/math/ilogb.cl
 ../../generic/lib/math/ldexp.cl
 ../../generic/lib/math/lgamma.cl
 ../../generic/lib/math/lgamma_r.cl
+../../generic/lib/math/log.cl
+../../generic/lib/math/log10.cl
+../../generic/lib/math/log1p.cl
+../../generic/lib/math/log2.cl
 ../../generic/lib/math/logb.cl
 ../../generic/lib/math/maxmag.cl
 ../../generic/lib/math/minmag.cl
 ../../generic/lib/math/modf.cl
 ../../generic/lib/math/nan.cl
+../../generic/lib/math/pow.cl
+../../generic/lib/math/pown.cl
+../../generic/lib/math/powr.cl
 ../../generic/lib/math/remainder.cl
 ../../generic/lib/math/remquo.cl
 ../../generic/lib/math/rootn.cl
 ../../generic/lib/math/rsqrt.cl
+../../generic/lib/math/sin.cl
+../../generic/lib/math/sincos.cl
+../../generic/lib/math/sincos_helpers.cl
+../../generic/lib/math/sinh.cl
+../../generic/lib/math/sinpi.cl
 ../../generic/lib/math/sqrt.cl
 ../../generic/lib/math/tables.cl
+../../generic/lib/math/tan.cl
 ../../generic/lib/math/tanh.cl
+../../generic/lib/math/tanpi.cl
 ../../generic/lib/math/tgamma.cl
diff --git a/libclc/generic/include/clc/clcfunc.h b/libclc/generic/include/clc/clcfunc.h
index 55b775ea3935..abb5484d6248 100644
--- a/libclc/generic/include/clc/clcfunc.h
+++ b/libclc/generic/include/clc/clcfunc.h
@@ -2,8 +2,10 @@
 #define _CLC_DECL
 #define _CLC_INLINE __attribute__((always_inline)) inline
 
-/* avoid inlines for SPIR-V since we'll optimise later in the chain */
-#if defined(CLC_SPIRV) || defined(CLC_SPIRV64)
+// avoid inlines for SPIR-V related targets since we'll optimise later in the
+// chain
+#if defined(CLC_SPIRV) || defined(CLC_SPIRV64) || defined(CLC_CLSPV) || \
+    defined(CLC_CLSPV64)
 #define _CLC_DEF
 #else
 #define _CLC_DEF __attribute__((always_inline))
```

aarongreig:
I can maybe have a look at adding something like NativeMathPass to filter out those extraneous conversions, although we're in a bit of weird place now wrt how much time we're allowed to spend and what we spend it on

aarongreig:
I suspect the perf regression may in part (or in whole) be due to increased linking time now that the builtins module is substantially larger - unless you've been measuring exclusively kernel run time?

rjodinchr:
I can take care of that, but most probably in February as I will be OOO for most of January.

For the perf part, I have not run anything, but just looking at the test it feels like it may generate a regressino on the kernel run time side as it now generates more SPIR-V instruction (maybe the Vulkan compiler will generate the same machine code in both the old and new version, but it is not sure).

PR #994, opened 2022-12-30
[WIP] Add support to generic address space

omarahmed1111:
# PR changes
* Add "TransformGenericVolatileMemoryAccess" which will replace volatile from loads and stores on generic pointers. It does that so that "mem2reg" and "InferAddressSpace" passes could get rid of generic pointers. Volatile information was not used currently in SPIRVProducerPass so it should be safe to remove it at the moment.
* Add support for ""to_local", "to_global", "to_private" builtins.
* Add support for AddrspaceCast instruction handling inside "LongVectorLoweringPass".
* Handle global variables (partially done).

# Global Variables Handling
* The current approach is to handle global variables with module constants by combining both of them in one buffer. If this direction was taken, it will need more handling to cases where we have module constants and global variables in the same opencl kernels. (There is a TODO for that inside the code)
* The existing code for that feature is experimental and need to be refactored and may break module constants but as a proof of concept for adding this feature.
* There is an alternative more concrete and suggested approach which is to add a new buffer for global variables different from module constants buffer.

**NOTE**: The current status of this PR may cause regressions on some CTS tests.


rjodinchr:
@omarahmed1111 , if it's okay with you, I'll take over the support of generic pointers.

This PR seems to be a mix of generic pointer and program scope global variables. I will only focus on generic pointers for now.

rjodinchr:
This PR has not been updated for a long time.
Feel free to reopen it if it is still relevant.
If you reopen it, please rebase and fix the conflicts.

PR #997, opened 2023-01-03
Infer function types

alan-baker:
Contributes to #816

* Three element vector lowering infers types of function declarations
  * Just have type inference return the function type in that case


dneto0 (approved):
Thanks!

PR #1000, opened 2023-01-11
Change how the conditionality of inlining works

### LINKED ISSUE #999: Clspv doesn't handle linking well
https://github.com/kpet/clvk/pull/429 added support for clLinkProgram by introducing an early compile to LLVM IR and a late "link" compile to SPIR-V. There is a problem, however, that clspv expects some compile options in the late step. I haven't done a full examination, `-cl-std` is used to control where the generic address space is inferred. This leads to problem with CL2.0 programs not inferring (or inlining) correctly. It is possible there are other options like this.

I'm filing this here because this clspv could infer about the inlining and generic address space based on the contents of the IR.

alan-baker:
Fixes #999

* Make all inlining passes unconditional
* Make generic address space inference unconditional
* Add functionality to inline entry points to look for generic address space and inline in that case
  * also strips constexprs
* Rename `RemovedCstExprFromFunction` to `RemoveCstExprFromFunction`

### DIFF
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -715,7 +715,7 @@ void ConvertInto(Type *Ty, IRBuilder<> &Builder,
   }
 }
 
-bool RemovedCstExprFromFunction(Function *F) {
+bool RemoveCstExprFromFunction(Function *F) {
   SmallVector<std::pair<Instruction *, unsigned>, 16> WorkList;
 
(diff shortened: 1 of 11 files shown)


alan-baker:
> nit: Should the help string on the option be updated to say it's always on?

It's not always on. It's on if requested via the option or if the generic address space is used. So I think the option is still correct, but to make some uses work we end up having to inline more.

PR #1002, opened 2023-01-12
Update deps

alan-baker:
* SPIRV-Tools and SPIRV-Headers switched to track main
* IRBuilder::CreateVectorSplat now uses i64 for indices so updated some tests as a result


dneto0 (approved):
Thanks!

Issue #1009, opened 2023-01-27
ThreeElementVectorLowering and opaque pointers


alan-baker:
In the test test/packed_struct_novec3.cl with opaque pointers, three element vector lowering behaves differently than for non-opaque pointers. The decision to lower is based on the types of certain instructions (like load return type or gep source element type), whereas with non-opaque pointers the decision to lower is based on whether the pointer was lowered.

This leads to a situation in the test where a gep into a packed structure with a vec3 is not lowered, but the subsequent load is lowered. That creates an implicit bitcast that is not properly handled by replace pointer bitcasts (since it expects vec3s to be handled by this pass).

Simple modifications lead to many other test failures where vec3 lowering no longer occurs.

Issue #1004, opened 2023-01-19
Pointer bitcasts and opaque pointers


alan-baker:
There are a few examples where opaque pointers are causing headaches for the pointer bitcast pass.

cts/basic/kernel_preprocessor_macros was fixed by #983, but that seems to lead to incorrect code for: test/PointerCasts/srcGTdst/load_cast_float2_to_float.ll
test/PointerCasts/srcGTdst/load_cast_float4_to_float.ll
test/PointerCasts/srcGT/dst/store_cast_float2_to_float.ll
test/PointerCasts/srcGT/dst/store_cast_float4_to_float.ll

cts/relationals/shuffle_array_cast generates code that requires handling multi-index GEPs due to an implicit gep changing how the interface type is inferred.

test/Scalarize/nested_struct.cl
test/Structs/insert_value_issue_14.cl
test/composite_construct.cl
test/composite_construct_array.cl
test/composite_construct_varying.cl
test/hack_inserts_constant.cl
test/hack_inserts_undef.cl
test/hack_inserts_varying.cl
test/hack_inserts_zero.cl
test/hack_undef_composite.cl
 - SROA breaks down a memcpy of a struct into geps based on bytes, but then stores i32s so pointer bitcast replacement gets confused

alan-baker:
The cts test basic/vector_swizzle is also affected.

PR #1010, opened 2023-01-27
Use opaque pointers by default

alan-baker:
Contributes to #816 

Filed #1009 for another pass issue. Disabled several tests due to #1004.

* Deprecate `--enable-opaque-pointers` option
  * Switch default value to enabled
* Fix some missing inference for some atomics types
* Add assertions that clspv can infer the necessary argument types
* Remove -keep-unused-arguments
  * Option leads to bad codegen with opaque pointers and is mostly used
    for testing anyways
* Add some extra pointer bitcast handling passes
  * simplification and replacement handle some cases that would
    otherwise lead to invalid or worse codegen
* Fix how non-readable is applied to images
* Perform a round of mem2reg early in the flow to add type inference
* Only infer types for the kernel arg map if the pointer arg is used
* Fix some generation of inttoptr (sort of)
* Fix image specialization so that partially specialized images are
  handled correctly
* Make identification of pointer to vec3 bitcasts more accurate in three
  element vector lowering
  * many cases now correctly stay as vec3
* Fixed a bug in constexpr removal for opaque pointers
* Fixed bad file handling
* Moved some included into guarded part of header
* Fixed a bug in ubo multi-versioning
* Fixed opaque pointer handling for function types to reduce duplicates
* More type inferencing
  * Allow constants to be inferred
  * Infer select and memcpy
* Fix a bug in ubo type transforms
* Use the right type in undo byval
* Many test updates


dneto0 (approved):
Great progress!

Issue #1013, opened 2023-02-17
LLVM no longer optimizing extract and insert chains


alan-baker:
Updating deps to 22b564c64b736f5a422b3967720c871c8f9eee9b fails test/LongVectorLowering/function.ll. It no longer removes the extract and insert value chains.

PR #1014, opened 2023-02-17
Update LLVM

alan-baker:
* Problem with insert/extract value chains (#1013) had to modify test expectations


dneto0 (approved):
Thanks for putting the bug number in the TODO

PR #1016, opened 2023-02-20
Fix optimisation of extract and insert chains

### LINKED ISSUE #1013: LLVM no longer optimizing extract and insert chains

rjodinchr:
Since llvm/llvm-project@9ca2c309ab0678fae48f96131b92754d0978c130 folding insertvalue and extractvalue with undef value is now considered unsafe. This is because it can lead to replacing undef with poison.

Instead it is recommanded to use poison.

Fix #1013


test/LongVectorLowering/function.ll
```diff
@@ -50,17 +50,14 @@ define spir_func <8 x i32> @test2(i32 %a) {
 ; CHECK-SAME: @test1([[FLOAT8]] [[X:%[^ ]+]])
 ; CHECK-SAME: !info [[MD:![0-9]+]]
 ; TODO(#1013): should just be X
```
alan-baker, line 52: Seems like these todos can be removed now.

alan-baker (changes requested):
Thanks for this! Please remove the todos from the test for #1013 and it looks good to go.

alan-baker (approved):

alan-baker:
There appear to be some test failures. Likely due to interactions with your other change.

PR #1023, opened 2023-02-24
Fix structured loops

alan-baker:
* Split conditional loop headers when both edges enter the body of the loop


dneto0 (approved):
LGTM.
Thanks!

PR #1024, opened 2023-02-26
Fix definition of get_global_offset() when push constants are used

kpet:
GetPushConstantPointer is designed to return a GEP into the push constant variable. When there is a single push constant or when indexing into the first one (i.e. the global offset), this leads to a GEP with all zero indices being created. With opaque pointers, this GEP is no longer created and CreateInBoundsGEP returns the original pointer. defineGlobalOffsetBuiltin was then assuming that the pointer returned pointed to a specific push constant rather than the global push constant variable itself which led to incorrect indices being used.

After this change, the extra indices are provided to GetPushConstantPointer so the full GEP into the global offset vector is created directly.

It seems that this is a generic issue and users who want to guard against this might now need to check whether CreateInBoundsGEP returned a GEP instruction or the original pointer to guarantee that a chained GEP uses the correct indices. This seems to break encapsulation.

Also support the case where the global push constant variable is used by a load instruction. This happens in a clvk unit test.

### DIFF
--- a/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
+++ b/lib/DefineOpenCLWorkItemBuiltinsPass.cpp
@@ -426,19 +426,19 @@ bool DefineOpenCLWorkItemBuiltinsPass::defineGlobalOffsetBuiltin(Module &M) {
   if (isSupportEnabled) {
     auto Dim = &*F->arg_begin();
     auto InBoundsDim = inBoundsDimensionIndex(Builder, Dim);
-    Value *Indices[] = {Builder.getInt32(0), InBoundsDim};
     Value *gep = nullptr;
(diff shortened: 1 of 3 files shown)


rjodinchr (approved):

kpet:
The clvk job is failing because the SPIRV-LLVM-Translator needs updating to compile with recent LLVM versions. I have an update ready locally but it regresses a fair amount of CTS tests for me as clvk still defaults to transparent pointers. Since transparent pointers are going away soon, I'm planning to transition to opaque pointers as part of the next update but current clspv+LLVM still lead to a high number of regressions (including whole CTS suites crashing, e.g. c11_atomics) and that feels like a bit too much to merge into `main`.

alan-baker (approved):

PR #1039, opened 2023-03-17
fix pointer bitcast managment with opaque pointers

rjodinchr:
fix basic/vector_swizzle and relationals/shuffle_array_cast
move utility functions from replacepointerbitcast to bitcastutils
also add new function getEleTypesBitWidths
do not consider GEP that link well as bitcast
move parts of replacepointerbitcast to simplifypointerbitcast
manage fake casts to avoid generation of unvalid spirv
fix simplify gep from gep for private objects
fix and add new tests


rjodinchr:
no regression on cts with swiftshader and nvidia

alan-baker (approved):

PR #1057, opened 2023-03-28
Fix kernel arg info generation for unused arguments when opaque pointers are enabled

kpet:
This is the pattern used by the CTS tests to validate clGetKernelArgInfo.

The information needed to report kernel arg info via the OpenCL API is carried by ArgumentInfo reflection instructions. These are made available via the last optional operand to Argument* reflection instructions which means those need to be produced for unused arguments too.

This changes relies on the metadata provided by Clang to deduce the argument type. Pseudo-resources are allocated to descriptor set -1 / binding -1 / local memory spec constant -1 for unused arguments and runtimes need to deal with this case (if the approach is accepted, this ought to be documented as part of the specification for the ClspvReflection instruction set).

This change also reinstates the ability of runtimes to correctly report errors when kernel parameters are set to the wrong type of object (though it forces the use of -cl-kernel-arg-info to do so).

### DIFF
--- a/include/clspv/ArgKind.h
+++ b/include/clspv/ArgKind.h
@@ -31,6 +31,7 @@ enum class ArgKind : int {
   Sampler,
   PointerUBO,
   PointerPushConstant,
+  Unknown,
 };
 
 // Converts an ArgKind to its string name.
--- a/lib/AllocateDescriptorsPass.cpp
+++ b/lib/AllocateDescriptorsPass.cpp
@@ -700,7 +700,8 @@ bool clspv::AllocateDescriptorsPass::AllocateKernelArgDescriptors(Module &M) {
           replacement = call;
         } break;
         case clspv::ArgKind::Local:
-          llvm_unreachable("local is unhandled");
+        case clspv::ArgKind::Unknown:
+          llvm_unreachable("local/unknown are unhandled");
         }
 
         if (ShowDescriptors) {
--- a/lib/ArgKind.cpp
+++ b/lib/ArgKind.cpp
@@ -103,7 +103,7 @@ clspv::ArgKind GetArgKindForType(Type *type, Type *data_type) {
 
 namespace clspv {
 
-PodArgImpl GetPodArgsImpl(Function &F) {
+PodArgImpl GetPodArgsImpl(const Function &F) {
   assert(F.hasMetadata(PodArgsImplMetadataName()));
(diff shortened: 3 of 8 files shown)


kpet:
Are you suggesting we'd allocate resource vars, etc for unused arguments or just attempt to use the IR type information to infer the ArgKind before falling back to metadata? What about the runtime interface? If we do allocate resource vars for unused arguments, descriptor set entries would have to be allocated and updated unless we introduce a way of marking arguments as used/unused (likely a new operand in that case).

alan-baker:
> Are you suggesting we'd allocate resource vars, etc for unused arguments or just attempt to use the IR type information to infer the ArgKind before falling back to metadata? What about the runtime interface? If we do allocate resource vars for unused arguments, descriptor set entries would have to be allocated and updated unless we introduce a way of marking arguments as used/unused (likely a new operand in that case).

If we allocate resources for them then clspv would go back to how it used to be. Yes, the runtime would allocate descriptor layouts for them, but this strikes as a case not worth optimizing for particularly. Using (-1,-1) seems like an abuse of the reflection information.

kpet:
I'm not saying it's an absolutely critical optimisation but, without support for bindless, the min requirements for the number of descriptors per stage/pipeline/etc are not that high. A couple of unused arguments here and there can easily make the difference between some applications fitting within the limits or not (and some apps, e.g. Darktable last I checked, have kernels with unused arguments so they can share the same signature as others and be driven by the same runtime code). IMHO that's a worthwhile optimisation.

Having said that, I went with the approach in this PR for two reasons:
- Last we discussed this, you were telling me that it was impossible to infer the argument kinds from types for all kinds of arguments. Sounds like TargetExtTypes might change this so this warrants a revisit.
- If we do want to avoid the resource allocation for unused arguments we ultimately need a way of marking them as such. I thought about introducing a new interface to report whether arguments are used or not but then realised that, in that case, and assuming we wouldn't want major surgery on the reflection instruction set, we'd either specify the descriptor set / binding /local argument spec constant id as "don't care" or some arbitrary value, hence the temptation to ~~abuse~~ creatively use the existing interface :).

I'll have a look at generating resource variables for unused arguments and update this PR if the change looks better. My priority really is to transition clvk to opaque pointers. We can discuss the best approach to not allocating resource variables at the Vulkan level for unused arguments later.

kpet:
Superseded by #1060

PR #1059, opened 2023-03-28
Call LinkBuiltInLibrary only when needed

lpavank:
By checking information from replaceable Builtins pass etc.
Improves BuildProgram Perf based on our testing.
And, brings us closer to our legacy/native stack to an extent.
Note: We have one more small change coming-up after this, but want to keep it separate.

Fix another minor issues to use OpenCL spec Builtin Names.

Co-authored/debugged/analyzed by Raja(rhusain1) and Gowtham(gowtham-sarc) from Samsung.


rjodinchr (changes requested):
When compiling with no option (especially no `-use-native-builtins`), builtins should come from the library.
With this PR, if the kernel is only using builtins for which we have native implementation, we will not link with the library while the actual implementation expect clspv to use the library implementation, not the native version.
I think we should take `clspv::Option::UseNativeBuiltins` and `clspv::Option::NativeMath` into account to make sure not to change the default behavior.

rjodinchr:
Maybe moving the code from `LinkBuiltinLibrary` in `NativeMathPass` would make sense. This is the first pass run in the pipeline.

lpavank:
Thanks @rjodinchr for the feedback. We will make the required changes and get back.

lpavank:
We had a offline conversation with Romaric (thanks @rjodinchr ), last week, and here is the update from that call.
1. Rather than doing what we tried to do with this change, we can fix up the device specific (builtins) that CLVK passes to CLSPV using (--use-native-builtins option), and we would like to update that for our current device and future device in here (https://github.com/kpet/clvk/blob/main/src/device_properties.cpp#L106.  We have that change, its being internally tested before bringing to upstream.
2. The other part of this change (which we did not include here yet, is change the llvm/lib/Bitcode/Reader/BitcodeReader.cpp ParseModule (that will be triggered by LinkInBuiltInLibrary) with the second parameter (MaterializeAll) to "false" instead of "true". Since, this was more on LLVM side, which is not in our control (per say), might require even more due deligence etc. But, Romaric was helpful in suggesting another way to achieve the same effect (perf wise) i.e. call "getLazyIRModule" instead of "ParseIR" from LinkBuiltInLibrary. We gathered some performance (buildprogram) data with that change (overall looks ok). We are doing more testing CL CTS wise on our end, before we push this change. And, will update this PR with this change soon (after internal review/testing).

rjodinchr (approved):
Just need to format `lib/Compiler.cpp` with `clang-format` and it should be good to go for me.

rjodinchr (approved):

rjodinchr:
Not sure what is happening here with all the tests.
Could you try to rebase your branch on top of tree?
Thanks

rjodinchr:
I think this is okay to go in

lpavank:
Thanks @rjodinchr let me know if any else needed here from my side, btw, the CLVK patch thats mentioned here (https://github.com/google/clspv/pull/1059#issuecomment-1520907292) is here https://github.com/kpet/clvk/pull/517.

alan-baker (approved):

PR #1060, opened 2023-03-29
Fix kernel arg info generation for unused arguments when opaque pointers are enabled

kpet:
- Add support for using kernel arg info metadata to deduce the type of kernel function arguments to InferType.
- Generate resource variables for unused arguments.
- Also add missing break in InferType. If fallthrough was intended we should add an attribute.


alan-baker (changes requested):
This looks good, thanks for the changes. Please just add a long vector argument (e.g. float16) to the tests to ensure we cover that case.

kpet:
I've modified an existing argument instead (other arguments covered non-long vectors already).

alan-baker (approved):

kpet:
Some of the asserts added under this PR are firing when running a full CTS sweep. I'll create another PR with a fix once I get a clean CTS run.

PR #1062, opened 2023-03-29
return an error when using -enable-opaque-pointers=0

rjodinchr:
transparent pointers are not supported anymore since we have updated the builtins library using opaque pointers.


alan-baker (changes requested):
This should be accompanied by a change in the deprecation notice in the OpenCLOnVulkan.md docs.

alan-baker (approved):

PR #1065, opened 2023-03-30
fix InferType for non pointer argument

rjodinchr:
clvk/api_tests is failing compiling the test added by this PR.

### DIFF
--- a/lib/Types.cpp
+++ b/lib/Types.cpp
@@ -302,6 +302,7 @@ Type *clspv::InferType(Value *v, LLVMContext &context,
               mdconst::dyn_extract<ConstantInt>(arg_node->getOperand(1))
                   ->getZExtValue();
           ordinal = old_index;
+          break;
         }
       }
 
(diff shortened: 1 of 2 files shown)


rjodinchr:
@kpet  can you have a look at this PR please?

lib/Types.cpp
```diff
@@ -317,8 +317,10 @@ Type *clspv::InferType(Value *v, LLVMContext &context,
           dyn_cast<MDString>(access_qual_op)->getString();
 
       if (access_qual_str == "none") {
-        assert(type_name_str.endswith("*") && "Only expect pointer types here");
-        auto type_name = type_name_str.drop_back(1);
+        auto type_name = type_name_str;
```
alan-baker, line 320: the first bit of InferType is:
```
if (!isa<PointerType>(v->getType()) {
  return v->getType();
}
```

So how is this getting hit?

rjodinchr: I think I found the real bug.

We were missing a break, which lead to overriding `ordinal`

alan-baker (approved):

kpet (commented):
LGTM, thanks! That loop isn't the easiest to read. Hopefully that's the last of it.

PR #1066, opened 2023-03-30
Remove opencl opaque struct support

### LINKED ISSUE #1036: Remove support for opaque struct based opencl types
Updating LLVM in #1035 added support for Clang codegen'ing OpenCL opaque types as `TargetExtType`. This support was added on top of existing support for pointers to opaque structs. The old codegen style support should be removed.

alan-baker:
Fixes #1036

* Remove support for samplers and images based on opaque structures


rjodinchr (approved):
Just need to remove `inner_type` from `ArgKind.cpp` and that should be good to go.

alan-baker:
> Just need to remove `inner_type` from `ArgKind.cpp` and that should be good to go.

I've done that. `GetArgKind` no longer needs the inferred type so there is a bunch of cleanup I can do in a followup PR.

PR #1069, opened 2023-03-31
Specialize unused image arguments

### LINKED ISSUE #1068: Image types are not de-dup'ed when one of the image arguments is unused
The following kernel

```
kernel void foo(write_only image2d_t im1, read_write image2d_t im2) {
    write_imagef(im1, (int2)(0,0), (float4)(0.f,1.f,2.f,3.f));
}
```

compiled with:

```
clspv -o test.spv -cl-std=CL2.0  -inline-entry-points -cl-kernel-arg-info test.cl
```

produces invalid SPIR-V:

```
error: line 32: Duplicate non-aggregate type declarations are not allowed. Opcode: TypeImage id: 13
  %13 = OpTypeImage %float 2D 0 0 0 2 Unknown
```

When both arguments are used the types are correctly de-dup'ed. Prior to https://github.com/google/clspv/pull/1060 / https://github.com/google/clspv/pull/1064 / https://github.com/google/clspv/pull/1065 or when `-cl-kernel-arg-info` is not provided, the types were not generated at all.

alan-baker:
Fixes #1068

* Specialize unused image arguments to prevent type duplication

### DIFF
--- a/test/SpecializeImageTypes/unused.cl
+++ b/test/SpecializeImageTypes/unused.cl
@@ -1,0 +1,6 @@
+// RUN: clspv %s -o %t.spv --cl-std=CL2.0 -inline-entry-points
+// RUN: spirv-val %t.spv
+
+kernel void foo(write_only image2d_t im1, read_write image2d_t im2) {
+    write_imagef(im1, (int2)(0,0), (float4)(0.f,1.f,2.f,3.f));
+}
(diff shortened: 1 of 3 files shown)


test/SpecializeImageTypes/unused.cl
```diff
@@ -0,0 +1,6 @@
+// RUN: clspv %s -o %t.spv --cl-std=CL2.0 -inline-entry-points
```
kpet, line 1: You need to add `-cl-kernel-arg-info` for the resource var to be produced. The issue didn't manifest itself without this.

kpet (changes requested):
Thanks!

PR #1075, opened 2023-04-11
implement bitreverse intrinsic

### LINKED ISSUE #1074: Issue with bitreverse llvm intrinsic
The following code fails to compile:
```cpp
typedef unsigned u32;

static u32 reverse_bits(u32 x) {
    x = (((x & 0xaaaaaaaa) >> 1) | ((x & 0x55555555) << 1));
    x = (((x & 0xcccccccc) >> 2) | ((x & 0x33333333) << 2));
    x = (((x & 0xf0f0f0f0) >> 4) | ((x & 0x0f0f0f0f) << 4));
    x = (((x & 0xff00ff00) >> 8) | ((x & 0x00ff00ff) << 8));
    return ((x >> 16) | (x << 16));
}

kernel void k(global u32* in, global u32* out) {
    unsigned index = get_global_id(0);
    out[index] = reverse_bits(in[index]);
}
```

The issue is that llvm converts the `reverse_bits` into an `llvm.bitreverse` intrinsic (at any opt level), but clspv is not able to handle it.

godbolt link : https://godbolt.org/z/fP357bYq4

rjodinchr:
fix #1074

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -4343,6 +4343,11 @@ SPIRVID SPIRVProducerPassImpl::GenerateInstructionFromCall(CallInst *Call) {
   case Intrinsic::ctpop: {
     return GeneratePopcount(Call->getType(), Call->getArgOperand(0), Context);
   }
+  case Intrinsic::bitreverse: {
+    SPIRVOperandVec Ops;
+    Ops << Call->getType() << Call->getArgOperand(0);
+    return addSPIRVInst(spv::OpBitReverse, Ops);
+  }
 
   default:
     break;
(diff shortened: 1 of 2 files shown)


kpet:
Any reason you're not using OpBitReverse?

rjodinchr:
It requires the `BitInstructions` capability. I did not check if all devices support `SPV_KHR_bit_instructions`.

kpet:
It's also enabled by the `Shader` capability that all devices support. If I remember correctly, the `BitInstructions` capability was added later as an alternative way of enabling that instruction for use in OpenCL (BTW we should implement https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_extended_bit_ops at some point).

PR #1081, opened 2023-04-18
Add first support for generic address space

### LINKED ISSUE #1077: Error with nullptr checks
The following code fails with an assertion about "Attempting to get element type of opaque pointer" caused by the commented line.
```cpp
inline float loop(const float *data, unsigned num) {
    float res = 0;
    for (unsigned j = 0; j < num; ++j) {
        if (!data) break; // issue here. Ok if commented out.
        res += data[j];
    }
    return res;
}

kernel void k(global float* in, global float* out) {
    unsigned index = get_global_id(0);
    out[index] = loop(in, index);
}
```

godbolt link: https://godbolt.org/z/dWe1v4cYP

rjodinchr:
This is taking over some work from #994
Fixes #1077

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -169,11 +169,15 @@ enum class SourceLanguage {
 
 SourceLanguage Language();
 
+std::set<FeatureMacro> EnabledFeatureMacros();
+
 // Returns true when the source language makes use of the generic address space.
 inline bool LanguageUsesGenericAddressSpace() {
   return (Language() == SourceLanguage::OpenCL_CPP) ||
          (Language() == SourceLanguage::OpenCL_C_20) ||
-         (Language() == SourceLanguage::OpenCL_C_30);
+         (Language() == SourceLanguage::OpenCL_C_30 &&
+          EnabledFeatureMacros().count(
+              clspv::FeatureMacro::__opencl_c_generic_address_space) > 0);
 }
 
 // Return the SPIR-V binary version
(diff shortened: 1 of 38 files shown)


test/AddressSpaceCast/to_global.cl
```diff
@@ -0,0 +1,18 @@
+// RUN: clspv --cl-std=CLC++ --inline-entry-points %s -o %t.spv --show-producer-ir &> %t.ll
+// RUN: spirv-val --target-env spv1.0 %t.spv
+// RUN: FileCheck %s < %t.ll
+
+// CHECK-NOT: addrspacecast
+
+float loop(const float *data, unsigned num) {
+    float res = 0;
+    for (unsigned j = 0; j < num; ++j) {
+        res += to_local(data)[j];
```
alan-baker, line 10: This seems wrong? The test name is to_global and the pointer that is passed in here is a global pointer. Are we actually generating correct code?

rjodinchr: Yes, this is an error.
The code generated was "correct", it was doing the same thing as the `incompatible_cast.ll` test.
When to call `to_<addrspace>` with an incompatible pointer, we just need to return `null`.

include/clspv/Option.h
```diff
@@ -172,7 +172,8 @@ SourceLanguage Language();
 // Returns true when the source language makes use of the generic address space.
 inline bool LanguageUsesGenericAddressSpace() {
   return (Language() == SourceLanguage::OpenCL_CPP) ||
-         ((Language() == SourceLanguage::OpenCL_C_20));
+         (Language() == SourceLanguage::OpenCL_C_20) ||
+         (Language() == SourceLanguage::OpenCL_C_30);
```
alan-baker, line 176: It feels like this should be based on the feature being enabled. I'm not sure by default is the right choice given it requires full inlining. If clvk wants to always enable it that's ok, but I'm a bit wary of forcing that decision on all users.

alan-baker (approved):

PR #1084, opened 2023-04-20
fix fixup for sqrt/rsqrt

### LINKED ISSUE #1083: Assertion hit while compiling kernels from Darktable
Compiling the attached source with the following options

```
clspv source.cl  -w -cl-fast-relaxed-math  -DNVIDIA=1 -I"/usr/share/darktable/kernels" -cl-single-precision-constant  -cl-kernel-arg-info   -fp16=0  -rewrite-packed-structs  -std430-ubo-layout  -decorate-nonuniform    -arch=spir  --use-native-builtins=acos,acosh,acospi,asin,asinh,asinpi,atan,atan2,atan2pi,atanh,atanpi,ceil,copysign,fabs,fdim,floor,fma,fmax,fmin,frexp,half_rsqrt,half_sqrt,isequal,isfinite,isgreater,isgreaterequal,isinf,isless,islessequal,islessgreater,isnan,isnormal,isnotequal,isordered,isunordered,ldexp,mad,rint,round,rsqrt,signbit,sqrt,tanh,trunc,  -spv-version=1.6  -max-pushconstant-size=256  -max-ubo-size=65536  -global-offset  -long-vector  -module-constants-in-storage-buffer  -cl-arm-non-uniform-work-group-size     -o compiled.spv
```

results in the following assertion being hit:

```
clspv: /mnt/work/repos/clvk/external/clspv/third_party/llvm/llvm/include/llvm/Support/Casting.h:578: decltype(auto) llvm::cast(From*) [with To = SelectInst; From = Value]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```

Darktable used to run on top of clvk (not sure it was the exact same version) so this may be a regression.

[source.txt](https://github.com/google/clspv/files/11275557/source.txt)

rjodinchr:
when sqrt/rsqrt input is a constant the IRBuilder optimize the select to directly get the right node. This lead to an assertion trying to cast select into an SelectInst.
Instead, just manage constant values staticaly

Fix #1083

### DIFF
--- a/lib/FixupBuiltinsPass.cpp
+++ b/lib/FixupBuiltinsPass.cpp
@@ -24,6 +24,10 @@
 using namespace clspv;
 using namespace llvm;
 
+namespace {
+double rsqrt(double input) { return 1.0 / sqrt(input); }
+} // namespace
+
 PreservedAnalyses FixupBuiltinsPass::run(Module &M, ModuleAnalysisManager &) {
   PreservedAnalyses PA;
   for (auto &F : M) {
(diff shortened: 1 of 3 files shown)


alan-baker (approved):

rjodinchr:
Do not merge. I've just noticed that I used `sqrt` while the function can either be `sqrt` or `rsqrt`

Issue #1071, opened 2023-04-06
Assertion in BitcastUtils when compiling kernels used by x264


kpet:
Compiling the attached source file with the following command line:

```
clspv source.cl  -cl-single-precision-constant  -cl-kernel-arg-info   -fp16=0  -rewrite-packed-structs  -std430-ubo-layout  -decorate-nonuniform    -arch=spir  -spv-version=1.6  -max-pushconstant-size=256  -max-ubo-size=65536  -global-offset  -long-vector  -module-constants-in-storage-buffer  -cl-arm-non-uniform-work-group-size     -o compiled.spv
```

results in the following assertion being hit:

```
clspv: /mnt/work/repos/clvk/external/clspv/lib/BitcastUtils.cpp:375: void BitcastUtils::GroupScalarValuesIntoVector(llvm::IRBuilder<>&, llvm::SmallVector<llvm::Value*, 8>&, unsigned int): Assertion `Values.size() % NumElePerVec == 0' failed.
```

[source.txt](https://github.com/google/clspv/files/11174007/source.txt)

rjodinchr:
I have had a first look at it.
The assertion is coming from the run of the ReplacePointerBitcastPass on the `parallel_sum` function.
What I don't understand is that we can see that this function is using the local `cache` variable from `mb_intra_cost_satd_8x8`. But when I look for where the `parallel_sum` function is called, I don't see any link with this variable.

Applying this patch:
```
diff --git a/lib/BitcastUtils.cpp b/lib/BitcastUtils.cpp
index fc497d32..fb3e5142 100644
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -372,8 +372,10 @@ void GroupScalarValuesIntoVector(IRBuilder<> &Builder,
   Type *SrcTy = Values[0]->getType();
   assert(!SrcTy->isVectorTy() && !SrcTy->isArrayTy());
   VectorType *DstTy = FixedVectorType::get(SrcTy, NumElePerVec);
-  assert(Values.size() % NumElePerVec == 0);
   unsigned int NumVector = Values.size() / NumElePerVec;
+  if (Values.size() > NumElePerVec && Values.size() % NumElePerVec != 0) {
+    Values.resize(NumVector * NumElePerVec);
+  }
   for (unsigned i = 0; i < NumVector; i++) {
     unsigned idx = i * NumElePerVec;
     Value *Vec = PoisonValue::get(DstTy);
```
[...]

Issue #723, opened 2021-03-27
Unreachable hit: Support above GEP on PointerBitcastPass


PlasmaPower:
This kernel https://gist.github.com/PlasmaPower/75b1e84f949d761e27bc283014ebb4ae hits the unreachable here https://github.com/google/clspv/blob/589d72ecbf5b1ea813d1b3d1b7a43cbd6d220c87/lib/ReplacePointerBitcastPass.cpp#L278

I've been unable to produce an MRE so far, but it's something with the LLVM getelementpointer.

PlasmaPower:
I've finally reduced this issue with the help of a program that randomly removes lines of code and checks if it still crashes:
```c
typedef uint arr[10];

void example() {
	const arr one = {1};
}
```

PlasmaPower:
This seems to be still failing, albeit with a different error message than the one indicated in the original issue:
```
  %4 = getelementptr inbounds <{ i32, [9 x i32] }>, <{ i32, [9 x i32] }>* %2, i32 0, i32 0  %4 = getelementptr inbounds <{ i32, [9 x i32] }>, <{ i32, [9 x i32] }>* %2, i32 0, i32 0[1]    1591586 segmentation fault (core dumped)  clspv kernel.cl
```
It might just be that my current build doesn't have debug assertions though

rjodinchr:
Hi @PlasmaPower,

This example is still failing on the same line you mentioned above.
It still needs to be fixed in clspv.

But even after this issue is fixed, your example won't compile for 2 reasons:
- printf is not supported yet
- the code is doing cast that are not supported in clspv yet (storing an i8 to an element declared as i64).

PR #1085, opened 2023-04-20
Support storing element smaller than the object it is stored in

### LINKED ISSUE #728: Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed
Kernel: https://gist.github.com/PlasmaPower/9c8abea0a6bdb31e76cb5b3b74a08661

This one takes a while to crash, so I might have a harder time reducing it. I'll look into making a release build with assertions enabled.

### LINKED ISSUE #1031: clspv crashes with char member access of local struct
Hi,

clspv (273c9923) crashes when I tried simple load and store functions of char member of local address struct.

local_struct_i8.cl,
```
struct Storage
{
  char m_i8[8];
  short m_i16[4];
};

template <typename Type, typename T>
Type load2G(__global const T* ptr)
{
  Type v;
  v.x = ptr[0];
  v.y = ptr[1];
  return v;
}

template <typename Type, typename T>
void store2L(const Type v, __local T* ptr)
{
  ptr[0] = v.x;
  ptr[1] = v.y;
}

__kernel void localStructTest(__global char* inout_i8,
                              __global short* inout_i16,
                              __local Storage* storage)
{
  const size_t lindex = get_local_linear_id();
  if (lindex == 0) {
    // Initialize the local storage with something
    {
      const char2 v = load2G<char2, char>(inout_i8);
      store2L<char2, char>(v, &storage[0].m_i8[0]);
    }
    {
      const short2 v = load2G<short2, short>(inout_i16);
      store2L<short2, short>(v, &storage[0].m_i16[0]);
    }
  }
}
```

Here is the build result,
```
clspv --cl-std=CLC++ --inline-entry-points local_struct_i8.cl -o local_struct_i8
%struct.Storage = type { [8 x i8], [4 x i16] }i8*** stack smashing detected ***: terminated
[1]    9509 abort (core dumped)  clspv --cl-std=CLC++ --inline-entry-points local_struct_i8.cl -o
```

I tried it on Ubuntu 22.04.

Thanks.

rjodinchr:
Support storing element smaller than the object it is stored in by reworking instruction to change the infertype of the destination object.

+ minor fixes in BitcastUtils and ReplacePointerBitcastPass

Fixes #728
Fixes #1031
Ref #1071
Ref #723


alan-baker:
Did you consider actually changing infer type to infer the smallest type? The current incarnation of InferType is mostly there to get up and running. If we ought to be inferring different types I'd be interested in localizing it to that function.

rjodinchr:
Yes, I tried quite some ideas to get that working only by modifying `InferTypes`. But it was always getting quite complicated and introduced big regressions when running the CTS.

I'll have a new look at it, because since I got this version working, the way I'm doing it changed quite a bit. So maybe this version can help me figure out a clean way to do the same thing by only modifying `InferTypes`.

rjodinchr:
I have moved what I could into InferTypes. It has surfaced quite some bugs, so fixing them here as well.

rjodinchr:
No regression on full CTS for swiftshader and nvidia

lib/BitcastUtils.cpp
```diff
@@ -945,13 +961,24 @@ Value *CreateRem(IRBuilder<> &Builder, unsigned rem, Value *Val) {
   }
 }
 
+bool IsArrayLike(StructType *Ty) {
+  Type *ElemTy = Ty->getStructElementType(0);
```
alan-baker, line 965: You should test for an empty struct first.

alan-baker (changes requested):
One minor thing then it looks ok.

lib/BitcastUtils.cpp
rjodinchr: Done

alan-baker (approved):

PR #1103, opened 2023-05-12
add support for vstore_half and generic pointers

### LINKED ISSUE #1079: Segmentation fault happens when using vload_half and vstore_half
Hi, I got a segmentation fault when I use vload_half and vstore_half.
I used clspv at (9166d322).

vloadstore_half_test.cl,
```
__kernel void vloadstoreHalfTest(__global half* inout, __local half* storage)
{
  const float k = 2.0f;
  // scalar
  {
    const float v1 = vload_half(0, inout);
    vstore_half(v1 * k, 0, storage);
    const float v2 = vload_half(0, storage);
    vstore_half(v2 * k, 0, inout);
  }
  // vector2
  {
    const float2 v1 = vload_half2(0, inout);
    vstore_half2(v1 * k, 0, storage);
    const float2 v2 = vload_half2(0, storage);
    vstore_half2(v2 * k, 0, inout);
  }
  // vector3
  {
    const float3 v1 = vload_half3(0, inout);
    vstore_half3(v1 * k, 0, storage);
    const float3 v2 = vload_half3(0, storage);
    vstore_half3(v2 * k, 0, inout);
  }
  // vector4
  {
    const float4 v1 = vload_half4(0, inout);
    vstore_half4(v1 * k, 0, storage);
    const float4 v2 = vload_half4(0, storage);
    vstore_half4(v2 * k, 0, inout);
  }
}
```

Here is the result,
```
clspv --cl-std=CLC++ --inline-entry-points --fp16 vloadstore_half_test.cl -o vloadstore_half_test
[1]    7915 segmentation fault (core dumped)  clspv --cl-std=CLC++ --inline-entry-points --fp16 vloadstore_half_test.cl -o
```

Thanks,

rjodinchr:
add vstore_half and sincos to the builtins with generic pointer list.

Reorganise passes to have a ReplaceOpenCLBuiltinPass before and after LowerAddrSpaceCastPass. Before is needed to deal with to_global/to_private/to_local. After is needed to deal with vstore_half.

Minor fixes in GetEquivalentBuiltinsWithoutGenericPointer.

Clean unused functions in InlineEntryPoints. This is needed to avoid having dead instructions not removed in loweraddrspacecastpass.

Fixes in LowerAddrSpaceCastPass:
- fix LowerAddrSpaceCastPass since we updated when we run this pass
- remove function that have 1 use with is in fact coming from the GlobalVariable used to make sure some libclc do not get removed because not yet used.

Fix #1079

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -547,6 +547,7 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     pm.addPass(clspv::ClusterPodKernelArgumentsPass());
 
     pm.addPass(clspv::InlineEntryPointsPass());
+    pm.addPass(clspv::FunctionInternalizerPass());
     // This pass needs to be after every inlining to make sure we are capable of
     // removing every addrspacecast. It only needs to run if generic addrspace
     // is used.
@@ -932,15 +933,14 @@ bool GetEquivalentBuiltinsWithoutGenericPointer(llvm::Module *module,
     };
     std::string str = get_fct_decl(F);
 
-    auto add_impl = [&found, &str, &stream_fct_decl,
-                     &stream_fct_list](std::string mangling, std::string AS) {
+    auto add_impl = [&found, &str, &stream_fct_decl, &stream_fct_list,
+                     &module](std::string mangling, std::string AS) {
       auto substr = [&str](size_t start, size_t end) {
         return str.substr(start, end - start);
       };
       if (!found) {
-        stream_fct_decl << "target datalayout = "
-                           "\"e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:"
-                           "128-v192:256-v256:256-v512:512-v1024:1024\"";
+        stream_fct_decl << "target datalayout = \""
+                        << module->getDataLayoutStr() << "\"";
       }
       std::string mangling_pattern = "PU3AS4";
       auto mangling_start = str.find(mangling_pattern);
--- a/lib/InlineEntryPointsPass.cpp
+++ b/lib/InlineEntryPointsPass.cpp
@@ -40,16 +40,6 @@ PreservedAnalyses clspv::InlineEntryPointsPass::run(Module &M,
     changed &= InlineFunctions(M);
   }
 
-  SmallVector<Function *> ToBeRemoved;
-  for (auto &F : M) {
-    if (F.getNumUses() == 0 && F.getCallingConv() != CallingConv::SPIR_KERNEL) {
-      ToBeRemoved.push_back(&F);
-    }
-  }
-  for (auto F : ToBeRemoved) {
-    F->eraseFromParent();
-  }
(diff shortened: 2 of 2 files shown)


lib/Builtins.cpp
```diff
@@ -778,7 +778,8 @@ glsl::ExtInst Builtins::getDirectOrIndirectExtInstEnum(
 bool Builtins::BuiltinWithGenericPointer(StringRef name) {
   if (name.contains("fract") || name.contains("frexp") ||
       name.contains("modf") || name.contains("remquo") ||
-      name.contains("lgamma_r"))
+      name.contains("lgamma_r") || name.contains("vstore_half") ||
```
alan-baker, line 781: Is this meant to only list function from libclc? Because there are other builtins that use the generic address space (e.g. atomics).

lib/Compiler.cpp
```diff
@@ -545,6 +545,16 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     //   %3 = load float %2
     pm.addPass(llvm::createModuleToFunctionPassAdaptor(llvm::PromotePass()));
     pm.addPass(clspv::ClusterPodKernelArgumentsPass());
+
+    pm.addPass(clspv::InlineEntryPointsPass());
```
alan-baker, line 549: It might be worth adding a pass of the function internalizer here too to avoid unnecessary work after inlining.

lib/Compiler.cpp
```diff
@@ -935,6 +937,11 @@ bool GetEquivalentBuiltinsWithoutGenericPointer(llvm::Module *module,
       auto substr = [&str](size_t start, size_t end) {
         return str.substr(start, end - start);
       };
+      if (!found) {
```
alan-baker, line 940: I'm confused by this. Why would the data layout be missing? Also it seems it should use whatever the current compile's target is (e.g. spir vs spir64). This seems to always be 32-bit pointers.

lib/Compiler.cpp
alan-baker: Oh I see later on you've basically incorporated that pass into inline entry points. Not sure which way is better, so I'll leave to you.

lib/Compiler.cpp
rjodinchr: You are right, this is not good, I'll rework it.

lib/Builtins.cpp
rjodinchr: Yes this is just for libclc, the goal being to keep function from libclc even though they might not be used.

lib/Compiler.cpp
rjodinchr: I have updated it to use function internalizer.

lib/Compiler.cpp
rjodinchr: When this module has no layout, it generates a warning message when linking it with the main one, because they are not identical.

alan-baker (approved):

PR #1117, opened 2023-05-22
Implement Printf

lpavank:
The bulk of the work for this PR came from:
https://github.com/google/clspv/pull/925

We (Samsung) took over Callum's work, to fix the issues with respect to some of the tests failures he had (the details of those are there in (https://github.com/google/clspv/commit/d2208094f5b31ac5b11f714ed89744beba506813) commit message towards the end.

Note: from Callum's original PR we removed this test (printf_physical_buffer.cl) test case, since we dont have a root cause for that failure yet. We were thinking going ahead with the PR, is a step in forward direction, and we can always come back and add that one test, please correct us if you disagree or have a different opinion here.


rjodinchr (changes requested):
Can you bring back Callum's modification of `lib/DeclarePushConstantsPass.cpp`, `lib/PushConstant.cpp` & `lib/PushConstant.h`?
To me those are needed if we aim at making it work.
Also please keep `printf_physical_buffer.cl` in this PR. If it fails, just add `// XFAIL: *` in it, so that the test suite can run it and check for failure on it.

lpavank:
> Can you bring back Callum's modification of `lib/DeclarePushConstantsPass.cpp`, `lib/PushConstant.cpp` & `lib/PushConstant.h`? To me those are needed if we aim at making it work. Also please keep `printf_physical_buffer.cl` in this PR. If it fails, just add `// XFAIL: *` in it, so that the test suite can run it and check for failure on it.

Thanks. Yes bringing back those, and will bring back the test as well with "expected failure".

lpavank:
The latest commit, should bring up all of Callum's Physical storage buffers commit from his original PR, with expected failure on the test. Please have a look.

rjodinchr:
Please fix the formatting, and the `max` issue (I think it is about the type of the input not being the same or something like that.
Then it should be good to go.
I'll take care of fixing `printf_physical_buffer.cl`, I already have the fix.

lpavank:
The formatting should be taken care of (not sure how it was missed in the earlier commit (https://github.com/google/clspv/pull/1117/commits/d24806a7dadcf875986dae7285caed294d3420d1) because I did run check-code-format.sh and fixed a few of formatting errors from Callum's cherry pick. Anyways, it (formatting) should be addressed.

lib/PrintfPass.cpp
```diff
@@ -0,0 +1,318 @@
...
+        return InitArray->getAsCString().str();
+      }
+    }
+  }
+
+  return "";
+}
+
+unsigned clspv::PrintfPass::GetPrintfStoreSize(const DataLayout &DL, Type *Ty) {
+  return std::max(4ul, DL.getTypeStoreSize(Ty).getFixedValue());
```
rjodinchr, line 62:
```suggestion
  return std::max(4ul, (unsigned long)DL.getTypeStoreSize(Ty).getFixedValue());
```

rjodinchr: It seems that on windows and macos `getFixedValue` returns an `unsigned long long`, so it needs to be casted to avoid a type mismatched in `std::max`.

rjodinchr (changes requested):
Please rebase your PR and fix the conflicts.
Then you just have to fix `lib/PrintfPass.cpp:62`.

lpavank:
Thanks @rjodinchr merged from upstream/main, resolved conflicts/formatting/max issue. Please let me know if any thing is needed here.

test/Printf/printf_physical_buffer.cl
```diff
@@ -0,0 +1,37 @@
+// RUN: clspv -arch=spir64 %s -o %t.spv -enable-printf -physical-storage-buffers
+// RUN: spirv-dis -o %t2.spvasm %t.spv
+// RUN: FileCheck %s < %t2.spvasm
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+// XFAIL: *
```
alan-baker, line 5: Please add a reason, or better yet file an issue and reference it here.

alan-baker (changes requested):
Almost there, just some small changes.

test/Printf/printf_physical_buffer.cl
rjodinchr: FYI, I will push a fix for this test when this PR is merged.

rjodinchr:
@lpavank , do you want me to take over this PR?

lpavank:
We took care of the latest comments from Alan and rebased as well. Hopefully this is inline with your expectations.

alan-baker (approved):

Issue #1109, opened 2023-05-16
clspv gets caught in the assertion in llvm::checkGEPType


byzin:
Hi,

Clspv gets caught in an assertion with the following structure member access and copy code,

```
// type_test1.cl
struct TestStruct
{
  short i16_;
  short pad1_;
  unsigned int u32_;
  unsigned char u8_;
  unsigned char pad2_;
  ushort u16_;
  float f_;
};

__kernel void test(__global const TestStruct* inputs, __global TestStruct* outputs)
{
  const size_t index = get_global_id(0);
  __global const TestStruct* in = inputs + index;

  // test1
  {
    __global TestStruct* o = outputs + (2 * index + 0);
    o->i16_ = in->i16_;
    o->u32_ = in->u32_;
    o->u8_ = in->u8_;
    o->u16_ = in->u16_;
    o->f_ = in->f_;
  }

  // test2
  {
    __global TestStruct* o = outputs + (2 * index + 1);
    o[0] = in[0];
  }
}
```
and here is the result with debug build clspv (d75d76b9),
```
% clspv --cl-std=CLC++ --inline-entry-points type_test1.cl -o type_test1                                                                                                                                                                                                                                                                                                                                         Downloads/test zibuntu
clspv: /home/sho/Projects/other/clspv/third_party/llvm/llvm/include/llvm/IR/Instructions.h:933: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
[1]    62846 abort (core dumped)  clspv --cl-std=CLC++ --inline-entry-points type_test1.cl -o type_test1
```

It's the assertion in the llvm function so it might be a regression of llvm but I reported here just in case.

The same assertion happend with the following member functino call,

```
// type_test2.cl
struct Storage
{
  void set(const int value)
  {
    __generic int* ptr = reinterpret_cast<__generic int*>(&value_);
    for (size_t i = 0; i < 4; ++i)
      ptr[i] = value + static_cast<int>(i);
  }

  int sum() const
  {
    __generic const int* ptr = reinterpret_cast<__generic const int*>(&value_);
    int sum = 0;
    for (size_t i = 0; i < 4; ++i)
      sum += ptr[i];
    return sum;
  }

  int4 value_;
};

__kernel void test(__global int* outputs, __local Storage* storage)
{
  const size_t storage_id = get_local_id(0);
  __local Storage* s = &storage[storage_id];

[...]

PR #1118, opened 2023-05-24
update pointer bitcast management

### LINKED ISSUE #1115: Incorrect code transform in simplify pointer bitcasts
`test6` in `test/PointerCasts/opaque_trivial_casts.ll` produces an incorrect transformation.

It assumes that certain address spaces cannot have non-zero first index in the GEP. While SPIR-V cannot use PtrAccessChain in those address spaces, producing incorrect code is worse than invalid code.

rjodinchr:
- clean GetIdxsForTyFromOffset for Constant Path with structure types
- reorganise simplify function
- add runOnUnneededIndices to avoid implicitCasts issue with #1109
- add debug log in simplify pointer bitcast pass
- update & add tests

This PR is needed to solve issue from #1109

Fix #1115

### DIFF
--- a/test/PointerCasts/opaque_trivial_casts.ll
+++ b/test/PointerCasts/opaque_trivial_casts.ll
@@ -78,19 +78,6 @@ entry:
   ret void
 }
 
-; CHECK-LABEL: define void @test6(ptr addrspace(1) %in) {
-; CHECK: entry:
-; CHECK:   getelementptr float, ptr addrspace(1) %in, i32 2
-; CHECK-NEXT: ret void
-; CHECK: }
-
-define void @test6(ptr addrspace(1) %in) {
-entry:
-  %gep1 = getelementptr float, ptr addrspace(1) %in, i32 1
-  %gep2 = getelementptr i32, ptr addrspace(1) %gep1, i32 1
-  ret void
-}
-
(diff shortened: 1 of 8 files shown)


test/PointerCasts/opaque_trivial_casts.ll
```diff
@@ -78,17 +78,16 @@ entry:
...
-; CHECK-LABEL: define void @test6(ptr %in) {
+; CHECK-LABEL: define void @test6(ptr addrspace(1) %in) {
 ; CHECK: entry:
-; CHECK:   getelementptr float, ptr %in, i32 1
+; CHECK:   getelementptr float, ptr addrspace(1) %in, i32 2
 ; CHECK-NEXT: ret void
 ; CHECK: }
 
-define void @test6(ptr %in) {
+define void @test6(ptr addrspace(1) %in) {
```
alan-baker, line 87: I think this is the same as test7 now.

alan-baker (commented):
I'm not sure how this is a fix for #1115. The test simply points out that we generate incorrect code. Changing the test to a different address space doesn't change that fact.

rjodinchr:
You are right, I did not notice that it was the same as `test7`.
I have removed `test6` and added a test showing how we should output an error when we encounter such invalid cast.

alan-baker (approved):

PR #1119, opened 2023-05-25
fix memcpy

### LINKED ISSUE #1109: clspv gets caught in the assertion in llvm::checkGEPType

rjodinchr:
As we are not capable of garante that a type won't change during the compilation passes, we cannot use copymemory as types might have changed between the time we create the node and when we want to produce the equivalent spirv.
Thus always rely on load/store to perform the copy.

fix #1109


rjodinchr:
I need to add some more comments in the code, and I forgot to add a test as well. I'll do that tomorrow

alan-baker (approved):

PR #1131, opened 2023-06-13
fix phi from gep in simplify pointer bitcast

rjodinchr:
This pattern has been seen in tflite.
Not lowering it leads to error such as:
```
OpPhi's result type <id> '282[%_ptr_StorageBuffer_uchar]' does not match incoming value <id> '123[%123]' type <id>'122[%_ptr_StorageBuffer_v4half]'.
```

### DIFF
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -18,6 +18,7 @@
 #include "clspv/AddressSpace.h"
 #include "clspv/Option.h"
 #include <cmath>
+#include <llvm/IR/Instructions.h>
 #include <llvm/Support/Debug.h>
 
 #define DEBUG_TYPE "bitcastutils"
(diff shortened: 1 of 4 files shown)


rjodinchr:
No regression on full cts with nvidia and swiftshader

lib/BitcastUtils.cpp
```diff
@@ -847,6 +848,30 @@ bool IsImplicitCasts(Module &M, DenseMap<Value *, Type *> &type_cache,
     source_ty =
         clspv::InferType(st->getPointerOperand(), M.getContext(), &type_cache);
     dest_ty = st->getValueOperand()->getType();
+  } else if (auto *phi = dyn_cast<PHINode>(&I)) {
+    if (phi->getNumOperands() > 0) {
+      auto RefOp = phi->getOperand(0);
+      auto RefOpTy = clspv::InferType(RefOp, M.getContext(), &type_cache);
```
alan-baker, line 854: Is there a reason the reference type isn't an inference on the result of the phi? That seems like a more likely destination type in terms of an implicit cast.

lib/BitcastUtils.cpp
```diff
@@ -847,6 +848,30 @@ bool IsImplicitCasts(Module &M, DenseMap<Value *, Type *> &type_cache,
     source_ty =
         clspv::InferType(st->getPointerOperand(), M.getContext(), &type_cache);
     dest_ty = st->getValueOperand()->getType();
+  } else if (auto *phi = dyn_cast<PHINode>(&I)) {
+    if (phi->getNumOperands() > 0) {
+      auto RefOp = phi->getOperand(0);
```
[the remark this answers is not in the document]
rjodinchr: Done

lib/BitcastUtils.cpp
rjodinchr: I think the type of a phi is determined by InferType using the users of the node. Then if the users of a phi do not have the same cast, I believe this is already handled by `simplifypointercast`.

Here what I am added is to target implicit cast regarding the phi inputs, regardless of the output and the users.
If `RefOpTy` uses the phi type (the smallest of the users), we would not know exactly what to do if an input type is different and smaller.

I'll look at it again on Monday to make sure of it.

alan-baker (approved):

PR #1132, opened 2023-06-13
fix build-clvk.sh since vulkan-headers are not part of clvk anymore

rjodinchr:


alan-baker:
It seems like the vulkan-headers are too old. I'd suggest checking out some specific version and using that.

alan-baker (approved):

PR #1136, opened 2023-06-16
Implement saturation mode for convert functions

AlexDemydenko:


rjodinchr:
I don't think this PR is actually doing anything.
I have worked on fixing the saturated conversion functions in libclc.
Please try this instead: https://github.com/google/clspv/pull/1141

AlexDemydenko:
My change wasn't on the top of the tree. This bug was fixed by one of the previous commits in a different way.
Now this PR is obsolete and #1141 I think so too.

PR #1139, opened 2023-06-21
add hack option to prevent unwanted driver optimisation

rjodinchr:
Some driver optimize pattern this pattern to completely remove the two
conversions, while libclc relies on this pattern to be executed to
perform rounded conversions.
```
%0 = OpConvertSToF %float %input
%1 = OpConvertFToS %int %0
```
The hack-convert-to-float option add a dummy instruction after
`OpConvertSToF` to prevent this optimisation.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -86,6 +86,10 @@ bool HackMulExtended();
 // compile-time constants when it is safe to do so.
 bool HackLogicalPtrtoint();
 
+// Returns true if a dummy instruction should be inserted after conversion to
+// float to prevent driver optimisation getting rid of the conversion.
+bool HackConvertToFloat();
+
 // Returns true if module-scope constants are to be collected into a single
 // storage buffer.  The binding for that buffer, and its intialization data
 // are given in the descriptor map file.
--- a/lib/Option.cpp
+++ b/lib/Option.cpp
@@ -165,6 +165,11 @@ llvm::cl::opt<bool> hack_mul_extended(
     "hack-mul-extended", llvm::cl::init(false),
     llvm::cl::desc("Avoid usage of OpSMulExtended and OpUMulExtended"));
 
+llvm::cl::opt<bool> hack_convert_to_float(
+    "hack-convert-to-float", llvm::cl::init(false),
+    llvm::cl::desc("Insert a dummy instruction after conversions to float to "
+                   "avoid driver optimization getting rid of the conversion"));
+
 llvm::cl::opt<bool>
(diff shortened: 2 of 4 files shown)


alan-baker:
SPIR-V describes `NoContraction` as only being applied to arithmetic instructions. We can't actually use it on conversions.
>Apply only to an [arithmetic instruction](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Arithmetic) to indicate the operation cannot be combined with another instruction to form a single operation. For example, if applied to an [OpFMul](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpFMul), that multiply can’t be combined with an addition to yield a fused multiply-add operation. Furthermore, such operations are not allowed to reassociate; e.g., add(a + add(b+c)) cannot be transformed to add(add(a+b) + c).

dneto0:
> perform rounded conversions

If the goal is to do a rounded conversion, then it should use a Round or RoundToEven extended instruction.

https://registry.khronos.org/SPIR-V/specs/1.0/GLSL.std.450.html#:~:text=Extended%20instructions%3A-,Round,-Result%20is%20the

dneto0:
> If the goal is to do a rounded conversion,

Whoops   float to int conversion is supposed to round to zero.

So it should be Trunc instead, I think.

rjodinchr:
libclc is doing this pattern [here](https://github.com/llvm/llvm-project/blob/main/libclc/generic/lib/gen_convert.py#L434).
The goal is to deviate from the initial value by doing this double conversion. Then the deviation is analysed to generate `rtz`, `rtn` and `rtp` rounding mode. libclc assumes that the standard conversion is round to even.

Nvidia, Intel, and AMD (using mesa and the old llvm compiler) all round to even when performing OpConvertFToS.
AMD (using the new ACO compiler) round to zero.

Isn't there a CTS somewhere supposed to check that those Vulkan implementations round OpConvertFToS as required in the specification?

I can avoid the optimization performed by at least intel and AMD by inserting a `FAdd` with zero between the 2 conversions. I will transform this PR to add a hack option to do that.

dneto0:
> Nvidia, Intel, and AMD (using mesa and the old llvm compiler) all round to even when performing OpConvertFToS.
AMD (using the new ACO compiler) round to zero.

> Isn't there a CTS somewhere supposed to check that those Vulkan implementations round OpConvertFToS as required in the specification?

I guess not. Ouch.

dneto0:
Confirmed.  Vulkan CTS does not cover these cases.  I'll file a bug.

alan-baker (approved):

PR #1143, opened 2023-06-26
Read kernel attributes and add them to non-semantic reflection information

rjodinchr:
reading the attrs
added kernel attrs as nonsemantic information
hacked kernel attribute information into nonsemantic information write annotations to metadata when possible
read attributes into spirv

This PR is just a clean rebase of #990

I have checked the implementation using a clean rebase of https://github.com/kpet/clvk/pull/489 on both swiftshader and nvidia. It fixes `test_api kernel_attributes`.

### DIFF
--- a/lib/AnnotationToMetadataPass.cpp
+++ b/lib/AnnotationToMetadataPass.cpp
@@ -1,0 +1,87 @@
+// Copyright 2022 The Clspv Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "llvm/IR/Constants.h"
+
+#include "AnnotationToMetadataPass.h"
+#include "Constants.h"
+
+using namespace llvm;
+
+PreservedAnalyses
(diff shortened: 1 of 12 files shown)


lib/AnnotationToMetadataPass.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+      ConstantArray *annotations_array =
+          dyn_cast<ConstantArray>(GV.getOperand(0));
+      for (auto &annotation_entry : annotations_array->operands()) {
+        ConstantStruct *annotation_struct =
+            dyn_cast<ConstantStruct>(annotation_entry.get());
+
+        auto op0 = annotation_struct->getOperand(0);
+        Function *entry_point = dyn_cast<Function>(
+            op0->getType()->isOpaquePointerTy() ? op0 : op0->getOperand(0));
```
alan-baker, line 44: This PR pre-dates the switch to opaque pointers. Please update to avoid these tests.

lib/SPIRVProducerPass.cpp
```diff
@@ -1549,6 +1566,24 @@ Type *SPIRVProducerPassImpl::CanonicalType(Type *type) {
     // Nothing
   } else if (type->getNumContainedTypes() != 0) {
     switch (type->getTypeID()) {
+    case Type::PointerTyID: {
```
alan-baker, line 1569: Is this needed? There is a test for is pointer type above. Is this also a relic of opaque pointer transitions?

alan-baker (changes requested):

alan-baker (commented):
There are also no tests for this PR, that should be fixed.

lib/SPIRVProducerPass.cpp
rjodinchr: This is a relic of opaque pointer, I have removed it

lib/AnnotationToMetadataPass.cpp
rjodinchr: I have clean this file

PR #1146, opened 2023-06-28
add option to set execution mode RoundingModeRTE

rjodinchr:


rjodinchr:
`CI-clvk-ubuntu-clang-debug ` is failing with the following reason:
```
[CLVK] supports_capability: Capability 4467 not yet mapped to a feature.
[CLVK] check_capabilities: Device does not support SPIR-V capability 4467 (spv::CapabilityRoundingModeRTE).
[CLVK] Missing support for required SPIR-V capabilities.
```
I think that is just not implemented in `swiftshader`. I am not sure how to deal with that. Ideas are welcome here.

rjodinchr:
clvk will need https://github.com/kpet/clvk/pull/563 to support the capability. But even with this commit `simple_test` is failing.

rjodinchr:
Maybe we should set the execution mode only if we generate `OpConvertUToF`/`OpConvertSToF`.

alan-baker (approved):

alan-baker:
It appears the logic for adding the capability is incorrect.

lib/SPIRVProducerPass.cpp
```diff
@@ -383,6 +383,16 @@ struct SPIRVProducerPassImpl {
...
     }
   }
+  bool hasConvertToF() { return HasConvertToF; }
+  void setConvertToF() {
+    if (!HasConvertToF && SpvVersion() >= SPIRVVersion::SPIRV_1_4 &&
+        (ExecutionModeRoundingModeRTE(RoundingModeRTE::fp16) ||
+         ExecutionModeRoundingModeRTE(RoundingModeRTE::fp32) ||
+         ExecutionModeRoundingModeRTE(RoundingModeRTE::fp64))) {
+      addCapability(spv::CapabilityRoundingModeRTE);
+      HasConvertToF = true;
```
alan-baker, line 393: The version could be dropped and the extension could be added if the version is <1.4

rjodinchr: I don't understand what extension you are talking about when version is <1.4

alan-baker: [SPV_KHR_float_controls](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_float_controls.html)

rjodinchr: Done

PR #1152, opened 2023-07-13
fix physicalStorageBuffers when a kernel is calling another kernel

rjodinchr:
We used to leave the calling conv of callinst of spir_kernel function as spir_kernel, while we had changed the function to be just a spir_function.
This used to lead to instcombine just getting rid of the call as (I think) it cannot find the function with this calling conv.

Set the calling conv to spir_func so that the call is not remove by mistake.

This patch is fixing some compiler tests of the CTS when running with physical addressing enabled.

### DIFF
--- a/lib/PhysicalPointerArgsPass.cpp
+++ b/lib/PhysicalPointerArgsPass.cpp
@@ -108,6 +108,12 @@ PreservedAnalyses clspv::PhysicalPointerArgsPass::run(Module &M,
     }
 
     F.setCallingConv(CallingConv::SPIR_FUNC);
+    for (auto &U : F.uses()) {
+      if (auto CI = dyn_cast<CallInst>(U.getUser())) {
+        CI->setCallingConv(CallingConv::SPIR_FUNC);
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


alan-baker:
Clspv has some larger issues with kernels calling kernels:
```
__attribute__((noinline))
kernel void bar(global int* in, global int* out) {
  *out = *in;
}

kernel void foo(global int *in, global int* out) {
  bar(in, out);
}
```

This leads to invalid IR even without physical storage buffers. Does this fix work more generally?

rjodinchr:
> This leads to invalid IR even without physical storage buffers. Does this fix work more generally?

No the issue is not the same. In the case you are talking about, the issue is about the function not being inline and we have an issue with the call Operator.

What I am trying to fix here is just a bug with physical storage buffers, where the call operator has just been removed without having inline the function.

We are already doing what I am added here in the other passes where we inline. It is just missing in this pass.

rjodinchr:
For the issue you mentioned, I think we would need something similar at what we did for physical addressing, but without the inlining nor the physical addressing part of it.

alan-baker (approved):

PR #1162, opened 2023-07-20
fix printf when reaching end of printf buffer

rjodinchr:
If we do not have the space to store all the arguments, we should not store them. It leads to memory corruption from out-of-bounds access on platform like lavapipe.

I thought of merging the tests for the printf id and the args, but we need both of them.
Here is an example to understand:
Considering we have a printf with 1 argument and a buffer size of 8 bytes (2 words).
The first word is for the counter of number of words written. We are left with one word.
When performing the printf, we will execute the atomic add, which will set the counter to 2 (1 for the printfid and 1 for the arg). If we do not store the printf id, the runtime won't be able to know whether we are trying to print 2 string without argument (each one needing 1 word) or if we are printing 1 string with 1 argument.

### DIFF
--- a/lib/PrintfPass.cpp
+++ b/lib/PrintfPass.cpp
@@ -100,11 +100,13 @@ void clspv::PrintfPass::DefinePrintfInstance(Module &M, CallInst *CI,
   auto *ConstantPrintfID = ConstantInt::get(Int32Ty, PrintfID);
   auto *ConstantAllocSize = ConstantInt::get(Int32Ty, AllocSizeInts);
   auto *ConstantBuffSize =
-      ConstantInt::get(Int32Ty, clspv::Option::PrintfBufferSize());
+      ConstantInt::get(Int32Ty, clspv::Option::PrintfBufferSize() / 4);
   const auto &DL = M.getDataLayout();
 
   auto *EntryBB = BasicBlock::Create(Ctx, "entry", Func);
-  auto *BodyBB = BasicBlock::Create(Ctx, "body", Func);
+  auto *CopyArgsBB = BasicBlock::Create(Ctx, "copy_args", Func);
+  auto *CopyPrintfIdBB = BasicBlock::Create(Ctx, "copy_printf_id", Func);
+  auto *TestPrintfIdBB = BasicBlock::Create(Ctx, "test_printf_id", Func);
   auto *ExitBB = BasicBlock::Create(Ctx, "exit", Func);
   IRBuilder<> IR{EntryBB};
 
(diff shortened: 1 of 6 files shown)


lib/PrintfPass.cpp
```diff
@@ -97,11 +97,13 @@ void clspv::PrintfPass::DefinePrintfInstance(Module &M, CallInst *CI,
   auto *ConstantNegOne = ConstantInt::get(Int32Ty, -1);
   auto *ConstantPrintfID = ConstantInt::get(Int32Ty, PrintfID);
   auto *ConstantAllocSize = ConstantInt::get(Int32Ty, AllocSizeInts);
+  auto *ConstantArgsSize = ConstantInt::get(Int32Ty, ArgsSize / 4);
   auto *ConstantBuffSize =
-      ConstantInt::get(Int32Ty, clspv::Option::PrintfBufferSize());
+      ConstantInt::get(Int32Ty, clspv::Option::PrintfBufferSize() / 4);
```
rjodinchr, line 102: The buffer size is used to compare it with the printf counter.
This counter counts the number of word (32bit element) written, thus we need to consider the size as a number of word

alan-baker (commented):
It feels odd to add an extra basic block. Both entry and header are calculating whether there is storage. It feels like they should be combined.

rjodinchr:
Printf needs to store both the printf id and the args.

If we have space for everything, we store everything, but if we just have space for the printf id, we need to store it and not the args. So in the end we have 2 checks. One to know whether we have space for the printf id, and then one for the args.

We could rearrange that, but we would still have 2 checks in the end I believe.

Here is what we have at the moment:
```
if (more than 1 word left in the buffer) {
  store printf id
  if (enough space for the args in the buffer) {
    store args
  }
}
```
As explained in the commit message, we really want to store the printf id if we have the space for it even if we do not have the space for the args.

About the test added, we are not checking the same thing in the entryBB and headerBB. In entryBB we are comparing against `atomicadd+1` and in headerBB against `atomicadd+2`.

rjodinchr:
I can reorganise the test so that we have only 1 test in the best case:
```
entry:
  cmp = enough space for everything
  br cmp, copy_args, test_printf_id
copy_args:
  store args
  br copy_printf_id
copy_printf_id:
  store printf_id
  br end
test_printf_id:
  cmp = enough space for printf id
  br cmp, copy_printf_id, end
end:
  return value
```
What do you think of that?

alan-baker:
> I can reorganise the tests so that we have only 1 test in the best case:
> 
> ```
> entry:
>   cmp = enough space for everything
>   br cmp, copy_args, test_printf_id
> copy_args:
>   store args
>   br copy_printf_id
> copy_printf_id:
>   store printf_id
>   br end
> test_printf_id:
>   cmp = enough space for printf id
>   br cmp, copy_printf_id, end
> end:
>   return value
> ```
> 
> What do you think of that?

What is the value of printf if we can't fit the args. Does CTS cover that case? It seems unnecessary. If we don't have space for everything just don't print. If that is not allowed by the spec then your suggestion is an improvement.

rjodinchr:
> What is the value of printf if we can't fit the args. Does CTS cover that case? It seems unnecessary. If we don't have space for everything just don't print. If that is not allowed by the spec then your suggestion is an improvement.

CTS does not cover that case. It is hard for it to do it because even so OpenCL has `CL_DEVICE_PRINTF_BUFFER_SIZE`, it does not specify what is in the buffer exactly, thus making it hard to know when the buffer should be full.

But the issue here is about the fact that any application could completely fill the buffer. Then the question is how does the runtime handle it?
In the current implementation, the runtime has the information of the number of words written in the buffer (stored in the first word of the buffer), the content of the buffer and the size of the buffer.
To know when to stop reading the buffer, the runtime is using the number of words written as long as it is smaller than the size of the buffer (meaning the buffer is not full).
But when the buffer is full, we have 2 possibilities:
- the last printf stored its last word in the last word of the buffer. In that case, it is easy and the runtime knows there is nothing after and it stops.
- the buffer is too small to store all the printf, thus we have a number of words written greater than the buffer size because whether or not we store something in the buffer, printf is atomicaly incrementing this counter (the number of words written). And we need to know when the last one is. As we do not have some kind of marker telling us that it is the end, as long as we did not reach the end of the buffer, we are looking for the printf id, and then we know how many argument we should expect. But if the first printf overflowing does not write its ID, it means that the runtime will read a random value in the buffer and will consider it as a printf id. Which can lead to various issues.

Please have a look at https://github.com/kpet/clvk/pull/585/files. There we have some tests for that. At the moment they are passing because we are lucky. Swiftshader zero allocates the buffer, thus allowing the test to pass. But running with lavapipe, it fails because we are missing this fix.

rjodinchr:
I think it also passes at the moment because the current implementation has an issue with the buffer size (fix here), making it write word out-of-bounds. It does not generate a bug at the moment, but could easily change if we change anything.

alan-baker (approved):

Issue #1185, opened 2023-08-17
Clspv crashes writing to nested objects in an array with -physical-storage-buffers


BukeBeyond:
```
struct V { int x, y; };
struct E { struct V v; };
struct S { struct E e[2]; };

kernel void Kernel(global struct S* s)
{
    #define Case1

    #ifdef Case1 // Fails
        s->e[0].v.x = 1;
        s->e[0].v.y = 1;
        s->e[1].v.x = 1;
    #endif

    #ifdef Case2 // OK, accesing array with pointer arithmetic
        global struct E* e0 = (global struct E*) (((long) s->e) + sizeof(struct E) * 0);
        e0->v.x = 1; 
        e0->v.y = 1;
        global struct E* e1 = (global struct E*) (((long) s->e) + sizeof(struct E) * 1);
        e1->v.x = 1;
        e1->v.y = 1;
    #endif

    #ifdef Case3 // OK
        s->e[0].v.x = 1;
        s->e[1].v.x = 1;
    #endif

    #ifdef Case4 // Fails
        s->e[0].v.y = 1;
        s->e[1].v.y = 1;
    #endif

    #ifdef Case5 // OK
        s->e[0].v.x = 1;
        s->e[1].v.y = 1;
    #endif

}
```
https://godbolt.org/z/7b4xcf3vP

PR #1186, opened 2023-08-17
Remove strict aliasing in SmallerTypeNotAliasing

### LINKED ISSUE #1184: Clspv crashes writing to a long field and a struct field in that order layout
```
struct E { int i0, i1; };

struct S
{
    long i; struct E e; // Fails
    // struct E e; long i; // OK
    // int i; struct E e; // OK
};

kernel void Kernel(global struct S* s)
{
    s->i = 1;
    s->e.i0 = 1;
    s->e.i1 = 1;
}
```
https://godbolt.org/z/bea6Y1rW1

rjodinchr:
strict aliasing is not needed in this function. On the contrary, it generates bug such as #1184.

No regression on CTS with swiftshader and nvidia

Fix #1184


rjodinchr:
I need to rework it, this is conflicting with my fix for #1185

rjodinchr:
I'll open a new one

Issue #1022, opened 2023-02-23
Provide a C ABI-compatible API?


nbaksalyar:
If I haven't missed anything, currently it is not really possible to use `clspv` if it is included as a library in projects using languages other than C++.

I think it should include an `extern "C"` API, exposing a function like `clspv::CompileFromSourceString`.
E.g., here's how it could work:

<details>

```c
// opaque result structure
struct clspv_result
{
    std::string *output_log_str;
    std::vector<uint32_t> *output;
};

extern "C" {

int compile_from_source_string(const char *program,
                               const char *options,
                               clspv_result *result)
{
    return clspv::CompileFromSourceString(std::string(program),
                                          std::string(),
                                          std::string(options),
                                          result->output,
                                          result->output_log_str);
}

clspv_result *clspv_result_alloc()
{
    clspv_result *res = new clspv_result;
    res->output = new std::vector<uint32_t>();
    res->output_log_str = new std::string();
    return res;
}

void clspv_result_free(clspv_result *res)
{
    delete res->output_log_str;
    delete res->output;
    delete res;
}

void clspv_result_get_output(clspv_result *res, uint32_t **output, size_t *output_size)
{
    *output = &*res->output->begin();
    *output_size = res->output->size();
}

void clspv_result_get_output_log(clspv_result *res, const char **output, size_t *output_size)
{
    *output = &(res->output_log_str->c_str())[0];
    *output_size = res->output_log_str->size();
}

}
```

</details>

Then, it can be packaged as a library and used from C, Rust, or any other language.

Issue #245, opened 2018-11-13
Error handling in CLSPV


alan-baker:
**This is not planned to be addressed in the near term. This is merely recording ideas to address the problem.**

CLSPV is a prototype compiler. It is not intended that it has complete OpenCL-C coverage at this point. Many of the passes are not robust to unexpected inputs. The result of this is that the compiler will crash frequently. It would be nice if the compiler would do this more gracefully. As #242 gets implemented, it is especially problematic for tools that embed clspv as a library running in the same process. These are ideas to address graceful errors. For a production quality compiler, clspv should handle the full breadth of acceptable language inputs and catch untranslatable cases in the frontend.

The following is based on the assumption that the majority of crashes occur in clspv-specific passes.

A possibility to handle errors more gracefully would be converting `llvm_unreachable` and `assert` calls into behaviour that clobbers the module, but allows the compilation to continue and produce an empty binary. An empty binary could be used to signal an error has occurred. There would be a large amount of mechanical work required to get all the passes to abort correctly.

An alternative to wiping out the module would be adding metadata to indicate the module is bad. Passes would then be modified to not operate on a module with this metadata. It does not eliminate the need for passes to abort, but it is potentially easier on the infrastructure than wiping out the module mid-pass. It would introduce a different problem instead of having a potentially invalid module getting caught in the verifier.

alan-baker:
This should also address #13

PR #1201, opened 2023-08-24
Introduce C API and Shared library

annestrand:
Introduce C API and shared library building of Clspv.
Additionally making ReflectionParser a static lib so that it's shared by both CLI and Clspv shared lib.

This PR addresses #1022


rjodinchr:
Note that the C++ API does not handle errors nicely (#245). This would be the case as well for any C API.

I think we will need some testing here. That would be in the `kokoro` folder I believe.
@alan-baker do you need to do things outside of this github repository to have new tests working in github CI?

annestrand:
Agreed - given that this just wraps around C++ API.

Thanks, will look into adding testing. Also, I somehow missed the `.clang-format` file ... d'oh!
Will fix that.

rjodinchr:
Tests are failing because `clspv-reflection` has been renamed `clspv-reflection-cli` I believe. Is it really necessary? It would mean that this PR needs to update every test using it. But I would personally prefer not to change the name if this is possible.

annestrand:
Ah sorry - did not account for that. Not necessary in this case - I can change it back.

I do also think there were some format errors:
```
+ ./utils/check_code_format.sh
Thu Aug 24 16:00:23 UTC 2023: Check formatting...
Found formatting errors!
```

rjodinchr (commented):
I feel like we should not create an `api` folder for it.
We could merge every part into the existing structure:
- `api/CMakeLists.txt` merge in `lib/CMakeLists.txt`
- `api/Clspv.cpp` merge in `lib/Compiler.cpp`
- `api/Clspv.h` merge in `include/clspv/Compiler.h`
- `api/README.md` moves to `docs`, with a mention in the top README, and maybe mention of the C++ API.

@alan-baker will have the final word on that, so wait for his point-of-view before changing everything.

api/Clspv.h
```diff
@@ -0,0 +1,85 @@
...
+                                           const char *spv_env_str,
+                                           uint32_t *spv_binary,
+                                           size_t spv_words,
+                                           const char **output_data);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // CLSPV_API_CLSPV_H_
```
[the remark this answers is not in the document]
annestrand: Done 👍

tools/reflection/main.cpp
```diff
@@ -60,19 +60,6 @@ int main(const int argc, const char *const argv[]) {
         std::cerr << "Error: invalid target env: " << argv[i] << "\n";
         return -1;
       }
-      switch (env) {
```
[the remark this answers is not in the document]
annestrand: Got it - I went ahead and dropped the commit for this.

tools/reflection/ReflectionParser.cpp
```diff
@@ -329,16 +330,32 @@ ReflectionParser::ParseInstruction(const spv_parsed_instruction_t *inst) {
 
 namespace clspv {
 
-bool ParseReflection(const std::vector<uint32_t> &binary, spv_target_env env,
+bool ParseReflection(const uint32_t* spv_binary,
```
alan-baker, line 333: This tool is deprecated and should not be used. APIs should parse the reflection data into their own internal data structures.

Please revert changes to this directory.

api/Clspv.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return CLSPV_ERROR;
+  }
+
+  {
+    std::scoped_lock<std::mutex> sl((*container)->containerMutex);
+    if ((*container)->reflectionData.tellp() != std::streampos(0)) {
+      (*container)->reflectionData.clear();
+    }
+
+    if (!clspv::ParseReflection(spv_binary, spv_words, spvEnv,
```
alan-baker, line 136: This is not the right function to use. It has intentionally not been kept up-to-date with interface changes. Users of clspv are responsible for parsing the embedded reflection data in the SPIR-V.

api/README.md
```diff
@@ -0,0 +1,166 @@
+# Clspv - C API wrapper
```
alan-baker, line 1: This belongs in docs as `C_API.md` or a similar name.

CMakeLists.txt
```diff
@@ -183,6 +187,9 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/lib)
 # Bring in our tools folder
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/tools)
 
+# Bring in our api folder
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/api)
```
alan-baker, line 191: I agree with Romaric that a new directory is not desirable.

api/Clspv.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <string>
+#include <vector>
+
+#include "Clspv.h"
+#include "clspv/Compiler.h"
+#include "reflection/ReflectionParser.h"
+#include "spirv-tools/libspirv.hpp"
+
+// Global API Mutex
+static std::mutex g_clspvMutex;
```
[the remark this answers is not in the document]
annestrand: It's technically different in that we now have this container/context object that wraps the C++ items. 
Hence why thread safety should be considered here.

Did you mean user/app side here? If so, yes we could explicitly tell users (say in the C_API.md doc) that they are responsible for synchronization here. I prefer the existing approach, but I'm open to your suggestion if we mention this in docs.

api/Clspv.h
```diff
@@ -0,0 +1,85 @@
...
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Create Clspv container.
+//
+// |container| - Opaque handle to created Clspv container.
+EXPORT ClspvError clspvCreateContainer(ClspvContainer *container);
```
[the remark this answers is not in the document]
annestrand: My oversight there perhaps 😅 - Will update and push.

tools/reflection/ReflectionParser.cpp
annestrand: My mistake - I did not know this tool was deprecated. Will update and push.

api/Clspv.cpp
annestrand: Ack.

api/README.md
annestrand: Ack.

CMakeLists.txt
annestrand: Ack.

api/Clspv.h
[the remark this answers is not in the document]
annestrand: Actually now I remember... we need to pass container ptr by reference (otherwise its just a copy).

We could modify create function to instead return ptr to created container. That might be a better approach.

rjodinchr (commented):
`ClspvContainer` is not needed anymore. It made sense when you had to pass data between the compiler function and the reflection parser one. But now it is not used as you only have the compiler function.

Removing it will largely reduce the size of this PR. Then you might consider moving `include/clspv/Clspv.h` in `include/clspv/Compiler.h`, and `lib/Clspv.cpp` in `lib/Compiler.cpp`

annestrand:
@rjodinchr - Sure, I can merge the separate header and source to `Compiler.h/.cpp` respectively.

However, on removing the container object, I don't think there is a clean way to free a dyn. alloced std::vector or std::string from the user side here when they/user no longer needs either the spv code or the build log respectively (since `clspv::CompileFromSourcesString` requires these C++ objs).

I think the merging to `Compiler.h/.cpp` should help reduce the PR size here.

rjodinchr:
> However, on removing the container object, I don't think there is a clean way to free a dyn. alloced std::vector or std::string from the user side here when they/user no longer needs either the spv code or the build log respectively (since `clspv::CompileFromSourcesString` requires these C++ objs).

This C-API is already coying `programs` (the C input) to `vprograms` (C++ `std::vector<std::string>`). I think it would need to do something similar for the outputs: meaning to copy the `std::vector`/`std::string` into buffers allocated with `malloc` that would be the user's responsibility to free.

We may provide a helper function to free those objects (maybe in the header so that the content is clearly visible?). But I think it would be important to have both versions in the doc (manual free, and using the helper function).

annestrand:
That's a fair point. Now, if that same `clspv::CompileFromSourcesString` had an overload at some point to allow passing C type strings, then we would not copy at all in the container approach I believe. Benefit of this overload could be if clspv is used inside of an OpenCL driver impl. where maintainer would not need to copy input strings to a C++ container.

That's out of scope for this PR, so back to your suggestion - I'm okay with removal of the container here and go with your approach. My only heartburn was the extra memcpy for the outputs, but I think I can look past that here from your point above.

docs/C_API.md
```diff
@@ -0,0 +1,137 @@
...
+                        "}                                                          \n";
+
+    // Compile
+    size_t outputSpvSize = 0;
+    char* outputSpv = nullptr;
+    char* outputBuildLog = nullptr;
+    ClspvError ret = clspvCompileFromSourcesString(1,
+                                                   NULL,
+                                                   (const char**)&clSrc,
+                                                   "--output-format=spv ",
```
rjodinchr, line 29: This is the default value, thus not really needed

include/clspv/Compiler.h
```diff
@@ -44,4 +45,59 @@ int CompileFromSourcesString(const std::vector<std::string> &programs,
...
+    size_t *output_binary_size, char **output_log);
+
+// Frees the output memory from clspvCompileFromSourcesString
+//
+// |output_binary|      - Handle to spv
+// |output_log|         - Handle to compiler build log
+static inline void clspvFreeOutputBuildObjs(char *output_binary,
+                                            char *output_log) {
+  if (output_binary) {
+    free(output_binary);
```
rjodinchr, line 90: free of nullptr is allowed, we should not check for it:
> If ptr is a null pointer, the function does nothing.
https://en.cppreference.com/w/c/memory/free

lib/Compiler.cpp
```diff
@@ -662,7 +663,7 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     pm.addPass(clspv::SimplifyPointerBitcastPass());
     pm.addPass(clspv::ReplacePointerBitcastPass());
     pm.addPass(llvm::createModuleToFunctionPassAdaptor(llvm::DCEPass()));
-
```
rjodinchr, line 665: This should not be part of this PR

include/clspv/Compiler.h
annestrand: Ack

docs/C_API.md
annestrand: Ack

rjodinchr (approved):

annestrand:
Hmm, not sure what went wrong with `CI-ubuntu-gcc-debug` run? Tried clicking "Details" to get log but I get a DNS error.

alan-baker:
I've restarted that build. It was an internal infrastructure error so it couldn't publish the logs.

alan-baker (approved):

rjodinchr:
Here what I found in the logs:
```
[2192/2195] Linking CXX executable bin/clspv
FAILED: bin/clspv 
: && /usr/bin/g++ -fPIC -fno-semantic-interposition -fvisibility-inlines-hidden -Werror=date-time -fno-lifetime-dse -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wno-missing-field-initializers -pedantic -Wno-long-long -Wimplicit-fallthrough -Wno-class-memaccess -Wno-redundant-move -Wno-pessimizing-move -Wno-noexcept-type -Wdelete-non-virtual-dtor -Wsuggest-override -Wno-comment -Wno-misleading-indentation -fdiagnostics-color -fno-rtti -fno-exceptions -Werror=unused-variable -Werror=switch -g  tools/driver/CMakeFiles/clspv.dir/main.cpp.o -o bin/clspv  lib/libclspv_core.a  third_party/llvm/lib/libclangCodeGen.a  third_party/llvm/lib/libLLVMCoverage.a  third_party/llvm/lib/libLLVMLTO.a  third_party/llvm/lib/libLLVMPasses.a  third_party/llvm/lib/libLLVMCodeGen.a  third_party/llvm/lib/libLLVMObjCARCOpts.a  third_party/llvm/lib/libLLVMCoroutines.a  third_party/llvm/lib/libLLVMIRPrinter.a  third_party/llvm/lib/libLLVMipo.a  third_party/llvm/lib/libLLVMLinker.a  third_party/llvm/lib/libLLVMInstrumentation.a  third_party/llvm/lib/libLLVMVectorize.a  third_party/llvm/lib/libLLVMBitWriter.a  third_party/llvm/lib/libLLVMCodeGenTypes.a  third_party/llvm/lib/libLLVMTarget.a  third_party/llvm/lib/libLLVMExtensions.a  third_party/llvm/lib/libclangFrontend.a  third_party/llvm/lib/libclangDriver.a  third_party/llvm/lib/libLLVMWindowsDriver.a  third_party/llvm/lib/libclangParse.a  third_party/llvm/lib/libLLVMOption.a  third_party/llvm/lib/libclangSerialization.a  third_party/llvm/lib/libclangSema.a  third_party/llvm/lib/libclangAnalysis.a  third_party/llvm/lib/libclangASTMatchers.a  third_party/llvm/lib/libLLVMFrontendHLSL.a  third_party/llvm/lib/libclangEdit.a  third_party/llvm/lib/libclangAST.a  third_party/llvm/lib/libLLVMFrontendOpenMP.a  third_party/llvm/lib/libLLVMScalarOpts.a  third_party/llvm/lib/libLLVMAggressiveInstCombine.a  third_party/llvm/lib/libLLVMInstCombine.a  third_party/llvm/lib/libLLVMTransformUtils.a  third_party/llvm/lib/libLLVMAnalysis.a  third_party/llvm/lib/libLLVMProfileData.a  third_party/llvm/lib/libLLVMSymbolize.a  third_party/llvm/lib/libLLVMDebugInfoPDB.a  third_party/llvm/lib/libLLVMDebugInfoMSF.a  third_party/llvm/lib/libLLVMDebugInfoBTF.a  third_party/llvm/lib/libLLVMDebugInfoDWARF.a  third_party/llvm/lib/libLLVMObject.a  third_party/llvm/lib/libLLVMIRReader.a  third_party/llvm/lib/libLLVMAsmParser.a  third_party/llvm/lib/libLLVMMCParser.a  third_party/llvm/lib/libLLVMTextAPI.a  third_party/llvm/lib/libLLVMBitReader.a  third_party/llvm/lib/libLLVMCore.a  third_party/llvm/lib/libLLVMRemarks.a  third_party/llvm/lib/libLLVMMC.a  third_party/llvm/lib/libLLVMBinaryFormat.a  third_party/llvm/lib/libLLVMDebugInfoCodeView.a  third_party/llvm/lib/libclangLex.a  third_party/llvm/lib/libclangBasic.a  third_party/llvm/lib/libLLVMTargetParser.a  third_party/llvm/lib/libclangSupport.a  third_party/llvm/lib/libLLVMBitstreamReader.a  third_party/llvm/lib/libLLVMSupport.a  third_party/llvm/lib/libLLVMDemangle.a  -lrt  -ldl  -lpthread  -lm  /usr/lib/x86_64-linux-gnu/libz.so && :
/usr/bin/ld: failed to set dynamic section sizes: Memory exhausted
collect2: error: ld returned 1 exit status
```

annestrand:
Yep, looks like shared lib building is adding some mem strain to that build machine (for GCC debug at least).

Do we know what the specs are for that machine/container? I would like to try reproducing it locally (I might have a workaround).

annestrand:
Okay, looks like GCC debug is clean. Now I'm getting new CI issue here - `CI-clvk-ubuntu-clang-debug`:
```
-- Using SPIR-V Headers from
          /tmpfs/src/github/clvk/external/SPIRV-Headers
CMake Error at external/SPIRV-LLVM-Translator/CMakeLists.txt:82 (find_package):
  Could not find a configuration file for package "LLVM" that is compatible
  with requested version "17.0.0".

  The following configuration files were considered but not accepted:

    /tmpfs/src/github/clvk/build/external/clspv/third_party/llvm/lib/cmake/llvm/LLVMConfig.cmake, version: 18.0.0git
    /bin/clang-8.0.0/lib/cmake/llvm/LLVMConfig.cmake, version: 8.0.0
```
@rjodinchr - Any clues for this one?

alan-baker:
> Okay, looks like GCC debug is clean. Now I'm getting new CI issue here - `CI-clvk-ubuntu-clang-debug`:
> 
> ```
> -- Using SPIR-V Headers from
>           /tmpfs/src/github/clvk/external/SPIRV-Headers
> CMake Error at external/SPIRV-LLVM-Translator/CMakeLists.txt:82 (find_package):
>   Could not find a configuration file for package "LLVM" that is compatible
>   with requested version "17.0.0".
> 
>   The following configuration files were considered but not accepted:
> 
>     /tmpfs/src/github/clvk/build/external/clspv/third_party/llvm/lib/cmake/llvm/LLVMConfig.cmake, version: 18.0.0git
>     /bin/clang-8.0.0/lib/cmake/llvm/LLVMConfig.cmake, version: 8.0.0
> ```
> 
> @rjodinchr - Any clues for this one?

That's a problem with LLVM-SPIRV-Translator not being up to date with the latest LLVM. Not your problem.

Issue #1204, opened 2023-08-28
UNREACHABLE executed at lib/BitcastUtils.cpp:1218


zwang20:
Compiling the attached source with the following command line

```
unzip a.zip
clspv -x ir -cl-mad-enable -DDEVICE_INFO=32 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=3 -DDEV_VER_MINOR=0 -D_OPENCL_COMPILER -DPLAINTEXT_LENGTH=19 -DUCS_2 -DCONST_CACHE_SIZE=65536 -DASCII -DASCII -DLOC_0=-1 -DLOC_1=-1 -DLOC_2=-1 -DLOC_3=-1 -DNUM_INT_KEYS=1 -DIS_STATIC_GPU_MASK=0 -cl-single-precision-constant -cl-kernel-arg-info -rounding-mode-rte=16,32,64 -rewrite-packed-structs -std430-ubo-layout -decorate-nonuniform -arch=spir --use-native-builtins=acos,acosh,acospi,asin,asinh,asinpi,atan,atan2,atan2pi,atanh,atanpi,ceil,copysign,fabs,fdim,floor,fma,fmax,fmin,frexp,half_rsqrt,half_sqrt,isequal,isfinite,isgreater,isgreaterequal,isinf,isless,islessequal,islessgreater,isnan,isnormal,isnotequal,isordered,isunordered,ldexp,mad,rint,round,rsqrt,signbit,sqrt,tanh,trunc, -spv-version=1.6 -max-pushconstant-size=256 -max-ubo-size=65536 -global-offset -long-vector -module-constants-in-storage-buffer -cl-arm-non-uniform-work-group-size -enable-printf -printf-buffer-size=1048576 a.bc
```

results in the following output

```
Err: SrcTy = i32 - DstTy = i8 - Ty = i32 - CstVal = 32
Unexpected offset for type in GetIdxsForTyFromOffset
UNREACHABLE executed at /path/to/clvk/external/clspv/lib/BitcastUtils.cpp:1218!
```

[a.zip](https://github.com/google/clspv/files/12450111/a.zip)

zwang20:
This particular example comes from `john --test --format=oldoffice-opencl`

zwang20:
This bug appears to be fixed as of 619ad2d233afe0ff04b9162d46c32b4ed00e8fd2

PR #1205, opened 2023-08-28
fix runOnImplicitCasts

rjodinchr:
when src_gep indices are all zeros, just use src_gep pointer, and don't try to rework the indices which can lead to unsupported cases

Ref #1204

### DIFF
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -574,19 +574,29 @@ bool clspv::SimplifyPointerBitcastPass::runOnImplicitCasts(Module &M) const {
         cast<GetElementPtrInst>(inst_gep->getPointerOperand());
 
     IRBuilder<> Builder{inst_gep};
-    uint64_t CstVal;
-    Value *DynVal;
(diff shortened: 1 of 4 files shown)


rjodinchr:
Maybe it would be even easier/better to just get read of all geps with all zero indices? I am not sure.

alan-baker:
> Maybe it would be even easier/better to just get read of all geps with all zero indices? I am not sure.

Something in the flow needs to reconstruct the appropriate GEP for code generation. Where would that happen?

rjodinchr:
> > Maybe it would be even easier/better to just get read of all geps with all zero indices? I am not sure.
> 
> Something in the flow needs to reconstruct the appropriate GEP for code generation. Where would that happen?

We could get rid of them in `SimplifyPointerBitcast` during the whole pipeline. And then just before `SPIRVProducerPass`, we could add the one really needed with `SimplifyPointerBitcast::runOnImplicitGEP` (that would have been removed from `SimplifyPointerBitcast` obviously).

I need to think about it more. Let's consider this fix first and we will see what to do next.

alan-baker (approved):

PR #1206, opened 2023-08-28
fix runOnPHIFromGEP

rjodinchr:
When a gep is used by 2 phis, make sure to process it only once.

Ref #1204

### DIFF
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -759,6 +759,7 @@ bool clspv::SimplifyPointerBitcastPass::runOnPHIFromGEP(Module &M) const {
   bool changed = false;
   DenseMap<Value *, Type *> type_cache;
 
+  DenseSet<GetElementPtrInst *> Seen;
   SmallVector<std::pair<GetElementPtrInst *, Type *>> Worklist;
   for (auto &F : M) {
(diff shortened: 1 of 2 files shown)


lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -759,7 +759,7 @@ bool clspv::SimplifyPointerBitcastPass::runOnPHIFromGEP(Module &M) const {
   bool changed = false;
   DenseMap<Value *, Type *> type_cache;
 
-  SmallVector<std::pair<GetElementPtrInst *, Type *>> Worklist;
+  DenseMap<GetElementPtrInst *, Type *> Worklist;
```
alan-baker, line 762: This makes iteration non-deterministic. Since we iterate through the worklist below this might change code generation.

Suggest using either UniqueVector or an explicit parallel map.

alan-baker (changes requested):

lib/SimplifyPointerBitcastPass.cpp
alan-baker: Let's say there is a block with 2 phis for pointers we are processing. Is it not possible modifications will occur that change the relative order of instructions based on which phi is processed first?

alan-baker (approved):

Issue #1202, opened 2023-08-25
sampler using unnormalised coordinates and 3d images


rjodinchr:
A sampler using unnormalised coordinates cannot be used to read/write 3d images:
```
[CLVK] Validation Error: [ VUID-vkCmdDispatch-None-02702 ] Object 0: handle = 0x625f640000000058, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; Object 1: handle = 0x7cd292000000004f, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 2: handle = 0x2a7f70000000053, type = VK_OBJECT_TYPE_SAMPLER; | MessageID = 0x4fc13051 | vkCmdDispatch: Descriptor set VkDescriptorSet 0x625f640000000058[] Image View VkImageView 0x7cd292000000004f[], type: VK_IMAGE_VIEW_TYPE_3D in Descriptor in binding #0 index 0is used by VkSampler 0x2a7f70000000053[]. The Vulkan spec states: If the VkPipeline
object bound to the pipeline bind point used by this command accesses a VkSampler object that uses unnormalized coordinates, that sampler must not be used to sample from any VkImage with a VkImageView of the type VK_IMAGE_VIEW_TYPE_3D, VK_IMAGE_VIEW_TYPE_CUBE, VK_IMAGE_VIEW_TYPE_1D_ARRAY, VK_IMAGE_VIEW_TYPE_2D_ARRAY or VK_IMAGE_VIEW_TYPE_CUBE_ARRAY, in any shader stage (https://www.khronos.org/registry/vulkan/specs/1.3-khr-extensions/html/vkspec.html#VUID-vkCmdDispatch-None-02702)
```
Here is my current plan to fix it:
- literal sampler inside a kernel, example:
  - `const sampler_t my_sampler = CLK_ADDRESS_CLAMP | CLK_FILTER_NEAREST | CLK_NORMALIZED_COORDS_FALSE;`
  - In that case we should be able to detect if the sampler is used to read/write 3d images. Then we can rework everything in the compiler to use a normalise sampler and translate the coordinates accordingly.
- as an kernel argument, example:
  - `kernel void foo(read_only image3d image, sampler_t sampler)`
  - This case is a bit more complex:
    - The host API doesn't know for which image the sampler will be used.
      - The compiler needs to expose whether the sampler is used to access 3d images or not.
    - The compiler doesn't know whether the sampler expects normalised of unnormalised coordinates.
[...]

PR #1215, opened 2023-08-31
always use normalized sampler to read image3d with literal sampler

rjodinchr:
We cannot use unormalized sampled with 3D images.

This commit intend to fix that for literal sampler (static sampler defined in the kernel).

When detecting a case of reading a image3d with a literal sampler using unnormalized coordinate:
- create a equivalent sampler using normalized coordinate
- normalized the coordinate

Ref #1202

### DIFF
--- a/docs/OpenCLCOnVulkan.md
+++ b/docs/OpenCLCOnVulkan.md
@@ -823,6 +823,13 @@ describe:
 
 All supported.
 
+Reading 3D images with a unormalized sampler is not allowed in Vulkan. Thus
+OpenCL source code performing this operation is emulated by normalizing the
+coordinate before reading the image using a normalized sampler.
+
+The normalization of the coordinate can introduce an accuracy issue as the
+floating point division used is not correctly rounded.
+
 #### cl_khr_subgroups extension
 
 The OpenCL extension `cl_khr_subgroups` requires SPIR-V 1.3 or greater and
--- a/lib/BuiltinsEnum.h
+++ b/lib/BuiltinsEnum.h
@@ -29,6 +29,7 @@ enum BuiltinType : unsigned int {
   kSpirvCopyMemory,
   kClspvSamplerVarLiteral,
   kClspvCompositeConstruct,
+  kClspvGetImageSizes,
   kType_Clspv_End,
 
   kType_Async_Start,
--- a/lib/BuiltinsMap.inc
+++ b/lib/BuiltinsMap.inc
@@ -1117,6 +1117,7 @@ static std::unordered_map<const char *, Builtins::BuiltinType, cstr_hash,
         {"clspv.sampler_var_literal", Builtins::kClspvSamplerVarLiteral},
         {"clspv.composite_construct", Builtins::kClspvCompositeConstruct},
 
+        {"clspv.get_image_sizes", Builtins::kClspvGetImageSizes},
 };
 
 #endif // CLSPV_LIB_BUILTINSMAP_INC_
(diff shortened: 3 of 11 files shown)


alan-baker:
This one makes me nervous for two reasons:
1. Sampler arguments are unhandled
2. It feels like there might be a precision issue for large images where the wrong coordinate is sampled, but I haven't done all the math.

rjodinchr:
Sampler arguments will be handled in another PR. I am working on it, it is almost ready. But it will require a spirv-headers PR as well.

I understand your concern with precision issues for large images. But I don't see a good way to handle it right now. What is for sure, is that the current implementation is not allowed by the VVL.

alan-baker:
I agree that I'm unsure of how else to handle them. Let's at least add note in the OpenCLOnVulkan.md.

rjodinchr:
I am not even sure there is an issue here.
The image size is at best stored in a i64, and most probably far smaller than that considering vulkan image size limits.
But even considering that, what we want is to differentiate `n / 2^64` with ` (n+1)/ 2^64`.
Floating points go far below `2^-64` with good precision (enough to differentiate them).
I might be making a mistake in this reasoning, but I don't see what precision issue we could have considering the real size of images in vulkan.

Thus, I am not sure what to add to the doc really.

rjodinchr:
Thinking about it again, it feels like the issue is not for small indexes, but for bigger ones.
Considering an image of size n, we might have an issue with index `n-x` (with `x` relatively small), as it would turn out to be `(n-x)/n` for which floating point might lack precision.

alan-baker:
> I am not even sure there is an issue here. The image size is at best stored in a i64, and most probably far smaller than that considering vulkan image size limits. But even considering that, what we want is to differentiate `n / 2^64` with ` (n+1)/ 2^64`. Floating points go far below `2^-64` with good precision (enough to differentiate them). I might be making a mistake in this reasoning, but I don't see what precision issue we could have considering the real size of images in vulkan.
> 
> Thus, I am not sure what to add to the doc really.

I guess I'd suggest a note on how they are implemented. My worry is that the floating point division can introduce error because it is not correctly rounded.

rjodinchr:
> > I am not even sure there is an issue here. The image size is at best stored in a i64, and most probably far smaller than that considering vulkan image size limits. But even considering that, what we want is to differentiate `n / 2^64` with ` (n+1)/ 2^64`. Floating points go far below `2^-64` with good precision (enough to differentiate them). I might be making a mistake in this reasoning, but I don't see what precision issue we could have considering the real size of images in vulkan.
> > Thus, I am not sure what to add to the doc really.
> 
> I guess I'd suggest a note on how they are implemented. My worry is that the floating point division can introduce error because it is not correctly rounded.

Make sense, I'll add something about it.

docs/OpenCLCOnVulkan.md
```diff
@@ -823,6 +823,13 @@ describe:
 
 All supported.
 
+Reading 3D images with a unormalised sampler is not allowed in SPIR-V. Thus
+OpenCL source code performing this operation is emulated by normalising the
+coordinate before reading the image using a normalised sampler.
+
+The normalisation of the coordinate can introduce an accuracy issue as the
+floatint point division used is not correctly rounded.
```
alan-baker, line 831: Nits: switch to normalize for consistency with the rest of the codebase. Also, unormalized 3D images are not allowed in Vulkan (instead of SPIR-V).

alan-baker (commented):
Just nits, otherwise looks good.

alan-baker (approved):

lpavank:
@rjodinchr/@alan-baker  one question on the limitation "Reading 3D images with a unormalized sampler is not allowed in Vulkan". Is there a need/plan to get this addressed in Vulkan specification (unless there is a reason not to)? 
Just wanted to know because like discussed some of these work arounds could/may have certain limitations and if not limitations, might be more code to maintain/keep track of on compiler side.

thoughts?

alan-baker:
I don't expect movement on this in Vulkan. I would expect it's not very widely used if they never felt the need to add it.

lpavank:
Agreed on "not very widely used" comment.

Issue #1231, opened 2023-09-14
clang generates definitions of sqrt


alan-baker:
https://reviews.llvm.org/D156743 added definitions of float-based sqrt to the opencl header. This leads to problems in clspv:
* long vectors are always present
* cllib linking is broken

Short term solution is to enable long vectors and strip builtin definitions prior to linking.

PR #1232, opened 2023-09-14
Update deps

alan-baker:
Contributes to #1231

* Update llvm, spirv-headers, and spirv-tools
* Clang defines sqrt for float in opencl-c-base.h now
  * need to strip it to facilitate linking with cllib
  * need to always enable long vector lowering
    * xfail'd some tests for now
* Add missing dependency for header generation on opencl-c-base.h


dneto0 (approved):

rjodinchr (approved):

alan-baker:
clvk bot seems to fail because the spirv translator is still requiring 17.

rjodinchr:
> clvk bot seems to fail because the spirv translator is still requiring 17.

I'll take care of updating everything on clvk since soon.

PR #1234, opened 2023-09-20
Implement read of 3D images with unnormalised sampler

rjodinchr:
We need to know where the sampler is coming from, so we need to inline every function containing a read image of a 3d image with non literal sampler.

This PR depends on KhronosGroup/SPIRV-Headers#377
We use the information from the sampler mask to know if you need to use the original coordinates or the one we have normalised.

Like for channel_image_order and channel_image_data_type we use metadata to pass information through the pipeline.


lib/SamplerUtils.cpp
```diff
@@ -13,19 +13,53 @@
...
+  }
+
+  return B.CreateFDiv(Coord, ImgDimFP);
+}
 
-  return B.CreateFDiv(Coord, ImgSizes);
+bool isReadImage3DWithNonLiteralSampler(CallInst *call) {
+  DenseMap<Value *, Type *> cache;
+  auto Name = call->getCalledFunction()->getName();
+  if (Name.contains("read_image") && Name.contains("ocl_sampler")) {
```
alan-baker, line 49: Prefer using the builtins lookup here. Also, I think infer type for the image is unnecessary since it is never a pointer anymore.

lib/AutoPodArgsPass.cpp
```diff
@@ -144,7 +178,8 @@ void clspv::AutoPodArgsPass::runOnFunction(Function &F) {
       clspv::Option::ClusterPodKernelArgs() && support_16bit_pc &&
       support_8bit_pc && fits_push_constant &&
       !clspv::UsesGlobalPushConstants(M) && !contains_array &&
-      !contains_image_channel_getter;
+      !contains_image_channel_getter &&
+      !contains_read_image_3d_non_literal_sampler;
```
alan-baker, line 182: What are the static push constant size requirements up to now? I notice there is no uniform fallback for these samplers in the reflection instructions. Should there be?

alan-baker: Separately, I would think that since 3d non-literally sampled images get inlined, this would be amenable to per-kernel push constants. Why is it disallowed?

rjodinchr: > What are the static push constant size requirements up to now? I notice there is no uniform fallback for these samplers in the reflection instructions. Should there be?

I don't know about the push constant size requirement, could you tell me more?
I did not add a fallback in the reflection. I think it is better to add it when we have a real need for it. We have added them for image metadata and they are still not used. Not sure if they will ever be used. I did not want to add it without having a real use-case for it.

> Separately, I would think that since 3d non-literally sampled images get inlined, this would be amenable to per-kernel push constants. Why is it disallowed?

I did it without really thinking about it. It is just that for image metadata we have it disallowed, and it felt pretty much to be in a similar scenario.
Here is the justification I had for the image metadata from https://github.com/google/clspv/pull/882:
> Do not use per-kernel push constant interfaces if the kernel call
either of the channel functions. We need the pod kernel arguments when
we redeclare globalpushconstant to add info about image channel.

I think it still applies here.

lib/SamplerUtils.cpp
rjodinchr: Done

lib/AutoPodArgsPass.cpp
alan-baker: There is an option that clvk uses to pass the push constant size to clspv. Vulkan only requires 128 bytes. So disregarding arguments, we use push constants for a variety of things. How close are we to using 128 bytes? The samplers are variable I know, but if we can only fit one or two into the minimum size we should have a more robust solution available.

rjodinchr: > How close are we to using 128 bytes?

I have no idea, but I have never had any issues with that yet.

clvk seems to be setting this value to the maximum possible: https://github.com/kpet/clvk/blob/main/src/device.cpp#L556

alan-baker: When I look at PushConstant.{cpp,h}, I see we could be using 120 bytes before this change. Most of that size is to support non-uniform workgroups. It feels like this problem is not as far away as we thought. Don't have to fix it now, but please file an issue to track running out of push constant space.

rjodinchr: I have created #1235 to track it

alan-baker:
Please rebase this on top of the headers, I've merged the necessary changes.

rjodinchr:
Amber does not have the new spirv-headers yet.

alan-baker (approved):

Issue #1222, opened 2023-09-06
UNREACHABLE executed at lib/BitcastUtils.cpp:1237


zwang20:
Compiling the attached source with the following command line
```
unzip a.zip
clspv -cl-mad-enable -DDEVICE_INFO=32 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=3 -DDEV_VER_MINOR=0 -D_OPENCL_COMPILER -DKEYLEN=64 -DSALTLEN=64 -DOUTLEN=32 -DSAFETY_FACTOR=160 -cl-single-precision-constant -cl-kernel-arg-info -rounding-mode-rte=16,32,64 -rewrite-packed-structs -std430-ubo-layout -decorate-nonuniform -arch=spir --use-native-builtins=acos,acosh,acospi,asin,asinh,asinpi,atan,atan2,atan2pi,atanh,atanpi,ceil,copysign,fabs,fdim,floor,fma,fmax,fmin,frexp,half_rsqrt,half_sqrt,isequal,isfinite,isgreater,isgreaterequal,isinf,isless,islessequal,islessgreater,isnan,isnormal,isnotequal,isordered,isunordered,ldexp,mad,rint,round,rsqrt,signbit,sqrt,tanh,trunc, -spv-version=1.6 -max-pushconstant-size=256 -max-ubo-size=65536 -global-offset -long-vector -module-constants-in-storage-buffer -cl-arm-non-uniform-work-group-size -enable-printf -printf-buffer-size=1048576 -x ir a.bc
```
results in the following output
```
Err: SrcTy = i32 - DstTy = i8 - Ty = i32 - CstVal = 16
Unexpected offset for type in GetIdxsForTyFromOffset
UNREACHABLE executed at /path/to/clvk/external/clspv/lib/BitcastUtils.cpp:1237
```
[a.zip](https://github.com/google/clspv/files/12531902/a.zip)

rjodinchr:
With the 4 PRs referenced above, this test is now failing because it requires physical addressing. 
Could you please run it with those 4 PRs and the following environment variables set:
```
CLVK_SPIRV_ARCH=spir64
CLVK_PHYSICAL_ADDRESSING=1
```

rjodinchr:
It also has the same issue as #1220 . Maybe wait for a fix for it before testing with physical addressing.

rjodinchr:
With #1240, you can try running with physical addressing.

zwang20:
[...]

PR #1236, opened 2023-09-22
fix rework on constant expr from phi nodes

rjodinchr:
constant expr inside phi nodes need to be inserted in the correct basic block to avoid inserting non-phi instructions before phi ones.

Ref #1222


alan-baker (approved):
(note to self: can be early in the block since none of the operands will depend on anything in the block)

PR #1241, opened 2023-09-27
fix phi on private pointer

### LINKED ISSUE #1224: UNREACHABLE executed at lib/SPIRVProducerPass.cpp:5028
Compiling the attached source with the following command line
```
unzip a.zip
clspv -cl-mad-enable -DDEVICE_INFO=32 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=3 -DDEV_VER_MINOR=0 -D_OPENCL_COMPILER -DHASH_LOOPS=923 -DPLAINTEXT_LENGTH=55 -cl-single-precision-constant -cl-kernel-arg-info -rounding-mode-rte=16,32,64 -rewrite-packed-structs -std430-ubo-layout -decorate-nonuniform -arch=spir --use-native-builtins=acos,acosh,acospi,asin,asinh,asinpi,atan,atan2,atan2pi,atanh,atanpi,ceil,copysign,fabs,fdim,floor,fma,fmax,fmin,frexp,half_rsqrt,half_sqrt,isequal,isfinite,isgreater,isgreaterequal,isinf,isless,islessequal,islessgreater,isnan,isnormal,isnotequal,isordered,isunordered,ldexp,mad,rint,round,rsqrt,signbit,sqrt,tanh,trunc, -spv-version=1.6 -max-pushconstant-size=256 -max-ubo-size=65536 -global-offset -long-vector -module-constants-in-storage-buffer -cl-arm-non-uniform-work-group-size -enable-printf -printf-buffer-size=1048576 -x ir a.bc
```
results in the following output
```
OpPtrAccessChain is not supported for this storage class
UNREACHABLE executed at /path/to/clvk/external/clspv/lib/SPIRVProducerPass.cpp:5028!
```
[a.zip](https://github.com/google/clspv/files/12534644/a.zip)

rjodinchr:
- fix isimplicitcast for phi nodes:
  - compare infertype of phi with type of incoming values instead of incoming values against each other.
- insert needed gep between gep an phi aliasing
- consider scalar-alloca to avoid adding wrong gep

Fix #1224

### DIFF
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -898,28 +898,16 @@ bool IsImplicitCasts(Module &M, DenseMap<Value *, Type *> &type_cache,
     source_ty = clspv::InferType(source, M.getContext(), &type_cache);
     dest_ty = st->getValueOperand()->getType();
   } else if (auto *phi = dyn_cast<PHINode>(&I)) {
-    if (phi->getNumIncomingValues() > 0) {
-      auto RefOp = phi->getIncomingValue(0);
-      auto RefOpTy = clspv::InferType(RefOp, M.getContext(), &type_cache);
-      for (unsigned i = 1; i < phi->getNumIncomingValues(); i++) {
+    auto RefTy = clspv::InferType(phi, M.getContext(), &type_cache);
+    if (RefTy) {
+      for (unsigned i = 0; i < phi->getNumIncomingValues(); i++) {
         auto Op = phi->getIncomingValue(i);
         auto OpTy = clspv::InferType(Op, M.getContext(), &type_cache);
-        if (OpTy && RefOpTy &&
-            SizeInBits(M.getDataLayout(), OpTy) >
-                SizeInBits(M.getDataLayout(), RefOpTy)) {
+        if (OpTy != RefTy) {
           source = Op;
           source_ty = OpTy;
-          dest_ty = RefOpTy;
-        } else if (OpTy && RefOpTy &&
-                   SizeInBits(M.getDataLayout(), OpTy) <
-                       SizeInBits(M.getDataLayout(), RefOpTy)) {
-          source = RefOp;
-          source_ty = RefOpTy;
-          dest_ty = OpTy;
-        } else if (OpTy != RefOpTy) {
-          source = Op;
-          source_ty = OpTy;
-          dest_ty = RefOpTy;
+          dest_ty = RefTy;
+          break;
         }
       }
     }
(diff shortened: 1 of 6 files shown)


lib/BitcastUtils.cpp
```diff
@@ -898,31 +898,24 @@ bool IsImplicitCasts(Module &M, DenseMap<Value *, Type *> &type_cache,
     source_ty = clspv::InferType(source, M.getContext(), &type_cache);
     dest_ty = st->getValueOperand()->getType();
   } else if (auto *phi = dyn_cast<PHINode>(&I)) {
-    if (phi->getNumIncomingValues() > 0) {
-      auto RefOp = phi->getIncomingValue(0);
-      auto RefOpTy = clspv::InferType(RefOp, M.getContext(), &type_cache);
-      for (unsigned i = 1; i < phi->getNumIncomingValues(); i++) {
+    auto phiImplicitCast = [&M, &type_cache, &phi, &source, &source_ty,
```
alan-baker, line 901: Feels strange to define a lambda only for here. Am I missing a reason to just have this be functionality?

test/PointerCasts/phi-from-gep.ll
```diff
@@ -23,5 +23,50 @@ block:
...
+define spir_kernel void @test3(ptr addrspace(1) %a, i32 %i) {
+entry:
+; CHECK: entry
+; CHECK-NEXT: [[gep:%[^ ]+]] = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %i, i32 0
+  %0 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %i
+  br label %end
+
+block:
+; CHECK: block
+; CHECK-NEXT: [[gepblock:%[^ ]+]] = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 0, i32 %i
```
alan-baker, line 63: This gep seems wrong. SPIR-V requires  that the index into the vector be in bounds. What happens if `i` is greater than 4? I think it should be more like:
```
%div = sdiv i32 %i, 4
%mod = smod i32 %i, 4
%gepblock = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %div, i32 %mod
```

Is this adjusted later? If so, can you add a two pass test that shows it?

alan-baker (changes requested):

lib/BitcastUtils.cpp
rjodinchr: I used a lambda because it felt easier to manage to control flow with return.
Going back to an inlined implementation

test/PointerCasts/phi-from-gep.ll
rjodinchr: This test is showing an issue that has nothing to do with this current PR. I have created #1243 to track the issue.
I have also changed this test to avoid having this wrong pattern.

alan-baker (approved):

Issue #1243, opened 2023-09-29
SimplifyPointerBitcast performs wrong simplification


rjodinchr:
considering the following example:
```
; RUN: clspv-opt %s -o %t.ll --passes=simplify-pointer-bitcast --debug &> %t.debug.ll
; RUN: FileCheck %s < %t.ll

; CHECK: [[gep:%[^ ]+]] = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %i

; Wrong simplification! we should have something with the div and mod of %i in the gep indices
; CHECK: getelementptr <4 x i32>, ptr addrspace(1) %a, i32 0, i32 %i

define spir_kernel void @test(ptr addrspace(1) %a, i32 %i) {
entry:
  %0 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %i
  %1 = getelementptr i32, ptr addrspace(1) %a, i32 %i
  ret void
}
```
Here is the debug output:
```
##runOnImplicitGEP (aliasing):
adding:   %1 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 0, i32 0
instead of operand 0 of:   %2 = getelementptr i32, ptr addrspace(1) %a, i32 %i

##runOnGEPFromGEP:
replace:   %1 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 0, i32 0
  %2 = getelementptr i32, ptr addrspace(1) %1, i32 %i
by:   %2 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 0, i32 %i
```
I am not sure whether `runOnImplicitGEP` or `runOnGEPFromGEP` is the issue. But the combination ends up with something wrong.

Instead we should be expecting:
```
  %0 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %i
  %div = sdiv i32 %i, 4
  %mod = smod i32 %i, 4
  %1 = getelementptr <4 x i32>, ptr addrspace(1) %a, i32 %div, i32 %mod
```

rjodinchr:
fixed by #1282

PR #1244, opened 2023-09-29
add example of issue 1243 to detect changes in clspv

rjodinchr:
Ref #1243


rjodinchr:
I would worry that we did not figure out the right IR now, thus missing a potential fix for it.

alan-baker (approved):

PR #1249, opened 2023-10-05
Implement the OpFRem instruction.

AlexDemydenko:
The test sute test_spirv_new from OpenCL-CTS fails.
Algorithms of the fmod() function require synchronization
between compilation procedure from *.cl and *.spv

The clspv*.bc has an imlementation of the fmod() function.
The source https://github.com/llvm/llvm-project/blob/main/libc/src/__support/FPUtil/generic/FMod.h

The vulkan driver may have another version of OpFRem that is worse but good enough to pass all graphics CTS tests.
```
float frem(float x, float y) {
  return x - y * trunc( x/y );
}
```

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -29,6 +29,7 @@
 #include <unordered_set>
 #include <utility>
 
+#include "llvm/ADT/APFloat.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/ADT/UniqueVector.h"
 #include "llvm/Analysis/LoopInfo.h"
@@ -561,6 +562,11 @@ struct SPIRVProducerPassImpl {
     SPIRVOperandVec Ops;
     Ops.emplace_back(LITERAL_STRING, V);
     return addSPIRVInst<TSection>(Op, Ops);
+  }
+
+  void addSPIRVLabel(const SPIRVID &RID) {
+    SPIRVOperandVec Ops;
+    SPIRVSections[kFunctions].emplace_back(spv::OpLabel, RID, Ops);
   }
 
   //
@@ -2324,7 +2330,7 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVConstant(Constant *C) {
     }
     Fn->print(errs());
     llvm_unreachable("Unhandled function declaration/definition");
- } else if (auto *ConstExpr = dyn_cast<ConstantExpr>(Cst)) {
+  } else if (auto *ConstExpr = dyn_cast<ConstantExpr>(Cst)) {
     // If there is exactly one use we know where to insert the instruction
     if (ConstExpr->getNumUses() == 1) {
       auto *User = *ConstExpr->user_begin();
@@ -3449,7 +3455,6 @@ spv::Op SPIRVProducerPassImpl::GetSPIRVBinaryOpcode(Instruction &I) {
       {Instruction::SDiv, spv::OpSDiv},
       {Instruction::FDiv, spv::OpFDiv},
       {Instruction::URem, spv::OpUMod},
-      {Instruction::FRem, spv::OpFRem},
       {Instruction::Or, spv::OpBitwiseOr},
       {Instruction::Xor, spv::OpBitwiseXor},
       {Instruction::And, spv::OpBitwiseAnd},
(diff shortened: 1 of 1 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -3449,7 +3455,6 @@ spv::Op SPIRVProducerPassImpl::GetSPIRVBinaryOpcode(Instruction &I) {
       {Instruction::SDiv, spv::OpSDiv},
       {Instruction::FDiv, spv::OpFDiv},
       {Instruction::URem, spv::OpUMod},
-      {Instruction::FRem, spv::OpFRem},
```
rjodinchr, line 3452: Just removing the mapping to `OpFRem` feels wrong.
Maybe have a look at what we did for the division: https://github.com/google/clspv/pull/891

AlexDemydenko: Yes, I also don't like the idea to apply changes in spirv on the last pass instead of produce LLVM-IR on one of the early passes.
And I also started with ReplaceOpenCLBuiltinPass. The problem is that I'm dealing with an "Instruction", not a "Builtins" function.

The clvk drivers use the standard SPIRV-LLVM-Translator project. And this project converts spv::OpFMod and spv::OpFRem to Instruction::FRem (not as a function and not as a Builtins).

I need a pass that is very similar to ReplaceOpenCLBuiltinPass but deals with all the Instructions and can replace any of them not only "Functions".

I can create a new pass that can be placed after ReplaceOpenCLBuiltinPass. And add a flag something like "External spirv/IR" to enable this flag/pass only for IR generated by any external parser.

AlexDemydenko: Have any comments or ideas?

rjodinchr: In what range is the current operator failing?
Implementing the floating point reminder with integers feels like a big drop in performance. 

Would `x - y * trunc( x/y )`  work, using the compliant fp division?

AlexDemydenko: 1) The problems accur all over the range, but always the delta in last few bits. 
Example:
```
fmod(0.89177298546, 0.05671297759)
*.cl    result = 0.04107832164
*.spv result = 0.04107832909
delta = 0.00000000745
```

2) "Floating point reminder" already exists in the compilation flow, in clspv*.bc. The problem is that it only works for *.cl -> compute.spv and doesn't work for kernel.spv -> comute.spv.

3) The algorithm that I use in the commit is based on what I get from cl -> * cpv flow.
4) Where is the problem. The OCL-CTS test has a list of OCL-functions that have equivalents in SpirV, and the test expects that simple kernels with these functions, one by one, from the spirv and cl sources must produce identical results.

rjodinchr: Do you have an application that needs that feature?
If this is only for compliance, clvk can be compiled without this support using `-DCLVK_ENABLE_SPIRV_IL=OFF` which kind of "solves" the issue.

AlexDemydenko: While I checking the requirements for our project.

Could you please share with me you vision on a couple questions.
Are you planing to handle the cl_khr_il_program extension + SPIRV in future?
Is the InputLanguage option public with full support  for any valid IL or is it just an option for debugging purposes that supports IL from a specific version of the LLVM FrontEnd?
`
static llvm::cl::opt<clang::Language> InputLanguage...
`

AlexDemydenko: While I checking the requirements for our project.

Could you please share with me you vision on a couple questions.
Are you planing to handle the cl_khr_il_program extension + SPIRV in future?
Is the InputLanguage option public with full support  for any valid IL or is it just an option for debugging purposes that supports IL from a specific version of the LLVM FrontEnd?
`
static llvm::cl::opt<clang::Language> InputLanguage...
`

rjodinchr: > Are you planing to handle the cl_khr_il_program extension + SPIRV in future?

In the near future, I would answer no for my side (chromeos).

> Is the InputLanguage option public with full support for any valid IL or is it just an option for debugging purposes that supports IL from a specific version of the LLVM FrontEnd?

The InputLanguage main goal is to be able to link OpenCL-C program together then it can be used for other things.

AlexDemydenko: I had discussion with other teams. I can assume that there are no requirements for this future now. But everyone mentions that it is  a "Nice to have", and we need to try ta save it.
There some different options such as cpv-fork, cvk-fork, etc.
But the most preferable option, in our opinion, it is to inject such changes directly to you repository :).

My proposal.
I can create a new pass, something like LLVMTranslatorPatch, that should cover  all the side effect that occur if the IL comes from SPIRV-LLVM-Translator.
To enable this pass, I will add llvm::cl::opt<bool> supportLLVMTranslator = false;
And I can add braces #ifdef SUPPORT_LLVM_TRANSLATOR with cmake-option if needed.

If it works for you, I'd be happy to prepare these changes for review.

AlexDemydenko: To move forward I need your response.
If it works for you I will be happy to prepare a new commit, if not then I will remove the pull request.

rjodinchr: @alan-baker what do you think?

alan-baker: I'm not a big fan of build time controls. I wouldn't want to have another set of bots for this, but then we wouldn't be testing it regularly. More palatable to me would be a runtime option which enables/disables a pass.  Could this be done as conditional transform in ReplaceLLVMIntrinsics?

AlexDemydenko:
The problem would be discussed in the https://github.com/google/clspv/issues/1325

PR #1252, opened 2023-10-18
Add fabs version not flushing to zero denormalized numbers

rjodinchr:
Both the OpenCL and the SPIR-V specification allow denormalized numbers to be flushed to zero.

But since llvm/llvm-project@5c0da5839de1bdc08f411a04305a9bdadf538ad5 patterns from libclc like in `cbrt` are converted to call to fabs:
```
float input;
int input_as_uint = as_uint(input);
int input_without_sign = input_as_uint & 0x7fffffff;
```
This is causing issue as the cbrt algorithm in libclc needs denormalized numbers not to be flushed to zero to be OpenCL compliant.

Unless clspv is called with fabs in the native builtins list, use a software version of fabs that guarantees denormalized numbers will not be flushed to zero.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -473,6 +473,7 @@ struct SPIRVProducerPassImpl {
   SPIRVID GenerateShuffle2FromCall(Type *Ty, Value *SrcA, Value *SrcB,
                                    Value *Mask);
   SPIRVID GeneratePopcount(Type *Ty, Value *BaseValue, LLVMContext &Context);
+  SPIRVID GenerateFabs(Value *Input);
   void GenerateInstruction(Instruction &I);
   void GenerateFuncEpilogue();
   void HandleDeferredInstruction();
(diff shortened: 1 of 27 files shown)


alan-baker (approved):

alan-baker:
> In fact should we instead prevent this optimisation from happening in llvm when compiling for OpenCL? as the transformation is not correct for OpenCL since denormalized numbers may be flushed to zero which is not the case in C (I think)?

I'm happy to merge this and open an issue for llvm. I expect any patch there would take longer. So we could have an issue to revert this if LLVM updates it.

rjodinchr:
> > In fact should we instead prevent this optimisation from happening in llvm when compiling for OpenCL? as the transformation is not correct for OpenCL since denormalized numbers may be flushed to zero which is not the case in C (I think)?
> 
> I'm happy to merge this and open an issue for llvm. I expect any patch there would take longer. So we could have an issue to revert this if LLVM updates it.

Sounds good, I'll open an issue in llvm and clspv then.

Issue #1253, opened 2023-10-18
math functions issue since instcombine transform "and" pattern to call to fabs


rjodinchr:
Some Math function test of the OpenCL-CTS regressed since clspv contains https://github.com/llvm/llvm-project/commit/5c0da5839de1bdc08f411a04305a9bdadf538ad5

clspv workaround: https://github.com/google/clspv/pull/1252

A cleaner fix might be to prevent the transformation in llvm: https://github.com/llvm/llvm-project/issues/69464

PR #1256, opened 2023-11-07
decorate libclc functions with noimplicitFloat

rjodinchr:
Ref #1253


rjodinchr:
I have tried to decorate within libclc, but clang is ignoring `__attribute__((noimplicitfloat))`

alan-baker (approved):

PR #1267, opened 2023-12-05
Add GN build infra

gowtham-sarc:
This change adds in the support for building using the GN build system. There were no changes to the existing CMake files.


alan-baker:
I believe this will reside in ANGLE, so closing this PR.

PR #1274, opened 2023-12-18
Fix Android build

jrprice:
SPIRV-Tools no longer builds any executables when targeting Android, so disable attempts to build and use `spirv-dis` in Clspv.


dneto0 (approved):
This seems ok.
I think it would be slightly better to set the *default* based on whether it's Android or not.
As this chagne goes, the existence of the option depends on whether the target is Android or not.

PR #1275, opened 2023-12-18
Add --producer-out-file to tests as needed

dneto0:
Otherwise, this fails tests in some environments that check the existence of the output file.


dneto0:
this is required for google3 import

jrprice (approved):

rjodinchr:
@Rekt3421 FYI

PR #1285, opened 2024-01-02
cmake: Use find_package(Python3...) instead of FindPythonInterp

dneto0:
The FindPythonInterp module was deprecated in CMake 3.12. Comply with CMP0148 deprecation policy in CMake 3.27.

Use FindPython3 module instead, added in CMake 3.12


alan-baker (approved):

dneto0:
still building locally, and it seems this isn't enough.

dneto0:
ok, now this passes in my local build.  Will wait for bots.

PR #1289, opened 2024-01-26
Fixed " undeclared identifier" for the atomic_store* functions with c…

AlexDemydenko:
…l-std=3.0 argument.

The OCL-CTS has failed the atomic_fence test in the test_c11_atomics testsuite .

The diference between std2.0 and std3.0  in the file opencl-c-base.h
```
// Define feature macros for OpenCL C 2.0
#if (__OPENCL_CPP_VERSION__ == 100 || __OPENCL_C_VERSION__ == 200)
#define __opencl_c_pipes 1
#define __opencl_c_generic_address_space 1
#define __opencl_c_work_group_collective_functions 1
#define __opencl_c_atomic_order_acq_rel 1
#define __opencl_c_atomic_order_seq_cst 1
#define __opencl_c_atomic_scope_device 1
#define __opencl_c_atomic_scope_all_devices 1
#define __opencl_c_device_enqueue 1
#define __opencl_c_read_write_images 1
#define __opencl_c_program_scope_global_variables 1
#define __opencl_c_images 1
#endif

// Define header-only feature macros for OpenCL C 3.0.
#if (__OPENCL_CPP_VERSION__ == 202100 || __OPENCL_C_VERSION__ == 300)
// For the SPIR and SPIR-V target all features are supported.
#if defined(__SPIR__) || defined(__SPIRV__)
#define __opencl_c_work_group_collective_functions 1
#define __opencl_c_atomic_order_seq_cst 1
#define __opencl_c_atomic_scope_device 1
#define __opencl_c_atomic_scope_all_devices 1
#define __opencl_c_read_write_images 1
#endif // defined(__SPIR__)
```

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -229,6 +229,12 @@ clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
     if (clspv::Option::ImageSupport()) {
       EnabledFeatureMacros.insert(clspv::FeatureMacro::__opencl_c_images);
     }
+    // The futures should be enabled for __OPENCL_C_VERSION__ == 300
+    // while the __SPIR__ and __SPIRV__ defines are undefined.
+    EnabledFeatureMacros.insert(
+        clspv::FeatureMacro::__opencl_c_atomic_scope_device);
+    EnabledFeatureMacros.insert(
+        clspv::FeatureMacro::__opencl_c_atomic_order_seq_cst);
 
     // TODO remove when feature macros are added to OpenCLExtensions.def
     // https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Basic/OpenCLExtensions.def#L119
--- a/test/Features/cl3-no-features.cl
+++ b/test/Features/cl3-no-features.cl
@@ -20,14 +20,6 @@
 #error __opencl_c_subgroups should not be defined
 #endif
 
-#ifdef __opencl_c_atomic_order_seq_cst
-#error __opencl_c_atomic_order_seq_cst should not be defined
-#endif
-
-#ifdef __opencl_c_atomic_scope_device
-#error __opencl_c_atomic_scope_device should not be defined
-#endif
-
 #ifdef __opencl_c_atomic_scope_all_devices
 #error __opencl_c_atomic_scope_all_devices should not be defined
 #endif
--- a/test/Features/cl3-some-features.cl
+++ b/test/Features/cl3-some-features.cl
@@ -24,8 +24,8 @@
 #error __opencl_c_atomic_order_seq_cst should be defined
 #endif
 
-#ifdef __opencl_c_atomic_scope_device
-#error __opencl_c_atomic_scope_device should not be defined
+#ifndef __opencl_c_atomic_scope_device
+#error __opencl_c_atomic_scope_device should be defined
 #endif
 
 #ifdef __opencl_c_atomic_scope_all_devices


AlexDemydenko:
Is there any issues with this change that blocks it to being submitted?

rjodinchr:
As mentioned by Alan, this is breaking some tests so we cannot merge it as it is.

But on top of that, I agree with Alan, this is something that just needs to be handled by whatever is calling `clspv`.
Here is an example with `clvk`: https://github.com/kpet/clvk/blob/d2df460ccf90491c822e0ebac39ceeb852179a6c/src/device.cpp#L695

I don't think we should merge that PR is `clspv`.

AlexDemydenko:
About the tests. These tests check that these default defines must be disabled or not. The problem is not that my PR makes compilation fails or making wrong compilation as side effect. My PR changes the default defines status and it has an inconsistence with the previous status. I will add changes to the tests for this PR.

AlexDemydenko:
About why I am do this in the clspv.
The root problem is that the clspv forces #undef for the __SPIR__ & __SPIRV__.
And it provides different behavior between std30 and std20.
The problem in the file opencl-c-base.h
`// Define feature macros for OpenCL C 2.0
#if (__OPENCL_CPP_VERSION__ == 100 || __OPENCL_C_VERSION__ == 200)
#define __opencl_c_pipes 1
#define __opencl_c_generic_address_space 1
#define __opencl_c_work_group_collective_functions 1
#define __opencl_c_atomic_order_acq_rel 1
#define __opencl_c_atomic_order_seq_cst 1
#define __opencl_c_atomic_scope_device 1
#define __opencl_c_atomic_scope_all_devices 1
#define __opencl_c_device_enqueue 1
#define __opencl_c_read_write_images 1
#define __opencl_c_program_scope_global_variables 1
#define __opencl_c_images 1
#endif

// Define header-only feature macros for OpenCL C 3.0.
#if (__OPENCL_CPP_VERSION__ == 202100 || __OPENCL_C_VERSION__ == 300)
// For the SPIR and SPIR-V target all features are supported.
#if defined(__SPIR__) || defined(__SPIRV__)
#define __opencl_c_work_group_collective_functions 1
#define __opencl_c_atomic_order_seq_cst 1
#define __opencl_c_atomic_scope_device 1
#define __opencl_c_atomic_scope_all_devices 1
#define __opencl_c_read_write_images 1
#endif // defined(__SPIR__)
`

AlexDemydenko:
The final question.
Why the defines __opencl_c_atomic_* is not allowed in the std30 if the are works fine with the std20 and exists there by default?

rjodinchr:
> The final question. Why the defines __opencl_c_atomic_* is not allowed in the std30 if the are works fine with the std20 and exists there by default?

I think std20 requires it, while std30 has it optional depending on whether the device supports it or not?

rjodinchr:
Here is what the specification states:
```
Optional core language features are described in this document. They are optional from OpenCL C 3.0 onwards and therefore are not supported by all implementations. When an OpenCL C 3.0 optional feature is supported, an associated feature test macro will be predefined.

The following table describes OpenCL C 3.0 or newer features and their meaning. The naming convention for the feature macros is __opencl_c_<feature_name>.

Feature macro identifiers are used as names of features in this document.
```
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#features

I think this PR should not be merged in `clspv`.

AlexDemydenko:
Ok. For the OCL-C these are optional features. It is clear.
But your target is spirv and spirv says that it is not optional it is a basic feature. All devices that supports spirv (vulkan) should  support these features and do not require any "Capabilities" for it.
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Memory_Semantics_-id-

What is the point of the reducing functionality if you target has to support it, if it says that supports the spirv?

rjodinchr:
All devices do not support `vulkanMemoryModelDeviceScope`, thus I don't think we can always add `__opencl_c_atomic_scope_device`.
> vulkanMemoryModelDeviceScope indicates whether the Vulkan Memory Model can use Device scope synchronization. This also indicates whether shader modules can declare the VulkanMemoryModelDeviceScope capability.

And for the memory model:
> vulkanMemoryModel indicates whether shader modules can declare the VulkanMemoryModel capability.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceVulkanMemoryModelFeatures.html

AlexDemydenko:
Is the target of clspv vulkan or spirv spec? I am showing you proofs based on the SPIRV specification, and you are answering to me with the examples from vulkan spec. I need to clarify this, to be on the same page with you.

Yes, in the vulkan 1.3, an extension appears that has options to support the device memory model & scope. But the problem is that vulkan has many different options for very specific needs. It has options to support graphics or compute pipelines (the GLCompute runtime)., This is doesn't means that you have to think about how to support a device that doesn't have a compute pipeline. The SPIRV language it is also an optional language (can compile directly with glsl, if we are talking about extensions that just exists in the vulkan specification).

Can I have a device in the vulkan devices list that doesn't support the memory model and memory scope? Yes. Does this device support the GLCompute execution module and SPIRV language? I don't think so. Because spirv has its own requirements in own spec and it require the memory scope.

rjodinchr:
Let's think of it another way. If we always add `__opencl_c_atomic_scope_device` it would mean that all devices have scope device atomic. This is clearly not the case. Whether you're talking OpenCL, SPIR-V or Vulkan, it doesn't matter.

It's not because SPIR-V exposes something without capability from version 1 that it is supported by every device and every language using SPIR-V.

As clspv generate what I would call Vulkan SPIR-V (there are plenty other SPIR-V language, like [OpenCL SPIR-V](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html)), it needs to generate something that works with Vulkan. In this case we should only define things like `__opencl_c_atomic_scope_device` if the device really supports it. And to know that it's vulkan which has the information, not the SPIR-V specification.

AlexDemydenko:
I want to share my picture of this from my side.
The solution consist of two parts: driver (for example clvk) and compiler(clspv).
Compiler must conform to  the Opencl-C/C++ and SPIRV specifications .
Driver must conform to  the vulkan specification.

Vulkan has an extension that changes the default behavior of SPIRV. It disables all scope modes except Relaxed.
Which module have to solve this issue the driver or the compiler? In my opinion it should be the driver.

The clspv follows the base SPIRV behavior and in this case everything is clear to me. I know what to expect from this module based on SPIRV specification. The compiler needs provide the clspv::Option::XXX  to change default behavior.

On other hand, I have a compiler that matches the behavior when the driver supports the CapabilityVulkanMemoryModel & CapabilityVulkanMemoryModelDeviceScope capability (which is optional for the driver) and vulkanMemoryModelDeviceScope = VK_FALSE if the application sets this struct what is optional too. In all other cases, I need to use --enable-feature-macros just to return compiler to the default SPIRV behavior, even if driver doesn't know about an existing CapabilityVulkanMemoryModel. 

In my opinion, the driver should somehow set "#undef __opencl_c_atomic_scope_device" only in case if it parses the VkPhysicalDeviceVulkanMemoryModelFeatures structure, otherwise it should just follow the default behavior of the compiler which follows the SPIRV specification.

rjodinchr:
I understand that we could have it the other way as you explained. But there are no added values of changing it right now. And to be honest I don't agree, I think the way it is now is the right one.

But in both cases (what we have now, and what you think is better), the driver needs to do something, either:
- undefine the macro if not supported
- define the macro if supported

This is basically the same logic in the driver code.

So I think, you should just implement it the way it is in clspv at the moment in your driver, and you should not need this PR anymore.

lpavank:
@here, excuse me for joining the conversation in the middle.
But, trying to connect the dots.
IIRC, this test in particular had a bug/issue which I believe was identified/fix back in late 2022/early 2023, and it was observed during Samsung collaboration effort with Codeplay (Intel).
Here is/was the fix for that.
https://github.com/KhronosGroup/OpenCL-CTS/pull/1603

Firstly, @AlexDemydenko can you please confirm that CTS you tried (when this test failed on our side) has or does not have the fix above please?
If it does not have the fix, I would recommend to pull that CTS fix in, and see what happens please.
PS: I understand @AlexDemydenko might not be aware of it, so sharing here, so that we are all on same page.

AlexDemydenko:
The latest version of OpenCL-CTS contains fixes for the current benchmark. Now such futures are not required for a successful compilation.

PR #1291, opened 2024-01-29
Explicitly annotate intentional fallthroughs

dneto0:
Use the C++17 attribute.
This avoids warnings, triggered in Clang 16 at least.


alan-baker (approved):
Assuming gcc and msvc also like this change.

Issue #1292, opened 2024-01-29
Canonicalization of GEPs to i8


alan-baker:
LLVM now canonicalizes some (constant) GEPs to perform address calculations in terms of bytes. This is part of larger change of moving towards `ptradd` instead of `getelementptr`. This will require changes in clspv to keep pace.

At the moment I'm seeing different failures:
* unit tests with incorrect codegen expectations (~115 per arch)
* cts failures that look functionally incorrect

For example, in `test/CommonBuiltins/min/half2_fmin.cl` the loads are broken down into 4 1-byte loads, a vector creation, and a bitcast from `v4uchar` to `v2half`. This is problematic for a few reasons:
1. Not all devices support byte addressing and clspv now requires it.
2. It relies on downstream drivers to properly coalesce memory accesses. This might be ok, but I don't have performance numbers that are easily accessible one way or the other.
3. It muddies the IR. This isn't *just* an aesthetic concern. There is a binary size increase in these cases due to more instructions getting used.

CTS failures:
* basic/vstore_local (fails char2 and short2)
* relationals/shuffle_copy
* relationals/shuffle_function_call
* relationals/shuffle_built_in
* relationals/shuffle_built_in_dual_input
* relationals/shuffle_array_cast

The likely solution is a change in type inference that can find a better type than i8 by looking at the GEP's users (right now it stops at any GEP).

rjodinchr:
This change in LLVM is negatively impacting performance on ChromeOS. Between 10% to 30% regression depending on the various benchmark we have.

alan-baker:
Clspv will generate invalid SPIR-V some cases due to this issue. Take `test/Coherent/frexp.cl` for example. When compiled with `-cl-native-math` the resulting SPIR-V generates the following validation error:
```
error: line 78: GLSL.std.450 Frexp: expected operand Exp data type to be a 32-bit int scalar or vector type
  %50 = OpExtInst %float %49 Frexp %45 %48
```

The `data + 1` results in gep canonicalization happening and clspv pick `i8` as the underlying type for `data`.

I plan to update the coherent tests so here is the source for reference:
```
kernel void foo(global int *data, global float* x) {
  float y = data[0];
  barrier(CLK_GLOBAL_MEM_FENCE);
  *x = frexp(y, data + 1);
}
```

dneto0:
[...]

PR #1294, opened 2024-01-30
fix GetIdxsForTyFromOffset with local type

rjodinchr:
Both resource and local type start with array of zero element. Also fix the startIdx for those as the "empty" array will be removed by the rework of the type getting rid of it.

### DIFF
--- a/lib/BitcastUtils.cpp
+++ b/lib/BitcastUtils.cpp
@@ -1264,16 +1264,19 @@ GetIdxsForTyFromOffset(const DataLayout &DataLayout, IRBuilder<> &Builder,
   assert(Src->getType()->isPointerTy());
   bool clspv_resource = false;
   if (auto call = dyn_cast<CallInst>(Src)) {
-    clspv_resource =
-        clspv::Builtins::Lookup(call->getCalledFunction()).getType() ==
-        clspv::Builtins::kClspvResource;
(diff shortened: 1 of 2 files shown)


lib/BitcastUtils.cpp
```diff
@@ -1264,16 +1264,19 @@ GetIdxsForTyFromOffset(const DataLayout &DataLayout, IRBuilder<> &Builder,
...
   }
 
   unsigned startIdx = 0;
   if ((isa<GlobalVariable>(Src) || clspv_resource || isa<AllocaInst>(Src)) &&
       SrcTy != DstTy) {
     Idxs.push_back(ConstantInt::get(Builder.getInt32Ty(), 0));
-    startIdx = 1;
+    if (!(SrcTy->isArrayTy() && GetNumEle(SrcTy) == 0)) {
+      startIdx = 1;
+    }
```
alan-baker, line 1279: This probably deserves a comment and perhaps helper method `IsUnsizedType()`. Alternatively it might be clearer to break this in a global variable and alloca case and then a resource (global or local) case and mention that interfaces use unsized arrays again.

alan-baker (approved):
Nit about explanation.

alan-baker (approved):

rjodinchr:
Ref #1292

PR #1295, opened 2024-01-30
allow unsupported qcom-accelerate-16-bit qualcomm specific option

rjodinchr:


kpet:
Or change applications that are meant to be portable but pass vendor options so they check they are targeting an implementation that can understand them. I'm not sure we want to start swallowing any and all options we don't know. What about the case of an application that suffers from a typo in an option name?

rjodinchr:
You made a good point @kpet, I agree.

Another possibility on the clspv side would be to skip only this qualcomm specific option?

What do you think @alan-baker?

alan-baker:
Skipping specific options with warnings sounds better than skipping any unknown option. I know I'd get frustrated if a typo was silently ignored. If it is easy to change the app that is preferable, but I could live with a skip list.

kpet:
Fine with a skip list too. Note that you could easily implement that in clvk (assuming that's only needed there). There's a mechanism for option substitution that we already use to swallow two standard options. We used to ignore at least one Arm-specific option but that's now gone.

rjodinchr:
I would also prefer to have it managed at the application level.
It feels a bit more tricky to manage in clvk, as we would need to add support for the assignment part (`qcom-accelerate-16-bit={true|false|0|1|2|...}`.

rjodinchr:
https://github.com/tensorflow/tensorflow/commit/7cf223dfe71b44b5ef49b7b51ead28772834b279

PR #1297, opened 2024-02-01
update LLVM & rework runOnUpgradeableConstantCasts

rjodinchr:
- fix GetIdxsForTyFromOffset when called on unsized types
- rework runOnUpgradeableConstantCasts to be more versatile/general
- add a case in runOnImplicitGEP for implicit cast between clspvResourceOrLocal and GEP
- fix runOnAllocaNotAliasing to avoid lowering alloca that do not need to be lowered
- Update tests
- Some tests need more work and are marked FAIL

Ref #1292


test/HalfStorage/clspv_vloada_half2_private.cl
```diff
@@ -1,4 +1,4 @@
-// RUN: clspv %target %s -o %t.spv
+// RUN: clspv %target %s -o %t.spv --print-after-all &> %t.ll
```
alan-baker, line 1: Debug code left around?

```suggestion
// RUN: clspv %target %s -o %t.spv
```

test/HalfStorage/vloada_half2_global.cl
```diff
@@ -10,7 +10,7 @@
 
 kernel void foo(global float2* A, global uint* B, uint n) {
   A[0] = vloada_half2(n, (global half*)B);
-  A[1] = vloada_half2(0, (global half*)B+2);
+  A[1] = vloada_half2(0, (global half*)(B+1));
```
alan-baker, line 13: This seems strange. Why is the checking code now correct for a different address?

test/PointerCasts/multiple_implcit_casts.ll
```diff
@@ -2,12 +2,14 @@
 ; RUN: FileCheck %s < %t.ll
 
 ; CHECK: @test1
-; CHECK: [[gep:%[a-zA-Z0-9_.]+]] = getelementptr %struct.Line, ptr addrspace(1) %lines, i32 %n, i32 1
-; CHECK: store float 0.000000e+00, ptr addrspace(1) [[gep]]
+; CHECK: [[gep1:%[a-zA-Z0-9_.]+]] = getelementptr %struct.Line, ptr addrspace(1) %lines, i32 %n, i32 1
```
alan-baker, line 5: This is a little awkward, maybe it would be better to add dce to the passes list.

test/VariablePointers/select_wg.cl
```diff
@@ -1,7 +1,9 @@
-// RUN: clspv %target %s -o %t.spv
+// RUN: clspv %target %s -o %t.spv --print-after-all --debug-only=SimplifyPointerBitcast &> %t.ll
```
alan-baker, line 1: Debug code left around?

alan-baker (approved):
Nits

test/HalfStorage/vloada_half2_global.cl
rjodinchr: The cast is applied before the add, so the address is the same, the difference is that before we used to have a pointer addition with a half type. This leads to a complete change in the generated code with half types all around.
With this new form, we are doing the pointer addition with a uint type.

alan-baker: Ah I missed the parentheses.

PR #1298, opened 2024-02-01
Fix expectations for PointerAccessChains/pointer_index_is_constant_1.cl

dneto0:
The new code is correct.  The constant indexing into an array of arrays of floats collapses into a constant index of an array of floats.

Ref #1292


dneto0:
cc: @rjodinchr

alan-baker (approved):
Looks ok. Arguably it should be better as a flat array assuming the underlying data type wasn't narrowed.

rjodinchr:
Thank you @dneto0.
I think I got to this one during the end, and I was a bit confused as to whether losing the arraystride was an issue or not.
In order to unblock TAP, I decided to just mark it as failing to come back to it later.
Thanks again for taking care of it.

PR #1302, opened 2024-02-07
upgrade GEPs of PHI when possible

rjodinchr:
As the last update of llvm introduced the canonicalization of GEPs to i8, we have a case where the user of a phi nodes is now a i8. This lead the inferering of the phi type to be lowered to i8 while it could have a higher type.

To avoid that, track possible upgrade of GEPs of PHI and upgrade them if all the user of the PHI can use the new higher type.

This is fixing the performance regression for the ChromeOS workloads.

Ref #1292


test/PointerCasts/1292-phi_users.cl
```diff
@@ -0,0 +1,315 @@
+// RUN: clspv %s -o %t.spv -cl-std=CL2.0 -inline-entry-points -cl-single-precision-constant -cl-kernel-arg-info -rounding-mode-rte=16,32 -fp64=0 -rewrite-packed-structs -std430-ubo-layout -decorate-nonuniform -hack-mul-extended -hack-convert-to-float -hack-image1d-buffer-bgra -arch=spir --use-native-builtins=ceil,copysign,exp2,floor,fma,fmax,fmin,half_exp,half_exp10,half_exp2,half_log,half_log10,half_log2,half_powr,half_rsqrt,half_sqrt,isequal,isfinite,isgreater,isgreaterequal,isinf,isless,islessequal,islessgreater,isnan,isnormal,isnotequal,isordered,isunordered,mad,rint,round,rsqrt,signbit,sqrt,trunc, -spv-version=1.6 -max-pushconstant-size=128 -max-ubo-size=1073741824 -global-offset -long-vector -module-constants-in-storage-buffer -cl-arm-non-uniform-work-group-size -enable-printf -printf-buffer-size=1048576
+// RUN: spirv-dis %t.spv -o %t.spvasm
+// RUN: spirv-val %t.spv --target-env spv1.6
```
alan-baker, line 3: What is actually being tested here? When I disable the change in SimplifyPointerBitcasts this still passes. This test is also enormous so it would be hard to figure out the problem in the future.

alan-baker (commented):
I cut down the IR based test, PTAL.

test/PointerCasts/1292-phi_users.cl
rjodinchr: This was failing when I had some bug in the implementation. But it was hard to extract a subset to just run one past, so I have added the whole kernel. We can remove it if you prefer.

rjodinchr:
> I cut down the IR based test, PTAL.

All good for me

test/PointerCasts/1292-phi_users.cl
alan-baker: I think I prefer that.

alan-baker (approved):

PR #1305, opened 2024-02-07
Add comments to lib/Types.cpp

dneto0:


lib/Types.cpp
```diff
@@ -214,42 +217,70 @@ Type *ExtractSmallerStructField(const DataLayout &DL, Type *Ty) {
...
+//   level of nesting "matches" the other, then return the aggregate or vector.
+//   They both have the same machine byte address, but we prefer the larger
+//   type.  Here two types "match" if they're the same, or if they are
+//   scalar numeric types (integer or float0 with the same bit size.
+// - If one type is a struct and is larger than the other type, but its
+//   smallest member is smaller than the other type, then return
+//   that member.
+// - Otherwise, if one type is smaller than the other, return the smaller one.
+// - When the sizes are the same, prefer the packed structure, if it exists.
 Type *SmallerTypeNotAliasing(const DataLayout &DL, Type *TyA, Type *TyB) {
```
dneto0, line 233: This method is possibly  misnamed.  Perhaps it should be PreferredTypeNotAliasing or something like that.

It's not always the smaller type.

lib/Types.cpp
```diff
@@ -214,42 +217,70 @@ Type *ExtractSmallerStructField(const DataLayout &DL, Type *Ty) {
...
 }
 
+// Returns a preferred type chosen from two types.
+// In particular:
+// - If either is nullptr, return the other.
+// - When one is an aggregate or vector whose first nested member at some
+//   level of nesting "matches" the other, then return the aggregate or vector.
+//   They both have the same machine byte address, but we prefer the larger
+//   type.  Here two types "match" if they're the same, or if they are
+//   scalar numeric types (integer or float0 with the same bit size.
```
rjodinchr, line 227:
```suggestion
//   scalar numeric types (integer or float with the same bit size.
```

rjodinchr (approved):

lib/Types.cpp
alan-baker: I think that was meant to be a closing parenthesis.

```suggestion
//   scalar numeric types (integer or float) with the same bit size.
```

lib/Types.cpp
```diff
@@ -214,42 +217,70 @@ Type *ExtractSmallerStructField(const DataLayout &DL, Type *Ty) {
...
+      return TyA;
+    }
+    if (BitcastUtils::FindAliasingContainedType(TyB, TyA, Steps, PerfectMatch,
+                                                DL, true)) {
+      // Prefer the composite type.
+      return TyB;
+    }
   }
 
+  // Handle more cases where TyA !=TyB
```
alan-baker, line 257:
```suggestion
  // Handle more cases where TyA != TyB
```

lib/Types.cpp
```diff
@@ -269,6 +300,8 @@ Type *SmallerTypeNotAliasing(const DataLayout &DL, Type *TyA, Type *TyB) {
   return TyA;
 }
 
+// Returns the type inferred by inspecting the users of |v|.
```
alan-baker, line 303:
```suggestion
// Returns the type of |v| inferred by inspecting its users.
```

alan-baker (approved):
Nits

lib/Types.cpp
```diff
@@ -214,42 +217,70 @@ Type *ExtractSmallerStructField(const DataLayout &DL, Type *Ty) {
...
 }
 
+// Returns a preferred type chosen from two types.
+// In particular:
+// - If either is nullptr, return the other.
+// - When one is an aggregate or vector whose first nested member at some
+//   level of nesting "matches" the other, then return the aggregate or vector.
+//   They both have the same machine byte address, but we prefer the larger
+//   type.  Here two types "match" if they're the same, or if they are
+//   scalar numeric types (integer or float with the same bit size.
```
alan-baker, line 227: In applying suggestions, I think the closing parenthesis was lost.

PR #1309, opened 2024-02-24
Copy metadata for new call functions when undoing pass by val

### LINKED ISSUE #1228: Segfaults when passing struct in global memory by value to a function.
The following kernel segfaults (w/ `-O0`):

```C
typedef struct Baz {
    float x;
} Baz;

static void foo(Baz baz) {
}

__kernel void test(__global Baz *baz) {
    foo(baz[0]);
}
```
[Godbolt link.](https://godbolt.org/z/8sndsEMG9)
On godbolt the following is being logged:

```
!dbg attachment points at wrong subprogram for function
!11 = distinct !DISubprogram(name: "test", scope: !12, file: !12, line: 9, type: !13, scopeLine: 9, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !21)
ptr @test
  call void @llvm.dbg.value(metadata %struct.Baz poison, metadata !31, metadata !DIExpression(DW_OP_constu, 0, DW_OP_swap, DW_OP_xderef)), !dbg !35
!35 = !DILocation(line: 0, scope: !32)
!32 = distinct !DISubprogram(name: "foo", scope: !12, file: !12, line: 6, type: !33, scopeLine: 6, flags: DIFlagPrototyped, spFlags: DISPFlagLocalToUnit | DISPFlagDefinition, unit: !0, retainedNodes: !21)
!32 = distinct !DISubprogram(name: "foo", scope: !12, file: !12, line: 6, type: !33, scopeLine: 6, flags: DIFlagPrototyped, spFlags: DISPFlagLocalToUnit | DISPFlagDefinition, unit: !0, retainedNodes: !21)
conflicting debug info for argument
  call void @llvm.dbg.value(metadata %struct.Baz poison, metadata !31, metadata !DIExpression(DW_OP_constu, 0, DW_OP_swap, DW_OP_xderef)), !dbg !35
!29 = !DILocalVariable(name: "baz", arg: 1, scope: !11, file: !12, line: 9, type: !15)
!31 = !DILocalVariable(name: "baz", arg: 1, scope: !32, file: !12, line: 6, type: !16)
LLVM ERROR: Broken module found, compilation aborted!
Program terminated with signal: SIGSEGV
Compiler returned: 139
```

Compilation no longer fails if I modify the function so that I keep the struct in global memory:

```C

typedef struct Baz {
    float x;
} Baz;

static void foo(__global Baz *baz) {
}

__kernel void test(__global Baz *baz) {
    foo(&baz[0]);
}
```

Rekt3421:
This change fixes #1228 by correctly copying over the metadata to any new calls generated as a result of undoing pass by val.

### DIFF
--- a/lib/UndoByvalPass.cpp
+++ b/lib/UndoByvalPass.cpp
@@ -144,6 +144,8 @@ PreservedAnalyses clspv::UndoByvalPass::run(Module &M,
 
         CallInst *NewCall = CallInst::Create(NewFunc, Args, "", Call);
         NewCall->setCallingConv(NewFunc->getCallingConv());
+        auto Instr = dyn_cast<Instruction>(User);
+        NewCall->copyMetadata(*Instr);
 
         Call->replaceAllUsesWith(NewCall);
         Call->eraseFromParent();
--- a/lib/UndoByvalPass.cpp
+++ b/lib/UndoByvalPass.cpp
@@ -144,8 +144,7 @@ PreservedAnalyses clspv::UndoByvalPass::run(Module &M,
 
         CallInst *NewCall = CallInst::Create(NewFunc, Args, "", Call);
         NewCall->setCallingConv(NewFunc->getCallingConv());
-        auto Instr = dyn_cast<Instruction>(User);
-        NewCall->copyMetadata(*Instr);
+        NewCall->copyMetadata(*Call);
 
         Call->replaceAllUsesWith(NewCall);
         Call->eraseFromParent();
--- a/test/issue-1228.cl
+++ b/test/issue-1228.cl
@@ -4,8 +4,7 @@ typedef struct Baz {
     float x;
 } Baz;
 
-static int foo(Baz baz) {
-	return 1;
+static void foo(Baz baz) {
 }
 
 __kernel void test(__global Baz *baz) {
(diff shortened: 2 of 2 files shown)


lib/UndoByvalPass.cpp
```diff
@@ -144,6 +144,8 @@ PreservedAnalyses clspv::UndoByvalPass::run(Module &M,
 
         CallInst *NewCall = CallInst::Create(NewFunc, Args, "", Call);
         NewCall->setCallingConv(NewFunc->getCallingConv());
+        auto Instr = dyn_cast<Instruction>(User);
+        NewCall->copyMetadata(*Instr);
```
rjodinchr, line 148:
```suggestion
        NewCall->copyMetadata(*Call);
```

test/issue-1228.cl
```diff
@@ -0,0 +1,13 @@
+// RUN: clspv -inline-entry-points -g -O0 %s -o %t.spv 
+
+typedef struct Baz {
+    float x;
+} Baz;
+
+static int foo(Baz baz) {
+	return 1;
```
rjodinchr, line 8: Why did you add this `return 1;`? If you really want to keep it, could we avoid mixing space and tab in the file?

Rekt3421: ah sorry I was doing some debugging and forgot to change it back..

rjodinchr (approved):

PR #1312, opened 2024-03-07
only skip memcpy in runOnUpgradeableConstantCasts

rjodinchr:
We what to upgrade all other function. Especially the calls ending up on atomics operation

### DIFF
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -737,12 +737,19 @@ bool clspv::SimplifyPointerBitcastPass::runOnUpgradeableConstantCasts(
         Value *source = nullptr;
         Type *source_ty = nullptr;
         Type *dest_ty = nullptr;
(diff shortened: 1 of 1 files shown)


lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -737,12 +737,19 @@ bool clspv::SimplifyPointerBitcastPass::runOnUpgradeableConstantCasts(
...
-        // calls like memcpy cannot be simplify by this function, skip them to
-        // avoid having to support them in function like
-        // `BitcastUtils::PointerOperandNum`.
+
+        auto isMemcpy = [](Instruction *I) {
+          auto memcpy = dyn_cast<CallInst>(I);
+          if (memcpy == nullptr)
+            return false;
+          return memcpy->getCalledFunction()->getName().starts_with(
+              "llvm.memcpy");
```
alan-baker, line 746: This might be better checking the intrinsic id.

alan-baker (approved):

PR #1316, opened 2024-03-15
build libclc libraries with clspv

rjodinchr:
Build libclc as part of the build system rather than relying on a checked-in bitcode file. This somewhat increases the duration of the build as clang, opt, etc are built as dependencies.

Ref #826

Based on #903


rjodinchr:
I am working on the google infra as well.

rjodinchr:
Updating llvm now generates vec8 when compiling `pow`. Not sure if it comes from an update in llvm or libclc.

rjodinchr:
Depends on https://github.com/llvm/llvm-project/pull/87622

rjodinchr:
Windows builds have been fixed by llvm/llvm-project@effb2f1917f11b58262d0e13aa085303b5896852

alan-baker (approved):

PR #1332, opened 2024-04-10
Remove NoContraction on fmul & fadd when compiling w/ cl-mad-enable

### LINKED ISSUE #1331: Improve `-cl-mad-enable` support
This is coming from https://github.com/krrishnarraj/clpeak runs.

`clpeak` compiles with `-cl-mad-enable`

`clspv` generates less efficient kernels than other OpenCL compilers because it decorates `OpFMul` and `OpFAdd` with `NoContraction`.

When `-cl-mad-enable` is set, we should not decorate those 2 nodes.

rjodinchr:
Fix #1331

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -302,6 +302,9 @@ bool PrintfSupport();
 // Returns the size of the printf buffer in bytes. The default value is 1MB.
 uint32_t PrintfBufferSize();
 
+// Returns true when compiling with '-cl-mad-enable'
+bool ClMadEnable();
+
 } // namespace Option
 } // namespace clspv
 
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -105,11 +105,6 @@ static llvm::cl::opt<bool>
                    llvm::cl::desc("This option disables all optimizations. The "
                                   "default is optimizations are enabled."));
 
-static llvm::cl::opt<bool> cl_mad_enable(
-    "cl-mad-enable", llvm::cl::init(false),
-    llvm::cl::desc("Allow a * b + c to be replaced by a mad. The mad computes "
-                   "a * b + c with reduced accuracy."));
-
 static llvm::cl::opt<bool> cl_no_signed_zeros(
     "cl-no-signed-zeros", llvm::cl::init(false),
     llvm::cl::desc("Allow optimizations for floating-point arithmetic that "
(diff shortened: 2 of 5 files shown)


oscarbg:
curious, this fix is for all floating point formats (fp16,fp32,fp64) or only fp32?
while waiting to be merged and updated clspv available on clvk so can test via github actions, can share what perf improvements you see in practice?
thanks..

rjodinchr:
> curious, this fix is for all floating point formats (fp16,fp32,fp64) or only fp32? while waiting to be merged and updated clspv available on clvk so can test via github actions, can share what perf improvements you see in practice? thanks..

This is for all floating point types.

Note that performance will be affected only if you are using `-cl-mad-enable` and not either of the following (all of those already preventing `NoContraction` decorations):
- `-cl-native-math`
- `-cl-unsafe-math-optimizations`
- `-cl-fast-relaxed-math`

Here the numbers I get from running `clpeak` on `half4`:
|half4 (GFLOPS)|Intel(R) Graphics (ADL-N)|AMD Radeon Graphics (RADV RAPHAEL_MENDOCINO)|
|---|---|---|
| w/ decorations| 538 | 389|
| w/o decorations| 950 | 638|

oscarbg:
thanks for sharing results !!

> Note that performance will be affected only if you are using -cl-mad-enable and not either of the following (all of those already preventing NoContraction decorations):
> 
>     -cl-native-math
>     -cl-unsafe-math-optimizations
>     -cl-fast-relaxed-math

not clear if this means clpeak source code has to be changed and rebuilt to get your results?

rjodinchr:
> not clear if this means clpeak source code has to be changed and rebuilt to get your results?

clpeak source code will not need to be changed nor rebuilt. The update of clspv will be enough.

What is the project with the github actions running clvk/clpeak you are looking at?

rjodinchr:
> thanks for info.. sorry if not clear I mean downloading latest clvk builds, like: https://github.com/kpet/clvk/actions/runs/8591300393 but for that clvk has to update clspv..

It can take a bit of time for clvk to update its clspv version. As clvk only relies on free github actions, I would advise you to fork clvk to be able to update clspv with the version you want to get builds as early as you want.

I guess you could already have a clvk build with the CL doing so.

alan-baker (approved):

PR #1336, opened 2024-04-19
Update LLVM

rjodinchr:


ben-clayton (approved):
Thank you!

lib/UndoByvalPass.cpp
```diff
@@ -126,7 +126,10 @@ PreservedAnalyses clspv::UndoByvalPass::run(Module &M,
       // Update caller site.
       for (auto User : Users) {
         // Create new call instruction for new function without byval.
-        CallInst *Call = cast<CallInst>(User);
+        CallInst *Call = dyn_cast<CallInst>(User);
```
dneto0, line 129: Yeah, it seems weird to have uses of the function F that aren't calls.  But certainly we're only replacing the calls here, so it's fine to skip. Otherwise the code asserts out anyway.

But I'm curious about whether the test suite actually exercises this case.  And if something does use F, what happens on the erase-from-parent down on line 157?

dneto0 (commented):
LGTM.

dneto0 (approved):
This a net improvement. Approving

Issue #1343, opened 2024-05-02
Error in LinkBuiltinLibrary


bmanga:
The latest clspv results in:
```llvm
target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"; Function Attrs: convergent inlinehint mustprogress norecurse nounwind
define linkonce_odr spir_func float @_ZN4core6frexpfEfPi(float %value, ptr %exp)
; Function Attrs: convergent inlinehint mustprogress norecurse nounwind
define linkonce_odr spir_func float @_ZN4core6frexpfEfPU3AS1i(float %value, ptr addrspace(1) %exp)
; Function Attrs: convergent inlinehint mustprogress norecurse nounwind
define linkonce_odr spir_func float @_ZN4core6frexpfEfPU3AS3i(float %value, ptr addrspace(3) %exp)
@clspv.builtins.used = appending global [3 x ptr] [ptr @_ZN4core6frexpfEfPi, ptr @_ZN4core6frexpfEfPU3AS1i, ptr @_ZN4core6frexpfEfPU3AS3i], section "llvm.metadata"
internal_additional_library:: error: expected '{' in function body
define linkonce_odr spir_func float @_ZN4core6frexpfEfPU3AS1i(float %value, ptr addrspace(1) %exp)
```
I can't share the module I'm trying to link, but the issue might have to do with `frexp`:
https://github.com/google/clspv/blob/118d65ebfd6a747dc588a871535ce8c0da04a88f/lib/Compiler.cpp#L1037

bmanga:
Functions that contain those names (frexp, modf, etc...) in their name and have a correct signature seem to trigger it. For example:
```cpp
void as_long_as_contains_frexp_in_name(float, int *) {}
```
godbolt: https://gcc.godbolt.org/z/aT6a6ExWo

rjodinchr:
I guess [Builtins::BuiltinWithGenericPointer](https://github.com/google/clspv/blob/118d65ebfd6a747dc588a871535ce8c0da04a88f/lib/Builtins.cpp#L779) needs to be improved. Can you submit a PR that fix your usecase?

rjodinchr:
Or maybe you could add functions declaration in `GetEquivalentBuiltinsWithoutGenericPointer` only if they are part of `library` (from `LinkBuiltinLibrary`). Using something like `library->getFunction(<name>)`?

[...]

PR #1344, opened 2024-05-06
Use demangled name for BuiltinWithGenericPointer

bmanga:
This tries to address #1343 

@rjodinchr should I also add a test?

### DIFF
--- a/lib/Builtins.cpp
+++ b/lib/Builtins.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "llvm/Support/Regex.h"
 #include "llvm/Support/raw_ostream.h"
 
 #include "Builtins.h"
@@ -777,11 +778,33 @@ glsl::ExtInst Builtins::getDirectOrIndirectExtInstEnum(
 }
 
 bool Builtins::BuiltinWithGenericPointer(StringRef name) {
-  if (name.contains("fract") || name.contains("frexp") ||
-      name.contains("modf") || name.contains("remquo") ||
-      name.contains("lgamma_r") || name.contains("vstore_half") ||
-      name.contains("sincos"))
-    return true;
+  StringRef prefix("_Z");
+  if (!name.starts_with(prefix)) {
+    return false;
+  }
+  size_t pos = prefix.size();
+  std::string demangled = GetUnmangledName(name.str(), &pos);
+  if (demangled.empty()) {
+    // It could be a local symbol (_ZLxx)
+    return false;
+  }
+
+  const char *list[] = {"fract",  "frexp",    "modf",
+                        "remquo", "lgamma_r", "sincos"};
+
+  for (const auto &fn_name : list) {
+    if (demangled == fn_name) {
+      return true;
+    }
+  }
+
+  // Check for vstore_half family of functions.
+  if (demangled.size() >= strlen("vstore_half")) {
+    static Regex regex("^vstore_half(2|3|4|8|16)?(_rte|_rtz|_rtp|_rtn)?$");
+    if (regex.match(demangled)) {
+      return true;
+    }
+  }
   return false;
 }
 
--- a/test/BuiltinsWithGenericPointer/issue-1343.cl
+++ b/test/BuiltinsWithGenericPointer/issue-1343.cl
@@ -1,0 +1,12 @@
+// RUN: clspv --cl-std=CLC++ --inline-entry-points %s --print-before=annotation-to-metadata --output-format=ll -O3 -o %t.ll
+// RUN: FileCheck %s < %t.ll
+
+// CHECK: define dso_local spir_func float @_Z22contains_frexp_in_namefPU3AS4i(float %0, ptr addrspace(4) %1)
+// CHECK: define dso_local spir_func void @_Z12vstore_half5Dv2_fjPU3AS4Dh(<2 x float> %0, i32 %1, ptr addrspace(4) %2)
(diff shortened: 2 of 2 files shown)


rjodinchr:
> should I also add a test?

It would be great to add the simple example you gave in the issue. Maybe just running clang front-end as in https://github.com/google/clspv/blob/main/test/emit_ir.cl#L1

lib/Builtins.cpp
```diff
@@ -777,11 +778,32 @@ glsl::ExtInst Builtins::getDirectOrIndirectExtInstEnum(
...
 bool Builtins::BuiltinWithGenericPointer(StringRef name) {
-  if (name.contains("fract") || name.contains("frexp") ||
-      name.contains("modf") || name.contains("remquo") ||
-      name.contains("lgamma_r") || name.contains("vstore_half") ||
-      name.contains("sincos"))
-    return true;
+  if (!name.starts_with("_Z")) {
+    return false;
+  }
+  size_t pos = 2;
```
rjodinchr, line 784:
```suggestion
  StringRef prefix("_Z");
  if (!name.starts_with(prefix)) {
    return false;
  }
  size_t pos = prefix.size();
```

lib/Builtins.cpp
```diff
@@ -777,11 +778,32 @@ glsl::ExtInst Builtins::getDirectOrIndirectExtInstEnum(
...
+                        "remquo", "lgamma_r", "sincos"};
+
+  for (const auto &fn_name : list) {
+    if (demangled == fn_name) {
+      return true;
+    }
+  }
+
+  // Check for vstore_half family of functions.
+  if (demangled.size() > 10) {
```
rjodinchr, line 801:
```suggestion
  if (demangled.size() >= strlen("vstore_halfx_rtx") {
```

test/BuiltinsWithGenericPointer/issue-1343.cl
```diff
@@ -0,0 +1,5 @@
+// RUN: clspv --cl-std=CLC++ --inline-entry-points %s -o %t.spv
```
rjodinchr, line 1:
```suggestion
// RUN: clspv --cl-std=CLC++ --inline-entry-points --output-format=ll %s -o %t.ll
```

test/BuiltinsWithGenericPointer/issue-1343.cl
```diff
@@ -0,0 +1,5 @@
...
+
```
rjodinchr, line 2: Please add some checks:
```
// RUN: FileCheck %s < %t.ll
// CHECK: something
```

rjodinchr (changes requested):

rjodinchr:
We don't like adding a test without explicit checks. It makes it quite complicated sometimes when it starts failing for whatever reason to know what the test was expecting to check.

As it's the link that is failing and it is performed after the GenerateIRFile, it makes sense that it does not work with `--output-format`.
What we can do instead is to use `--print-before=annotation-to-metadata` to get the `ll` file after the link. So that we can check that the content is what we expect and nothing more.

bmanga:
Mh I've tried that and it seems like clspv crashes before being able to dump anything to the file...

rjodinchr:
Without the fix, it should crash before anything is dumped. Am I missing something?

bmanga:
Ah ok I misunderstood your previous comment. I thought you wanted the test to fail through FileCheck and I assumed that flag would output something I could act upon without crashing the compiler.

Updated. Let me know if that's ok :)

test/BuiltinsWithGenericPointer/issue-1343.cl
```diff
@@ -0,0 +1,12 @@
+// RUN: clspv --cl-std=CLC++ --inline-entry-points %s --print-before=annotation-to-metadata --output-format=ll -O3 -o %t.ll
+// RUN: FileCheck %s < %t.ll
+
+// CHECK: define dso_local spir_func float @_Z22contains_frexp_in_namefPU3AS4i(float %0, ptr addrspace(4) %1)
+// CHECK: define dso_local spir_func void @_Z12vstore_half5Dv2_fjPU3AS4Dh(<2 x float> %0, i32 %1, ptr addrspace(4) %2)
+// CHECK: define dso_local spir_func void @_Z13xvstore_half2Dv2_fjPU3AS4Dh(<2 x float> %0, i32 %1, ptr addrspace(4) %2)
+
```
rjodinchr, line 7: The way it is right now is not failing if I do not use the version of `BuiltinWithGenericPointer` in this CL.
```suggestion
// RUN: clspv --cl-std=CLC++ --inline-entry-points %s --print-before=annotation-to-metadata -o %t.spv &> %t.ll
// RUN: FileCheck %s < %t.ll
// RUN: spirv-val %t.spv

// CHECK: define dso_local spir_func float @_Z22contains_frexp_in_namefPU3AS4i(float %0, ptr addrspace(4) %1)
// CHECK: define dso_local spir_func void @_Z12vstore_half5Dv2_fjPU3AS4Dh(<2 x float> %0, i32 %1, ptr addrspace(4) %2)
// CHECK: define dso_local spir_func void @_Z13xvstore_half2Dv2_fjPU3AS4Dh(<2 x float> %0, i32 %1, ptr addrspace(4) %2)

void kernel foo() {}
```

rjodinchr (changes requested):

test/BuiltinsWithGenericPointer/issue-1343.cl
bmanga: Ah ok... I'm pretty sure I tested it but maybe I got something wrong. I'll commit your suggestion :)

PR #1345, opened 2024-05-07
Support reference types in IsSupportedType

### LINKED ISSUE #1342: Issue with C++ references
Some code triggers problems with references (I can't seem to be able to trigger it with a small example, sorry!), with an assertion 
```
IsSupportedType lacks support for QualType: const __generic core::Vector3f &
Type not covered by IsSupportedType.
UNREACHABLE executed at [...]/lib/FrontendPlugin.cpp:288!
```

From a quick look the `IsSupportedType` function only checks for pointer types, ignoring reference types:
https://github.com/google/clspv/blob/118d65ebfd6a747dc588a871535ce8c0da04a88f/lib/FrontendPlugin.cpp#L194
Changing the check to
```cpp
if (Ty->isPointerType() || Ty->isReferenceType()) 
```
seems to overcome the issue.

bmanga:
Closes #1342

### DIFF
--- a/lib/FrontendPlugin.cpp
+++ b/lib/FrontendPlugin.cpp
@@ -191,7 +191,7 @@ private:
     }
 
     // First check if we have a pointer type.
-    if (Ty->isPointerType()) {
+    if (Ty->isPointerType() || (!IsKernelParameter && Ty->isReferenceType())) {
       const Type *pointeeTy = Ty->getPointeeType().getTypePtr();
       if (pointeeTy && pointeeTy->isVoidType() &&
           !clspv::Option::Int8Support()) {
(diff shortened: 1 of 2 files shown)


rjodinchr:
Could you add a test please?

bmanga:
@rjodinchr done :)

lib/FrontendPlugin.cpp
```diff
@@ -191,7 +191,7 @@ struct ExtraValidationConsumer final : public ASTConsumer {
     }
 
     // First check if we have a pointer type.
-    if (Ty->isPointerType()) {
+    if (Ty->isPointerType() || (!IsKernelParameter && Ty->isReferenceType())) {
```
alan-baker, line 194: Is `IsKernelParameter` required? Wouldn't kernel argument references be rejected elsewhere in clang?

bmanga: Doen't look like it: https://gcc.godbolt.org/z/fE84j8s7q

alan-baker (approved):

PR #1348, opened 2024-05-12
Support PHI in LowerAddrSpaceCastPass

### LINKED ISSUE #1341: Segfault with default initialization
Given the following code:

```cpp
struct Vector4 {
    float data[4] = {};
};

kernel void fun() {
#ifdef FAIL
    Vector4 m;
#else
    Vector4 m = {};
#endif
}
```

Compiling using `-cl-std=CLC++ --inline-entry-points  -DFAIL` results in 
```
Instruction not handled:   %arrayinit.cur.i = phi ptr addrspace(4) [ %m.ascast, %entry ], [ %arrayinit.next.i, %arrayinit.body.i ]
```

I'm assuming this has something to do with where the initializer data is stored and is related to #1288 , but I wonder if this is easier to fix.

bmanga:
* Adds support for phi instructions in LowerAddrSpaceCast
* Adds code to remove dead phi instructions (overcomes the circular dependency)
* Adds test

Note that I have made the assumption that the first incoming value to the phi node will not depend on the phi node itself... I guess this could be improved in the future.

Closes #1341

### DIFF
--- a/test/AddressSpaceCast/issue-1341.ll
+++ b/test/AddressSpaceCast/issue-1341.ll
@@ -1,7 +1,8 @@
 ; RUN: clspv-opt %s -o %t.ll --passes=lower-addrspacecast
 ; RUN: FileCheck %s < %t.ll
 
-; CHECK: %0 = phi ptr [ %{{.*}}, %entry ], [ %{{.*}}, %arrayinit.body ]
+; CHECK: arrayinit.body:
+; CHECK-NEXT: {{.*}} = phi ptr [ %{{.*}}, %entry ], [ %{{.*}}, %arrayinit.body ]
 ; CHECK-NOT: %arrayinit.cur = phi ptr addrspace(4) [ %arrayinit.begin, %entry ], [ %arrayinit.next, %arrayinit.body ]
 
 define dso_local spir_kernel void @fun() {
--- a/test/AddressSpaceCast/issue-1341.ll
+++ b/test/AddressSpaceCast/issue-1341.ll
@@ -1,0 +1,23 @@
+; RUN: clspv-opt %s -o %t.ll --passes=lower-addrspacecast
+; RUN: FileCheck %s < %t.ll
+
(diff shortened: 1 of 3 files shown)


test/AddressSpaceCast/issue-1341.ll
```diff
@@ -0,0 +1,22 @@
+; RUN: clspv-opt %s -o %t.ll --passes=lower-addrspacecast
+; RUN: FileCheck %s < %t.ll
+
+; CHECK: %0 = phi ptr [ %arrayinit.begin1, %entry ], [ %arrayinit.next4, %arrayinit.body ]
```
rjodinchr, line 4: `%arrayinit.begin1` and `%arrayinit.next4` can easily change, so we should not rely on those hardcoded strings.

rjodinchr (changes requested):

test/AddressSpaceCast/issue-1341.ll
bmanga: Fixed. Also added check to ensure the old phi is removed.

test/AddressSpaceCast/issue-1341.ll
```diff
@@ -0,0 +1,23 @@
+; RUN: clspv-opt %s -o %t.ll --passes=lower-addrspacecast
+; RUN: FileCheck %s < %t.ll
+
+; CHECK: %0 = phi ptr [ %{{.*}}, %entry ], [ %{{.*}}, %arrayinit.body ]
```
rjodinchr, line 4: We should also not rely on `%0`
```suggestion
; CHECK: arrayinit.body:
; CHECK-NEXT: {{.*}} = phi ptr [ %{{.*}}, %entry ], [ %{{.*}}, %arrayinit.body ]
```

rjodinchr (changes requested):

rjodinchr (approved):

PR #1349, opened 2024-05-14
Update libclc builtin after changes in llvm

rjodinchr:
https://github.com/llvm/llvm-project/pull/84934


rjodinchr:
https://github.com/llvm/llvm-project/pull/92126

rjodinchr:
Once https://github.com/llvm/llvm-project/pull/92126 is merged, I'll update the llvm version here and this PR will be ready for review.

It  will be needed for google internal infra, as the llvm update will break clspv tests

PR #1365, opened 2024-05-26
Improve handling of PHI in LowerAddrSpaceCast

### LINKED ISSUE #1364: Handle PHINode in LowerAddrSpaceCast better
#1348 added initial support for phi nodes in LowerAddrSpaceCast.
However, it made the assumption that the first entry was never dependent on the node itself and could be used to figure out the type of the replacement node. So get a crash with an alternative solution.

bmanga:
Adds logic to check for phi dependent operands
Figures out the phi type from non-dependent types
Test with dependent incoming value first

It also keeps the phi node valid if the incoming values have different address spaces by re-inserting some addrspacecasts. It will crash at the spirv producer but an additional pass could try and fix it up before then.

Closes #1364

### DIFF
--- a/lib/LowerAddrSpaceCastPass.cpp
+++ b/lib/LowerAddrSpaceCastPass.cpp
@@ -350,26 +350,32 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
   return ptrToInt;
 }
 
-static bool DependsOnPhiNode(const Value *Value, const PHINode *PhiNode,
-                             DenseSet<const llvm::Value *> &Visited) {
-  if (!isa<Instruction>(Value)) {
-    return false;
-  }
-  if (Visited.contains(Value)) {
+static bool DependsOnPhiNode(const Value *V, const PHINode *PhiNode) {
+  DenseSet<const llvm::Value *> Visited;
+  if (!isa<Instruction>(V)) {
     return false;
   }
 
-  Visited.insert(Value);
+  SmallVector<const Value *, 16> Stack;
+  Stack.push_back(V);
 
-  // If the value is the PHI node itself, return true
-  if (Value == PhiNode) {
-    return true;
-  }
+  while (!Stack.empty()) {
+    const Value *Current = Stack.pop_back_val();
+    if (Visited.contains(Current)) {
+      continue;
+    }
+    Visited.insert(Current);
 
-  // Recursively check if any of the operands depend on the PHI node
-  for (const auto *Op : cast<Instruction>(Value)->operand_values()) {
-    if (DependsOnPhiNode(Op, PhiNode, Visited)) {
+    if (Current == PhiNode) {
       return true;
+    }
+
+    if (const auto *Inst = dyn_cast<Instruction>(Current)) {
+      for (const auto *Op : Inst->operand_values()) {
+        if (!Visited.contains(Op)) {
+          Stack.push_back(Op);
+        }
+      }
     }
   }
 
@@ -377,13 +383,7 @@ static bool DependsOnPhiNode(const Value *Value, const PHINode *PhiNode,
 }
 
 Value *clspv::LowerAddrSpaceCastPass::visitPHINode(llvm::PHINode &I) {
-  if (!isGenericPTy(I.getType())) {
-    // Nothing to do.
-    return &I;
-  }
-
   IRBuilder<> B(&I);
-
   unsigned N = I.getNumIncomingValues();
 
   // Analyse the incoming values anche check whether the types agree.
@@ -391,17 +391,13 @@ Value *clspv::LowerAddrSpaceCastPass::visitPHINode(llvm::PHINode &I) {
   // lead to infinite recursion. Instead, we delay processing them until
   // after we have registered the replacement.
   SmallVector<Value *, 2> Replacements(N);
-
-  DenseSet<const llvm::Value *> VisitedForSelfDep;
   SmallVector<unsigned, 2> DependentOperands;
   Type *CommonTy = nullptr;
   bool HasCommonTy = true;
 
   for (unsigned j = 0; j < N; ++j) {
-    VisitedForSelfDep.clear();
     auto *V = I.getIncomingValue(j);
-
-    if (DependsOnPhiNode(V, &I, VisitedForSelfDep)) {
+    if (DependsOnPhiNode(V, &I)) {
       DependentOperands.push_back(j);
     } else {
       V = visit(V);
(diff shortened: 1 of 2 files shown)


lib/LowerAddrSpaceCastPass.cpp
```diff
@@ -350,20 +350,92 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
...
+  if (Value == PhiNode) {
+    return true;
+  }
+
+  // Recursively check if any of the operands depend on the PHI node
+  for (const auto *Op : cast<Instruction>(Value)->operand_values()) {
+    if (DependsOnPhiNode(Op, PhiNode, Visited)) {
+      return true;
+    }
+  }
```
rjodinchr, line 374: I would prefer to avoid recursion when possible

lib/LowerAddrSpaceCastPass.cpp
```diff
@@ -350,20 +350,92 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
...
+
+  return false;
+}
+
 Value *clspv::LowerAddrSpaceCastPass::visitPHINode(llvm::PHINode &I) {
+  if (!isGenericPTy(I.getType())) {
+    // Nothing to do.
+    return &I;
+  }
```
rjodinchr, line 383: Is it really needed? do you have an example for it?

lib/LowerAddrSpaceCastPass.cpp
```diff
@@ -350,20 +350,92 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
...
+    CommonTy = B.getPtrTy(clspv::AddressSpace::Generic);
+    for (unsigned j = 0; j < N; ++j) {
+      if (Replacements[j]->getType() != CommonTy) {
+        Replacements[j] = B.CreateAddrSpaceCast(Replacements[j], CommonTy);
+      }
+    }
+  }
+
+  auto *Phi = B.CreatePHI(CommonTy, N);
+
```
rjodinchr, line 428: The goal is to create a phi that is not using generic addrspace. I don't understand what's happening here then.

lib/LowerAddrSpaceCastPass.cpp
```diff
@@ -350,20 +350,92 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
...
-  for (unsigned i = 1; i < I.getNumIncomingValues(); ++i) {
-    Phi->addIncoming(visit(I.getIncomingValue(i)), I.getIncomingBlock(i));
+  // Now that we have registered the replacement, we can process the dependent
+  // operands.
+  for (auto i : DependentOperands) {
+    Replacements[i] = visit(Replacements[i]);
+  }
+
+  for (unsigned i = 0; i < I.getNumIncomingValues(); ++i) {
+    Phi->addIncoming(Replacements[i], I.getIncomingBlock(i));
```
rjodinchr, line 438: We have potentially created a generic phi node, and now we are visiting the replacement which may lead to incoming value with other addrspace making the phi illegal?

rjodinchr (changes requested):
It feels like more tests are needed to cover the different scenario

lib/LowerAddrSpaceCastPass.cpp
bmanga: So, if the phi types have been resolved to different address spaces (the code has a conditional assignment from different address spaces), we can't fix this phi node at the moment. So they are added back to keep the node valid, but we know that generic phi nodes after this pass are there because of these conditional assignments.

The reason I did this is because I'm working on a pass that will try to fix them later on (I'll post a draft PR soon :) )

lib/LowerAddrSpaceCastPass.cpp
bmanga: I guess the visit could be skipped here, but I just thought it might simplify something before giving up when it hits the phi node again (as they are dependent).
I could skip it if you prefer

rjodinchr: If you do not have an example where it is useful, please remove.

lib/LowerAddrSpaceCastPass.cpp
rjodinchr: The goal of this pass is to remove generic addrspace. Could this be implemented in that pass then?

lib/LowerAddrSpaceCastPass.cpp
bmanga: I've removed the handling for not common types for now.
Visiting the nodes after we register the replacement is still required though to simplify the dependent instructions.

lib/LowerAddrSpaceCastPass.cpp
bmanga: It can't because this pass would crash, but I've removed this for now. I'll post a new implementation in a separate PR

lib/LowerAddrSpaceCastPass.cpp
bmanga: Replaced

lib/LowerAddrSpaceCastPass.cpp
```diff
@@ -350,20 +350,87 @@ Value *clspv::LowerAddrSpaceCastPass::visitPtrToIntInst(PtrToIntInst &I) {
...
-  auto Phi = B.CreatePHI(V1->getType(), I.getNumIncomingValues());
+  unsigned N = I.getNumIncomingValues();
+
+  // Analyse the incoming values anche check whether the types agree.
+  // Ignore any incoming value that depends on the node itself, as that would
+  // lead to infinite recursion. Instead, we delay processing them until
+  // after we have registered the replacement.
+  SmallVector<Value *, 2> Replacements(N);
+
+  DenseSet<const llvm::Value *> VisitedForSelfDep;
```
rjodinchr, line 396: This variable is only used inside `DependsOnPhiNode`, I think it should be defined there, not here.

bmanga: This way we avoid re-allocating the storage for the DenseSet and it's cleared between calls. But I can move it into the function if you don't think it's worthwhile

rjodinchr: Keeping the code as readable as possible is better in my opinion.

bmanga: Ok, changed :)

rjodinchr (approved):

PR #1373, opened 2024-06-21
fix fabs for fp16

rjodinchr:


alan-baker (commented):
Should we fix f64 while we're at it?

rjodinchr:
According to clspv [documentation](https://github.com/google/clspv/blob/main/docs/OpenCLCOnVulkan.md#64-bit-types), fp64 is not supported. I am never looking at it closely. Does clspv support fp64?

alan-baker:
Probably not well. I think it could be attempted by enabling the extension, but it's unclear what we'd generate. Let's leave since we document a lack of support.

PR #1380, opened 2024-07-03
Fix trunc to bool handling

### LINKED ISSUE #1379: Incorrect SPIR-V conversion for trunc to i1
The current code in the SPIRV-Producer pass for `trunc` to `i1` converts it into `!= 0`, but the LLVM langref says:
> The ‘trunc’ instruction truncates the high order bits in value and converts the remaining bits to ty2.

This means that only bit 0 in the input should determine the result.

For example, the kernel:
```
__kernel void test(__global float* out)
{
    size_t i = get_global_id(0);

    float foo = 0;
    switch (i & 1)
    {
        case 0:
            foo = i;
            break;
    }
    out[i] = foo;
}
```
should write the global_id to even numbered locations in `out`, but will only write it to `out[0]`.

gnl21:
LLVM defines '%trunc = trunc i32 %2 to i1' as discarding the high bits of %2 prior to conversion, but the SPIRV-Producer pass was implementing it as '%2 != 0'. And away the high bits before doing the conversion.

Fixes #1379.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -4768,8 +4768,12 @@ void SPIRVProducerPassImpl::GenerateInstruction(Instruction &I) {
         // But OpUConvert only takes integer types in input, not boolean type.
         // Instead use OpINotEqual.
         SPIRVOperandVec Ops;
-        Ops << Ty << I.getOperand(0)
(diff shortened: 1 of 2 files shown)


rjodinchr:
> It looks like the MSVC autobuilder isn't working? It's complaining about the Python version available, so I don't think it can be related to this change.

This is a known issue, it's not related to this change.

Issue #1371, opened 2024-06-21
printf does not handle "\0"


rjodinchr:
https://compiler-explorer.com/z/Thv1cjGr9

The following kernel does not generate any nodes for `"\0"`:
```
__kernel void test_kernel()
{
    printf("%s", "\0");
    printf("%s", "test");
}
```

dneto0:
This causes the test_printf conformance test to fail because has a case like `"foo\0foo"`

https://github.com/KhronosGroup/OpenCL-CTS/blob/070052c890c83053c5243e32b5bec8b84b04344d/test_conformance/printf/util_printf.cpp#L749

Rekt3421:
taking a look at this now.

PR #1384, opened 2024-07-18
Support printf string args with nul

dneto0:
Add a test

Fixed: #1371


Rekt3421 (approved):
Thank you!

rjodinchr (approved):
Tests are failing because `test/Printf/printf_string_64bit_ptr.ll` needs to be updated (adding/removing the `\00` in it)

Issue #1396, opened 2024-09-11
Char* has wrong stride


Rekt3421:
```
/**
 * This kernel function only fills a buffer with the sentence 'Hello World!'.
 **/

__kernel void helloWorld(__global char* data){
    data[0] = 'H';
    data[1] = 'e';
    data[2] = 'l';
    data[3] = 'l';
    data[4] = 'o';
    data[5] = ' ';
    data[6] = 'W';
    data[7] = 'o';
    data[8] = 'r';
    data[9] = 'l';
    data[10] = 'd';
    data[11] = '!';
    data[12] = '\n';
    data[13] = 0;
}
```
it generates the following error

```
error: 28: Structure id 10 decorated as Block for variable in StorageBuffer storage class must follow relaxed storage buffer layout rules: member 0 contains an array with stride 1 not satisfying alignment to 4
  %_struct_10 = OpTypeStruct %_runtimearr_uint

Build Status: -2Build Log:	 clvk-RoBPoh/source.cl:22:1: warning: null character ignored
   22 | <U+0000>
      | ^
clvk-RoBPoh/source.cl:22:2: warning: no newline at end of file
   22 | <U+0000>
      |         ^
changing __global char* type to __global int* resolves the issue.
```

rjodinchr:
Ref https://github.com/kpet/clvk/issues/716

PR #1397, opened 2024-09-12
change stride to 4 instead of 1 if int8 is not supported

Rekt3421:
please see issue #1396 for more details

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -2091,8 +2091,8 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVType(Type *Ty, bool needs_layout) {
         // Ops[2] = Stride Number(Literal Number)
         Ops.clear();
 
-        // if stride is 1 automatically set it to 2 to avoid stride issues.
         auto CurrStride = static_cast<uint32_t>(GetTypeAllocSize(EleTy, DL));
+        // if stride is 1 automatically set it to 4 to avoid stride issues.
         Ops << RID << spv::DecorationArrayStride
             << getCorrectedStride(CurrStride);
 
@@ -6014,6 +6014,7 @@ void SPIRVProducerPassImpl::HandleDeferredDecorations() {
     // Ops[1] = Decoration (ArrayStride)
     // Ops[2] = Stride number (Literal Number)
     SPIRVOperandVec Ops;
+    // if stride is 1 automatically set it to 4 to avoid stride issues.
     Ops << id << spv::DecorationArrayStride << getCorrectedStride(stride);
 
     addSPIRVInst<kAnnotations>(spv::OpDecorate, Ops);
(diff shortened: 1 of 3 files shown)


Rekt3421:
Is a unit test needed for this? Seems a little trivial @rjodinchr ?

rjodinchr (changes requested):
This is not the only place where we create `spv::DecorationArrayStride`.
I would like to have every place fixed.

Also, why `2`? when int8 is not supported, it is converted to int32 not int16.

We would definitely need tests here. At least the one from  the issue.

rjodinchr:
This one feels like a good one as well:
```
// RUN: clspv -int8=0 %target %s -o %t.spv
// RUN: spirv-dis -o %t2.spvasm %t.spv
// RUN: FileCheck %s < %t2.spvasm
// RUN: spirv-val --target-env vulkan1.0 %t.spv

constant uchar b[4] = {[0]=42, [1]=13, [2]=0, [3]=5};

void kernel __attribute__((reqd_work_group_size(4, 1, 1))) foo(global uchar* a)
{
  *a = b[get_local_id(0)];
}

// CHECK-DAG:  [[uint:%[^ ]+]] = OpTypeInt 32 0
// CHECK-DAG:  OpTypePointer StorageBuffer [[struct:%[^ ]+]]
// CHECK-DAG:  [[struct]] = OpTypeStruct [[runtimearr:%[^ ]+]]
// CHECK-DAG:  OpDecorate [[runtimearr]] ArrayStride 4
```
You will note that even if you change `ArrayStride 4` to `ArrayStride 2` in the test, it is still not passing with 2. It would be failing with the following message:
```
Structure id 15 decorated as Block for variable in StorageBuffer storage class must follow standard storage buffer layout rules: member 0 contains an array with stride 2 not satisfying alignment to 4
```

Rekt3421:
Will add the test from the issue as well.

lib/SPIRVProducerPass.cpp
```diff
@@ -2084,8 +2084,12 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVType(Type *Ty, bool needs_layout) {
         // Ops[2] = Stride Number(Literal Number)
         Ops.clear();
 
-        Ops << RID << spv::DecorationArrayStride
-            << static_cast<uint32_t>(GetTypeAllocSize(EleTy, DL));
+        // if stride is 1 automatically set it to 2 to avoid stride issues.
+        auto CurrStride = static_cast<uint32_t>(GetTypeAllocSize(EleTy, DL));
+        if (CurrStride == 1 && !Int8Support()) {
+          CurrStride = 4;
+        }
```
rjodinchr, line 2091: Please avoid code duplication. Make a function for it.

rjodinchr (changes requested):

lib/SPIRVProducerPass.cpp
```diff
@@ -2084,8 +2091,10 @@ SPIRVID SPIRVProducerPassImpl::getSPIRVType(Type *Ty, bool needs_layout) {
         // Ops[2] = Stride Number(Literal Number)
         Ops.clear();
 
+        // if stride is 1 automatically set it to 2 to avoid stride issues.
```
rjodinchr, line 2094: The comment is not correct anymore, please update it.

Rekt3421: thank you!

rjodinchr:
The `macos` build are failing because this branch has not been rebase on ToT.

PR #1398, opened 2024-09-16
fix runOnUpgradeableConstantCasts

rjodinchr:
Get the right operand number when looking at phi nodes.


lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -929,6 +929,13 @@ bool clspv::SimplifyPointerBitcastPass::runOnUpgradeableConstantCasts(
         GetElementPtrInst::Create(dest_ty, ptr, NewGEPIdxs, "", GEPInfo.gep);
 
     unsigned PointerOperandNum = BitcastUtils::PointerOperandNum(I);
+    if (auto phi = dyn_cast<PHINode>(I)) {
```
alan-baker, line 932: Before structurizing the CFG, there might be some odd loops such that phis have more than two sets of operands. There could a case where the same gep appears more than once in such a phi. Will that case get handled in a subsequent loop or would we lose the necessary info?

rjodinchr: I have updated the test, it is taken into account already

alan-baker (approved):

Issue #1400, opened 2024-09-17
Add OpLine Information for Atomic Instructions


CapZTr:
Hi! `clspv` does not generate OpLine information for any atomic instructions so far. Does someone have a idea of how it can be fixed?

rjodinchr:
I think `clspv` is not passing the debug information somewhere in the flow.

Taking an example and compiling it with `--print-before-all` should help identify the pass.

@CapZTr Do you want to have a look yourself? Otherwise @Rekt3421 can find sometime to look at it I think.

CapZTr:
Hey @rjodinchr , thanks for the quick reply! This issue is fixed and I created the PR #1403 for it:-)

PR #1403, opened 2024-09-18
keep source-level location also for atomic instructions

CapZTr:
Fixed the bug #1400 that missing `OpLine` information for atomic instructions

### DIFF
--- a/lib/SPIRVOp.cpp
+++ b/lib/SPIRVOp.cpp
@@ -30,7 +30,7 @@ Instruction *InsertSPIRVOp(Instruction *Insert, spv::Op Opcode,
                            ArrayRef<Attribute::AttrKind> Attributes,
                            Type *RetType, ArrayRef<Value *> Args,
                            const MemoryEffects &MemEffects) {
-  
+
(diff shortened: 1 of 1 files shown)


lib/SPIRVOp.cpp
```diff
@@ -65,7 +68,14 @@ Instruction *InsertSPIRVOp(Instruction *Insert, spv::Op Opcode,
     ArgValues.push_back(Arg);
   }
 
-  return CallInst::Create(func, ArgValues, "", Insert);
+  Instruction *NewInst = CallInst::Create(func, ArgValues, "", Insert);
+
+  // Set the source location for the new instruction
+  if (DbgLoc) {
+    NewInst->setDebugLoc(DbgLoc);
+  }
```
rjodinchr, line 76:
```suggestion
  NewInst->setDebugLoc(Insert->getDebugLoc());
```
I don't think there is a reason to make it conditional.

rjodinchr (changes requested):

rjodinchr (approved):

PR #1406, opened 2024-10-15
rename Intrinsic::getDeclaration to Intrinsic::getOrInsertDeclaration

dneto0:
Update LLVM
    
    - Rename Intrinsic::getDeclaration to Intrinsic::getOrInsertDeclaration
    - Update one test to avoid disabling int8

Bug: https://crbug.com/373696543


dneto0:
This is needed after LLVM change https://github.com/llvm/llvm-project/pull/111752

test/ProgramScopeConstants/in_storage_buffer_descriptor_map.cl
```diff
@@ -1,4 +1,4 @@
-// RUN: clspv %target %s -o %t.spv -module-constants-in-storage-buffer -int8=0
```
dneto0, line 1: @rjodinchr observed that Clang is inserting explicit padding members.

In this case we now have  an i8 first member, a [3 x i8] inserted second member.  Everything breaks when artificially trying to remap i8 to u32.  Use the default, which is actually support i8.

rjodinchr (approved):

PR #1409, opened 2024-10-17
windows bot: fix python version

rjodinchr:


dneto0 (approved):
Nice!

I had an open issue to update the VM itself, but .... never got around to it.

PR #1418, opened 2024-11-10
The compiler produce invalid spirv with enabled PhysicalPointerArgsPass.

AlexDemydenko:
The tool spirv-val return errors for some spirv binaries.

The test list with OpenCL-CTS:
```
vector_swizzle
```

Kernel example:
[vector-swizzle.txt](https://github.com/user-attachments/files/17692645/vector-swizzle.txt)

### DIFF
--- a/lib/ReplacePointerBitcastPass.cpp
+++ b/lib/ReplacePointerBitcastPass.cpp
@@ -354,6 +354,19 @@ void ComputeStore(IRBuilder<> &Builder, StoreInst *ST, Value *OrgGEPIdx,
   unsigned DstEleTyBitWidth = SizeInBits(Builder, DstEleTy);
 
   Type *OrigSrcTy = SrcTy;
+
+  // PhysicalPointerArgsPass uses IntToPtrInst to convert an integer argument
+  // that contains a pointer value into a pointer variable back.
+  // Other LLVM passes can change the pointer types into other pointer types.
+  // For example the type char4* into int*.
+  // The problem is that spirv creates a global variable for the pointer
+  // argument, and the source argument type for all getelementptr must be
+  // synchronized to bind them to the global variable.
+  if (dyn_cast<IntToPtrInst>(Src)) {
+    DenseMap<Value *, Type *> TypeCache;
+    Type *arg_type = clspv::InferType(Src, Builder.getContext(), &TypeCache);
+    OrigSrcTy = arg_type ? arg_type : OrigSrcTy;
+  }
   SmallVector<Value *, 4> AddrIdxs;
   ReduceType(Builder, IsGEPUser, OrgGEPIdx, SrcTy, DstTyBitWidth, NewAddrIdxs,
              AddrIdxs, ToBeDeleted);
--- a/test/PhysicalStorageBuffers/physical_pointers_vector-swizzle.cl
+++ b/test/PhysicalStorageBuffers/physical_pointers_vector-swizzle.cl
@@ -1,0 +1,23 @@
+// RUN: clspv  %s -o %t.spv -cl-std=CL3.0 -no-8bit-storage=pushconstant -no-16bit-storage=pushconstant -spv-version=1.6 -arch=spir64 -physical-storage-buffers
+// RUN: spirv-dis %t.spv -o %t.spvasm
+// RUN: FileCheck %s < %t.spvasm
+// RUN: spirv-val --target-env vulkan1.3spv1.6 %t.spv
+
+// CHECK-DAG: [[uint:%[a-zA-Z0-9_]+]] = OpTypeInt 32 0
+// CHECK-DAG: [[uchar:%[a-zA-Z0-9_]+]] = OpTypeInt 8 0
+// CHECK-DAG: [[v4uchar:%[a-zA-Z0-9_]+]] = OpTypeVector [[uchar]] 4
+// CHECK-DAG: [[ptr_SB_v4uchar:%[a-zA-Z0-9_]+]] = OpTypePointer PhysicalStorageBuffer [[v4uchar]]
+// CHECK-DAG: [[uint_1:%[a-zA-Z0-9_]+]] = OpConstant [[uint]] 1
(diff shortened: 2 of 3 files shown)


alan-baker:
What kinds of errors? When I run:
```
clspv vector-swizzle.cl -o vector-swizzle.spv -int8=1 -physical-storage-buffers -arch=spir64
```

The resulting binary passes spirv-val. What command line are you using?

rjodinchr:
This is at least missing a test I would say.

AlexDemydenko:
Command line from the clvk driver:
```
clspv -DTYPE=char4 -DBASETYPE=char -cl-std=CL3.0 -inline-entry-points -cl-single-precision-constant -cl-kernel-arg-info -no-8bit-storage=pushconstant -no-16bit-storage=pushconstant -rounding-mode-rte=16,32 -fp64=0 -rewrite-packed-structs -std430-ubo-layout -decorate-nonuniform -arch=spir64 --use-native-builtins=ceil,exp2,floor,fma,fmax,fmin,half_exp2,half_log2,half_rsqrt,half_sqrt,isequal,isgreater,isgreaterequal,isless,islessequal,islessgreater,isnotequal,log2,mad,round,rsqrt,sqrt, -spv-version=1.6 -max-pushconstant-size=128 -max-ubo-size=4294967295 -global-offset -long-vector -module-constants-in-storage-buffer -cl-arm-non-uniform-work-group-size -enable-feature-macros=__opencl_c_images,__opencl_c_3d_image_writes,__opencl_c_read_write_images,__opencl_c_atomic_order_acq_rel,__opencl_c_atomic_scope_device,__opencl_c_subgroups,__opencl_c_int64 -enable-printf -printf-buffer-size=1048576 -arch=spir64 -physical-storage-buffers vector-swizzle.cl
```

Error from spirv-val:
```
error: line 99: OpPtrAccessChain result type (OpTypeInt) does not match the type that results from indexing into the base <id> (OpTypeVector).
  %56 = OpPtrAccessChain %_ptr_PhysicalStorageBuffer_uint %37 %uint_4
```

AlexDemydenko:
ping

lib/ReplacePointerBitcastPass.cpp
```diff
@@ -354,6 +354,11 @@ void ComputeStore(IRBuilder<> &Builder, StoreInst *ST, Value *OrgGEPIdx,
   unsigned DstEleTyBitWidth = SizeInBits(Builder, DstEleTy);
 
   Type *OrigSrcTy = SrcTy;
+  if (dyn_cast<IntToPtrInst>(Src)) {
```
rjodinchr, line 357: Could you add a comment to explain what is happening here, and why it is needed

rjodinchr (commented):
Could you also add another test using `clspv-opt` targeting the `replace-pointer-bitcast` pass only.
Thank you

lib/ReplacePointerBitcastPass.cpp
AlexDemydenko: Added a comment with explanation.

lpavank:
@rjodinchr and @alan-baker any further inputs for @AlexDemydenko here?

rjodinchr:
> @rjodinchr and @alan-baker any further inputs for @AlexDemydenko here?

Maybe you have missed my comment from last month: https://github.com/google/clspv/pull/1418#pullrequestreview-2453475977
"Could you also add another test using `clspv-opt` targeting the `replace-pointer-bitcast` pass only."

lpavank:
> > @rjodinchr and @alan-baker any further inputs for @AlexDemydenko here?
> 
> Maybe you have missed my comment from last month: [#1418 (review)](https://github.com/google/clspv/pull/1418#pullrequestreview-2453475977) "Could you also add another test using `clspv-opt` targeting the `replace-pointer-bitcast` pass only."

@rjodinchr believe @AlexDemydenko did answer (https://github.com/google/clspv/pull/1418#discussion_r1890121743). Please confirm if thats what you wanted or no.
PS: Just trying to facilitate and/or get this change checked since its blocking one test on our end.

rjodinchr:
> > > @rjodinchr and @alan-baker any further inputs for @AlexDemydenko here?
> > 
> > 
> > Maybe you have missed my comment from last month: [#1418 (review)](https://github.com/google/clspv/pull/1418#pullrequestreview-2453475977) "Could you also add another test using `clspv-opt` targeting the `replace-pointer-bitcast` pass only."
> 
> @rjodinchr believe @AlexDemydenko did answer ([#1418 (comment)](https://github.com/google/clspv/pull/1418#discussion_r1890121743)). Please confirm if thats what you wanted or no. PS: Just trying to facilitate and/or get this change checked since its blocking one test on our end.

At the moment, both added tests are not passing (on every bots). Please make sure they are passing using ToT.
Maybe you can also reduce the `.ll` test to only the necessary part so that it's easier to maintain (here is an example of a reduced test: https://github.com/google/clspv/blob/main/test/PointerCasts/opaque_implicit_gep.ll)

test/PointerCasts/physical_pointers_vector-swizzle.ll
```diff
@@ -0,0 +1,31 @@
+; RUN: clspv-opt %s -o %t.ll --passes=replace-pointer-bitcast
```
rjodinchr, line 1: That test is passing even without your fix, so it is not testing what you expect.

test/PhysicalStorageBuffers/physical_pointers_vector-swizzle.cl
```diff
@@ -0,0 +1,22 @@
+// RUN: clspv  %s -o %t.spv -cl-std=CL3.0 -no-8bit-storage=pushconstant -no-16bit-storage=pushconstant -spv-version=1.6 -arch=spir64 -physical-storage-buffers
```
rjodinchr, line 1: That test is still not passing (with or without the fix)

rjodinchr:
Please run the tests locally and make sure they fail without the fix and pass with it

rjodinchr:
I had a closer look at the PR. I don't think this is the right way to deal with that issue.
Here is what I propose: https://github.com/google/clspv/pull/1434

AlexDemydenko:
Yes, you propose works for me.

PR #1432, opened 2024-12-23
Add --validate option to amber/run_tests.py

dneto0:


dneto0:
This marginally helped my testing/debugging flow.

alan-baker (changes requested):
It looks like this changes the Amber bot to not run validation anymore. Suggest adding `--validate` to the `build-amber.sh` kokoro script.

dneto0:
> It looks like this changes the Amber bot to not run validation anymore. Suggest adding `--validate` to the `build-amber.sh` kokoro script.

that doesn't seem right?   Amber's `-d` option *disables* validation.  So before this change, the run_test.py never enabled validation.   

But I'm checking with a local compile.

PR #1434, opened 2025-01-06
remove dead-code from ReplacePointerBitcastPass

rjodinchr:
That code was added to manage cases where the source type is smaller than the gep result type. Nowadays that case is handle in various other way making that if statement useless has no test is executing it.
On top of that the new test added is suffering from it, so removing it is actualy fixing that test as well.


AlexDemydenko (approved):
This change fixes <OpenCL-CTS>/test_basic vector_swizzle

PR #1438, opened 2025-01-15
Generate fma from mad when allowed by compile options

rjodinchr:


lib/SPIRVProducerPass.cpp
```diff
@@ -1507,6 +1505,8 @@ SPIRVProducerPassImpl::GetStorageClass(unsigned AddrSpace) const {
     return spv::StorageClassPrivate;
   case AddressSpace::PushConstant:
     return spv::StorageClassPushConstant;
+  default:
+    llvm_unreachable("Unsupported OpenCL address space");
   }
```
rjodinchr, line 1510: This is to prevent an `implicit-fallthrough` error on windows bot

alan-baker (approved):

PR #1439, opened 2025-01-15
Split continues with convergent calls in some cases

alan-baker:
* Adds a new transform to FixupStructuredCFG to split a continue (latch) into two blocks under certain circumstances:
  * The loop header is a conditional branch to the body and latch
  * The latch has two predecessors
  * The latch contains a convergent call

* This transformation prevents clspv forces (along with the breakConditionalHeader transform in the same pass) to prevent convergent operations from being placed in the loop continue contruct. Instead they end up as a structured selection in the body. This ensures reconvergence more robustly than previously. SPIRV-Cross, for example, inlines continues into the body under the assumption that reconvergence is not expected

### DIFF
--- a/lib/FixupStructuredCFGPass.cpp
+++ b/lib/FixupStructuredCFGPass.cpp
@@ -134,7 +134,7 @@ void clspv::FixupStructuredCFGPass::isolateConvergentLatch(
 
     // Break the latch such that it is a single-entry single-exit block.
     // This will force later transforms in this fixup to break the loop header
-    // which puts the whole loop body as a secltion.
+    // which puts the whole loop body as a selection.
     if (latch_terminator->isConditional()) {
       // Safety valve: if this is not an exiting block then the loop is not
       // structured as expected.
--- a/lib/FixupStructuredCFGPass.cpp
+++ b/lib/FixupStructuredCFGPass.cpp
@@ -132,9 +132,9 @@ void clspv::FixupStructuredCFGPass::isolateConvergentLatch(
     if (!latch_terminator)
       continue;
 
-    // Break the latch such that it is single-entry single-exit block.
+    // Break the latch such that it is a single-entry single-exit block.
     // This will force later transforms in this fixup to break the loop header
-    // which puts the whole loop body as secltion.
+    // which puts the whole loop body as a secltion.
     if (latch_terminator->isConditional()) {
       // Safety valve: if this is not an exiting block then the loop is not
       // structured as expected.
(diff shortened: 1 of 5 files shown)


lib/FixupStructuredCFGPass.cpp
```diff
@@ -75,6 +75,97 @@ void clspv::FixupStructuredCFGPass::removeUndefPHI(Function &F) {
...
+    }
+
+    auto *latch_terminator =
+        dyn_cast_or_null<BranchInst>(latch->getTerminator());
+    if (!latch_terminator)
+      continue;
+
+    // Break the latch such that it is single-entry single-exit block.
+    // This will force later transforms in this fixup to break the loop header
+    // which puts the whole loop body as secltion.
```
dneto0, line 137: typo "as a selection".

lib/FixupStructuredCFGPass.cpp
```diff
@@ -75,6 +75,97 @@ void clspv::FixupStructuredCFGPass::removeUndefPHI(Function &F) {
...
+    if (!has_convergent_call) {
+      continue;
+    }
+
+    auto *latch_terminator =
+        dyn_cast_or_null<BranchInst>(latch->getTerminator());
+    if (!latch_terminator)
+      continue;
+
+    // Break the latch such that it is single-entry single-exit block.
```
dneto0, line 135: missing word: it  is *a* single-entry...

test/FixupStructuredCFG/split_convergent_continue_cond_branch.ll
```diff
@@ -0,0 +1,43 @@
+; RUN: clspv-opt --passes=fixup-structured-cfg %s -o %t.ll
+; RUN: FileCheck %s < %t.ll
+
+; CHECK: entry:
+; CHECK: [[new_header:[a-zA-Z0-9_.]+]]:
```
dneto0, line 5: similarly here: recommend matching the branch to new header as well.

dneto0 (approved):
LGTM thanks!

Some suggested typo fixes, and slightly tighter/explanatory matching in the tests.

lib/FixupStructuredCFGPass.cpp
```diff
@@ -132,9 +132,9 @@ void clspv::FixupStructuredCFGPass::isolateConvergentLatch(
     if (!latch_terminator)
       continue;
 
-    // Break the latch such that it is single-entry single-exit block.
+    // Break the latch such that it is a single-entry single-exit block.
     // This will force later transforms in this fixup to break the loop header
-    // which puts the whole loop body as secltion.
+    // which puts the whole loop body as a secltion.
```
dneto0, line 137: "selection"

dneto0 (approved):

PR #1444, opened 2025-01-23
kokoro: amber and dawn builds use gcc 13

dneto0:


alan-baker (approved):

rjodinchr:
I see only 10 successful checks (we have 13 usually). Re-running the bots before merging to make sure everything is passing.

PR #1445, opened 2025-01-23
kokoro: Use clang 13 for format check

dneto0:


rjodinchr (approved):

rjodinchr:
I see only 12 successful checks (we have 13 usually). Re-running the bots before merging to make sure everything is passing.

PR #1450, opened 2025-02-11
Fix/remove dead const globals for PhysicalStorageBuffers support

annestrand:
Move PhysicalPointerArgsPass before:
- AutoPodArgsPass
- DeclarePushConstantsPass
- DefineOpenCLWorkItemBuiltinsPass

Also call NormalizeGlobalVariables at end of
PhysicalPointerArgsPass to remove dead const globals.

Otherwise, subsequent PCs will see 8 byte offset
for a dead/removed module constant storage buffer.

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -246,7 +246,7 @@ clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
           enabled) {
         instance.getPreprocessorOpts().addMacroDef(str);
       }
-      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled) {
+      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled){
         instance.getPreprocessorOpts().addMacroUndef(str);
       }
     }
@@ -704,7 +704,7 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     pm.addPass(clspv::SimplifyPointerBitcastPass());
     pm.addPass(clspv::ReplacePointerBitcastPass());
     pm.addPass(llvm::createModuleToFunctionPassAdaptor(llvm::DCEPass()));
-
+  
     pm.addPass(clspv::UndoTranslateSamplerFoldPass());
 
     if (clspv::Option::ModuleConstantsInStorageBuffer()) {
@@ -1198,6 +1198,7 @@ int CompilePrograms(const std::vector<std::string> &programs,
         ProgramToModule(context, "source", program, output_log, &error));
     if (error != 0)
       return error;
+
   }
   assert(modules.size() > 0 && modules.back() != nullptr);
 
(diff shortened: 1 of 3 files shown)


annestrand:
@rjodinchr - I keep seeing format issues in Kokoro even though I sanitized with clang-format (CI link shows no information). Any help/clues on this one?

annestrand:
Yep, rebase did the trick. Thanks :)

lib/Compiler.cpp
```diff
@@ -246,7 +246,7 @@ clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
           enabled) {
         instance.getPreprocessorOpts().addMacroDef(str);
       }
-      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled){
```
rjodinchr, line 249: Not needed, please remove

lib/Compiler.cpp
```diff
@@ -701,7 +704,7 @@ int RunPassPipeline(llvm::Module &M, llvm::raw_svector_ostream *binaryStream) {
     pm.addPass(clspv::SimplifyPointerBitcastPass());
     pm.addPass(clspv::ReplacePointerBitcastPass());
     pm.addPass(llvm::createModuleToFunctionPassAdaptor(llvm::DCEPass()));
-
```
rjodinchr, line 704: Not needed, please remove

lib/Compiler.cpp
```diff
@@ -1195,7 +1198,6 @@ int CompilePrograms(const std::vector<std::string> &programs,
         ProgramToModule(context, "source", program, output_log, &error));
     if (error != 0)
       return error;
-
```
rjodinchr, line 1198: Not needed, please remove

rjodinchr (commented):
This is a 2 in 1 PR.
I understand the need for normalization at the end of the PhysicalPointerArgPass, and there is a test for it.
Could you explain a bit more why moving the PhysicalPointerArgPass? And I believe it would deserve at least one test, if not more.

lib/Compiler.cpp
annestrand: Sorry, format applied to whole file. Ack.

lib/Compiler.cpp
annestrand: Ack.

annestrand:
@rjodinchr - Thanks for the review.

> Could you explain a bit more why moving the PhysicalPointerArgPass?

Yeah, I found the following unwanted effects (routines within these passes) if I did not move PhysicalPointerArgPass before these below passes:

`DeclarePushConstantsPass` :
```C++
  // This block evaluates to true (which is not what I want in the added physical_dead_const_global.cl kernel)
  if (clspv::ShouldDeclareModuleConstantsPointerPushConstant(M)) {
    PushConstants.push_back(clspv::PushConstant::ModuleConstantsPointer);
  }
```

`AutoPodArgsPass` :
```C++
  // For added physical_dead_const_global.cl kernel, global_pc_size is 8, but this should be 0 for this kernel
  const auto global_pc_type = clspv::GlobalPushConstantsType(M);
  const auto global_pc_size =
      DL.getTypeStoreSize(global_pc_type).getKnownMinValue();
```

rjodinchr (approved):

rjodinchr:
It's always hard to be sure that this kind of change is not introducing regressions/bugs. I've played with it a little and did not see anything wrong. @alan-baker could you have a quick look as well please?

alan-baker (approved):

PR #1452, opened 2025-02-19
kokoro: simplify check-format jobs

dneto0:
- rename presubmit job config file. The fact it is for check-format is already in the directory name. Keep the old file until the Google-internal config file has been renamed.

- presubmit and continuous use the same build script


dneto0:
Google-internal config change is cl/728681584

PR #1454, opened 2025-02-20
kokoro: check-format uses a streamlined formatter image

dneto0:
Avoid downloading all of LLVM and Clang source during check-format.

Instead, take clang-format-diff.py from the image.

Update ./utils/check_code_format.sh so it can use env vars:
 - CLANG_FORMAT to find clang-format
 - CLANG_FORMAT_DIFF to find clang-format-diff.py

Both of these are set in the Docker image.

BUG=397439901

### DIFF
--- a/kokoro/check-format/build-docker.sh
+++ b/kokoro/check-format/build-docker.sh
@@ -13,30 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This script should be run from the project root directory.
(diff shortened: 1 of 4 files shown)


dneto0:
This check-format job took only 56 seconds; the previous one that downloads the llvm source tree to get clang-format-diff.py takes more than 7 minutes.
This is a win and a cleanup.

Landing it.

PR #1465, opened 2025-03-05
Add generation of reflection for workgroup variables size

rjodinchr:


rjodinchr:
This is missing tests

rjodinchr:
Both `CI-clvk-ubuntu-clang-debug` and `CI-amber-ubuntu-gcc-release` are expected to fail as they are using an older version using `SPIRV-Headers`.

lib/SPIRVProducerPass.cpp
```diff
@@ -2948,6 +2951,10 @@ void SPIRVProducerPassImpl::GenerateGlobalVar(GlobalVariable &GV) {
   if (!(module_scope_constant_external_init &&
         clspv::Option::PhysicalStorageBuffers())) {
     var_id = addSPIRVGlobalVariable(ptr_id, spvSC, InitializerID, interface);
+    if (spvSC == spv::StorageClass::StorageClassWorkgroup) {
+      WorkgroupVariableSizeMap.insert(std::make_pair(
+          var_id.get(), BitcastUtils::SizeInBits(DL, GV.getValueType())));
```
alan-baker, line 2956: The reflection description isn't clear, but I'm surprised this is bits and not bytes.

alan-baker: [CL_KERNEL_LOCAL_MEM_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_LOCAL_MEM_SIZE) says it returns bytes. Probably better to match that no?

rjodinchr: I agree, thanks

alan-baker (approved):

PR #1470, opened 2025-03-07
Fix explicit type layout generation

alan-baker:
* Update SPIRV-Headers and SPIRV-Tools
* Change the producer to always discriminate between laid out and non-laid out storage classes instead of being version based
* Introduce function to handle layout conversions
  * OpCopyLogical for 1.4 or later
  * recursive extract and construct for earlier versions

Test modifications take a couple forms:
* simplification (dropping composite extracts)
* increasing spir-v version and using OpCopyLogical

Tested internally against expected CTS passes.

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -14,7 +14,7 @@
       "subrepo" : "KhronosGroup/SPIRV-Headers",
       "branch" : "main",
       "subdir" : "third_party/SPIRV-Headers",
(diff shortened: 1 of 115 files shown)


rjodinchr (approved):

dneto0 (approved):
Nice

dneto0:
the kokoro bots are backed up.  I'll land this and take responsibility if it went wrong.

Issue #1473, opened 2025-03-14
While-loop infinite hang in SimplifyPointerBitcastPass


annestrand:
I found a simple kernel that brings clspv into a never-ending loop case on `SimplifyPointerBitcastPass` in this chunk of code:
```C++
// Loop through our individual simplification passes until they stop changing
  // things.
  bool changed = true;
  while (changed) {
    changed = false;    

    changed |= runOnTrivialBitcast(M);
    changed |= runOnBitcastFromBitcast(M);
    changed |= runOnImplicitGEP(M);
    while (runOnUpgradeableConstantCasts(M)) {
      changed = true;
    }
    changed |= runOnUnneededIndices(M);
    changed |= runOnImplicitCasts(M);
    changed |= runOnAllocaNotAliasing(M);
    changed |= runOnPHIFromGEP(M);
    changed |= runOnGEPFromGEP(M);
  }  
```
Seems GEP from below kernel keeps endlessly changing between `runOnUpgradeableConstantCasts` and `runOnImplicitGEP`.

Here is the kernel that triggers the issue:
```C
// Useless kernel, but triggers issue in clspv
__kernel void test()
{ 
    __local int x[64];
    __local int8* xPtr;
    __local int8 y;    
    
    xPtr = (__local int8 *)x;
    y = xPtr[7];
}
```
LLVM IR BB for this on this pass yields:
```
entry:
  %arrayidx = getelementptr inbounds <8 x i32>, ptr addrspace(3) @test.x, i32 7
  %0 = load <8 x i32>, ptr addrspace(3) %arrayidx, align 32
  store <8 x i32> %0, ptr addrspace(3) @test.y, align 32
  ret void
```
Which looks fine. clspv normalizes GEP here during `runOnImplicitGEP`, which is also fine:
```
entry:
  %0 = getelementptr [64 x i32], ptr addrspace(3) @test.x, i32 0, i32 56
  %1 = load <8 x i32>, ptr addrspace(3) %0, align 32
  store <8 x i32> %1, ptr addrspace(3) @test.y, align 32
  ret void
```
However, on `runOnUpgradeableConstantCasts`, on the Load-Instruction analysis, it transforms the ptr-source GEP (i.e. `%0`) back into a vectorized form:
```
%0 = getelementptr <8 x i32>, ptr addrspace(3) @test.x, i32 7 
```
Then the endless cycle repeats due to `changed == true`, endlessly running the main while-loop in `SimplifyPointerBitcastPass`.

[...]

PR #1475, opened 2025-03-27
Fix SimplifyPointerBitcastPass hang on const GEP GVs

annestrand:
This change marks all-constant-indice-global-GEPs via metadata tag as an indicator to avoid further transforms on subsequent sub-passes.

This fixes issue where SimplifyPointerBitcastPass endlessly loops on converting and reconverting these particular GEPs.

fixes: #1473

### DIFF
--- a/lib/SimplifyPointerBitcastPass.cpp
+++ b/lib/SimplifyPointerBitcastPass.cpp
@@ -432,8 +432,7 @@ bool clspv::SimplifyPointerBitcastPass::runOnImplicitGEP(Module &M) const {
         } else if (auto gep = dyn_cast<GetElementPtrInst>(&I)) {
           if (IsClspvResourceOrLocal(gep->getPointerOperand())) {
             GEPCastList.push_back(dyn_cast<GetElementPtrInst>(&I));
-          } else if (isa<GlobalVariable>(gep->getPointerOperand()) &&
-                     gep->hasAllConstantIndices()) {
+          } else if (IsGVConstantGEP(gep)) {
             GEPGVList.push_back(dyn_cast<GetElementPtrInst>(&I));
           }
         }
@@ -787,8 +786,7 @@ bool clspv::SimplifyPointerBitcastPass::runOnUpgradeableConstantCasts(
               IsClspvResourceOrLocal(gep->getPointerOperand())) {
             continue;
           }
-          if (isa<GlobalVariable>(gep->getPointerOperand()) &&
-              gep->hasAllConstantIndices())
+          if (IsGVConstantGEP(gep))
             continue;
 
           uint64_t cstVal;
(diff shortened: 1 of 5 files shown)


annestrand:
@rjodinchr - Local testing on my end looks clean for this (compared to my other patch in Issue-thread)
```
...
PASS: clspv :: LongVectorLowering/clspv_fract.ll (3728 of 3729)
PASS: clspv :: VectorLoadStore/vstore2_constant_float2.ll (3729 of 3729)

Testing Time: 11.63s

Total Discovered Tests: 3729
  Passed           : 3728 (99.97%)
  Expectedly Failed:    1 (0.03%)
```

lib/SimplifyPointerBitcastPass.cpp
```diff
@@ -572,6 +572,13 @@ bool clspv::SimplifyPointerBitcastPass::runOnImplicitGEP(Module &M) const {
     LLVM_DEBUG(dbgs() << "\n##runOnImplicitGEP (from GV):\nreplacing: ";
                gep->dump());
     LLVM_DEBUG(dbgs() << "by: "; new_gep->dump(););
+
+    // Tag all-constant-indice-global-GEPs to stay in normalized form to
+    // prevent conversion-thrashing with latter sub-passes
+    // TODO: this could be better handled without the use of metadata tagging
```
rjodinchr, line 578: Looking at it quickly, it feels like the condition would be something like: `isa<GlobalVariable>(gep->getPointerOperand()) && gep->hasAllConstantIndices()`.
Have you tried it instead of using `gep->getMetadata("clspv.const_gep_gv")`?

annestrand: Thanks @rjodinchr - yep that works as well. Went ahead and updated this PR 👍

rjodinchr: That condition is now repeated 3 times in that file. Maybe we could have a function for it? `isConstantGEP` or something similar?

rjodinchr (changes requested):
Please add at least one test for it.

test/PointerCasts/issue-1473.cl
```diff
@@ -0,0 +1,25 @@
+// RUN: clspv --long-vector %s -o %t.spv
+// RUN: spirv-dis %t.spv -o %t.spvasm
+// RUN: spirv-val --target-env vulkan1.0 %t.spv
+
```
rjodinchr, line 4:
```suggestion

// This test used to involve an infinite loop

```

rjodinchr (commented):
The `.cl` test can easily stop checking the issue without letting us know.
Please add the following which is more focus on the issue fixed by that pull request:
```
; RUN: clspv-opt %s -o %t.ll --passes=simplify-pointer-bitcast,replace-pointer-bitcast
; RUN: FileCheck %s < %t.ll

; This test used to involve an infinite loop

; CHECK:  [[gep:%[^ ]+]] = getelementptr [8 x i32], ptr addrspace(3) @test.x, i32 0
; CHECK-COUNT-8: getelementptr i32, ptr addrspace(3) [[gep]], i32

target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
target triple = "spir-unknown-unknown"

@test.x = internal addrspace(3) global [8 x i32] undef, align 4

define dso_local spir_kernel void @test(ptr addrspace(1) align 32 %output) {
entry:
  %arrayidx1 = getelementptr inbounds <8 x i32>, ptr addrspace(3) @test.x, i32 0
  %0 = load <8 x i32>, ptr addrspace(3) %arrayidx1, align 32
  ret void
}
```

rjodinchr (approved):

PR #1498, opened 2025-06-18
Add support for sub_group_shuffle and sub_group_shuffle_xor

xueliang-zhong-arm:
- Map `sub_group_shuffle()` to `OpGroupNonUniformShuffle`.
- Map `sub_group_shuffle_xor()` to `OpGroupNonUniformShuffleXor`.
- Added sub_group_shuffle support in builtin enumeration, mappings, and macro definitions.
- Added tests for sub_group_shuffle and sub_group_shuffle_xor.

Change-Id: Ice29eadca2bdf4c16c0e236324fecee4bfb9d8da


rjodinchr (changes requested):
Please keep things as sorted as possible (alphabetical wise).

I understand that `BuiltinsMap.inc` is already not sorted. Do not sort already existing lines, but add yours in a coherent place please.

lib/BuiltinsEnum.h
```diff
@@ -343,6 +343,8 @@ enum BuiltinType : unsigned int {
   kSubGroupBarrier,
   kSubGroupAll,
   kSubGroupAny,
+  kSubGroupShuffle,
```
rjodinchr, line 346: Not sorted

lib/SPIRVProducerPass.cpp
```diff
@@ -4164,6 +4164,14 @@ SPIRVProducerPassImpl::GenerateSubgroupInstruction(
     addCapability(spv::CapabilityGroupNonUniformVote);
     op = spv::OpGroupNonUniformAny;
     break;
+  case Builtins::kSubGroupShuffle:
```
rjodinchr, line 4167: Not sorted

rjodinchr (changes requested):

lib/SPIRVProducerPass.cpp
xueliang-zhong-arm: Sorted in the new commit.

rjodinchr (approved):

PR #1505, opened 2025-08-03
Fix trial of building failure on systems where function rsqrt is declared as cmath builtin (reopened)

Dalmurii:
On some machine, compilation tends to fail due to the pre-declared rsqrt on cmath.  
Namespace will be named `clspv_local` to avoid the declaration crash.  
Plus, Method will be static to avoid global declaration to avoid.

The change will look something like this.

```cc
/** Naming annonymous */
namespace clspv_local {
static double rsqrt(double input) { return 1.0 / sqrt(input); }
} // namespace

bool FixupBuiltinsPass::runOnFunction(Function &F) {
  auto &FI = Builtins::Lookup(&F);
  switch (FI.getType()) {
  case Builtins::kSqrt:
    return fixupSqrt(F, sqrt);
  case Builtins::kRsqrt:
    return fixupSqrt(F, clspv_local::rsqrt); /** Using rsqrt with namespace since rsqrt is not part of the regular cmath as I know. */
  case Builtins::kReadImagef:
  case Builtins::kReadImagei:
  case Builtins::kReadImageui:
    if (clspv::Option::HackImage1dBufferBGRA() &&
        !FI.getParameter(1).isSampler()) {
      return fixupReadImage(F);
    } else {
      return false;
    }
  default:
    return false;
  }
}
```

This pr has been reopened since the username has been changed.
https://github.com/google/clspv/pull/1504

### DIFF
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -11,22 +11,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#include "clspv/Compiler.h"
-
-#include <cassert>
-#include <fstream>
-#include <iostream>
-#include <numeric>
-#include <ostream>
-#include <sstream>
-#include <string>
-
-#include "Builtins.h"
-#include "BuiltinsEnum.h"
-#include "Constants.h"
-#include "FrontendPlugin.h"
-#include "Passes.h"
-#include "Types.h"
 #include "clang/Basic/FileManager.h"
 #include "clang/Basic/TargetInfo.h"
 #include "clang/CodeGen/CodeGenAction.h"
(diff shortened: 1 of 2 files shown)


rjodinchr:
> > Thank you for your contribution. Please format your patch with `clang-format`, then I think it should be ready to go in.

>  Thanks for reminding of this. I have run `clang-format -style=google -i lib/Compiler.cpp` and I think things are working fine.

You don't want to format the whole file. It is modifying lines you did not change.
Also we have a `.clang-format` at the root of the project, which means that you should not specify the style (FYI, `clspv` is using the LLVM style).

Dalmurii:
Oh which means I should run clang-format on root?

Dalmurii:
Would I specify CLANG_FORMAT as llvm running it with plain sh seems to want me to do so.

Dalmurii:
I would need to ask, will this be compiled whilst cmake building? 
I see neither clang-format file no utils/clang-format-diff.py the shell is mentioning.

Dalmurii:
> Please undo your changes to `Compiler.cpp`.
I have edited the `lib/Compiler.cpp` to original source on main branch.

Dalmurii:
I have removed the visible difference except the last line (line 1403).

Issue #1506, opened 2025-08-03
I see no C interface header yet and I am willing to write one.


Dalmurii:
The headers for core library of clspv doesn't seem to be compatible for C interface.

I would like either to make new headers which is compatible for C interface, or edit existing headers to make it compatible both for C and C++ with preprocessing.

When it comes to making new headers for C interface, I would write new headers in new directory something like this:
> include/clspv/c-interface

When it comes to editing existing headers to make it compatible both for C/C++ changes of directory might look like:
> include/clspv

Dalmurii:
Yes, I am compiling with C compiler. That was a personal issue of mine related to this.
I have resolved that issue by adding the preprocessor and making the header `Compiler.h` compatible both for C/C++ you have mentioned.

rjodinchr:
My understanding is that we have added a C interface in `Compiler.h`. If we cannot include `Compiler.h` when compiling with a C compiler, it's not an issue on your side only.

I just want to make sure of what's happening on @annestrand side.

In the meantime, please go ahead with your PR. We should be to unblock you.

Dalmurii:
Opening the pull request related to this. I am closing this issue.

PR #1508, opened 2025-08-05
C interface header

Dalmurii:
Fixing trial of this issue: https://github.com/google/clspv/issues/1506

- Made `include/clspv/Compiler.h` compatible with least C99.
- Added qualifier `const` on input of clspvCompileFromSourcesString

 Changes to be committed:
	modified:   include/clspv/Compiler.h
	modified:   lib/Compiler.cpp

### DIFF
--- a/include/clspv/Compiler.h
+++ b/include/clspv/Compiler.h
@@ -15,11 +15,17 @@
 #ifndef CLSPV_INCLUDE_CLSPV_COMPILER_H_
 #define CLSPV_INCLUDE_CLSPV_COMPILER_H_
 
+#ifdef __cplusplus
 #include <cstdint>
 #include <cstdlib>
 #include <string>
 #include <vector>
+#else
+#include <stdint.h>
+#include <stdlib.h>
+#endif
 
+#ifdef __cplusplus
 namespace clspv {
 // DEPRECATED: This function will be replaced by an expanded API.
 int Compile(const int argc, const char *const argv[]);
@@ -44,6 +50,7 @@ int CompileFromSourcesString(const std::vector<std::string> &programs,
                              std::vector<uint32_t> *output_buffer,
                              std::string *output_log);
 } // namespace clspv
+#endif
 
 // C API
 typedef enum ClspvError {
@@ -57,6 +64,14 @@ typedef enum ClspvError {
 #define EXPORT __declspec(dllexport)
 #else
 #define EXPORT __attribute__((visibility("default")))
+#endif
+
+#ifndef clspv_restrict
+
+/** Optional restrict qualifier for given C API with C++ implementation where
+ * restrict is not the thing. */
+#define clspv_restrict
+
 #endif
 
 #ifdef __cplusplus
@@ -74,22 +89,27 @@ extern "C" {
 // |options|            - String of options to pass to Clspv compiler.
 // |output_binary|      - Handle to compiler output/result.
 // |output_binary_size| - Size of compiler output/result (in bytes).
-// |output_log|         - Handle to compiler build log.
+// |output_log|         - Handle to compiler build log,
+// 			  will assume log is not needed if param is NULL.
+// 			  its value will be NULL when no build log is empty.
 EXPORT ClspvError clspvCompileFromSourcesString(
-    const size_t program_count, const size_t *program_sizes,
-    const char **programs, const char *options, char **output_binary,
-    size_t *output_binary_size, char **output_log);
+    const size_t program_count, const size_t *clspv_restrict program_sizes,
+    const char *const clspv_restrict *clspv_restrict programs,
+    const char *clspv_restrict options,
+    char *clspv_restrict *clspv_restrict output_binary,
+    size_t *clspv_restrict output_binary_size,
+    char *clspv_restrict *clspv_restrict output_log_opt);
 
 // Frees the output memory from clspvCompileFromSourcesString
 //
 // |output_binary|      - Handle to spv
 // |output_log|         - Handle to compiler build log
-static inline void clspvFreeOutputBuildObjs(char *output_binary,
-                                            char *output_log) {
-  free(output_binary);
-  output_binary = NULL;
-  free(output_log);
-  output_log = NULL;
+static inline void clspvFreeOutputBuildObjs(char *clspv_restrict output_binary,
+                                            char *clspv_restrict output_log) {
+  if (output_binary)
+    free(output_binary);
+  if (output_log)
+    free(output_log);
 }
 
 #ifdef __cplusplus
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -245,7 +245,7 @@ clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
           enabled) {
         instance.getPreprocessorOpts().addMacroDef(str);
       }
-      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled){
+      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled) {
         instance.getPreprocessorOpts().addMacroUndef(str);
       }
     }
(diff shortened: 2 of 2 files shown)


lib/Compiler.cpp
```diff
@@ -245,7 +245,7 @@ clang::TargetInfo *PrepareTargetInfo(CompilerInstance &instance) {
           enabled) {
         instance.getPreprocessorOpts().addMacroDef(str);
       }
-      if (feat == clspv::FeatureMacro::__opencl_c_int64 && !enabled){
```
rjodinchr, line 248: Should not be modified, please remove

lib/Compiler.cpp
```diff
@@ -1206,7 +1206,6 @@ int CompilePrograms(const std::vector<std::string> &programs,
         ProgramToModule(context, "source", program, output_log, &error));
     if (error != 0)
       return error;
-
```
rjodinchr, line 1209: Should not be modified, please remove

include/clspv/Compiler.h
```diff
@@ -74,22 +90,27 @@ extern "C" {
...
-  free(output_binary);
-  output_binary = NULL;
-  free(output_log);
-  output_log = NULL;
+static inline void clspvFreeOutputBuildObjs(char *clspv_restrict output_binary,
+                                            char *clspv_restrict output_log) {
+  if (output_binary)
+    free(output_binary);
+  if (output_log)
+    free(output_log);
```
rjodinchr, line 113: `free(NULL)` is well defined, no reason to avoid it with a check.
> The free() function frees the memory space pointed to by ptr, which must have been returned by a previous call to malloc(), calloc() or realloc(). Otherwise, or if free(ptr) has already been called before, undefined behavior occurs. If ptr is NULL, no operation is performed.
https://linux.die.net/man/3/free

I think  we should not change that section

include/clspv/Compiler.h
```diff
@@ -74,22 +90,27 @@ extern "C" {
 // |options|            - String of options to pass to Clspv compiler.
 // |output_binary|      - Handle to compiler output/result.
 // |output_binary_size| - Size of compiler output/result (in bytes).
-// |output_log|         - Handle to compiler build log.
+// |output_log|         - Handle to compiler build log,
+// 			  will assume log is not needed if param is NULL.
+// 			  its value will be NULL when no build log is empty.
```
rjodinchr, line 95: It's not clear to me what the precision you are trying to add here.
Maybe something like that?
```suggestion
// |output_log|         - Handle to compiler build log. `output_log` can be NULL
//                        in which case no log will be generated.
```

include/clspv/Compiler.h
```diff
@@ -59,6 +67,14 @@ typedef enum ClspvError {
...
 #endif
 
+#ifndef clspv_restrict
+
+/** Optional restrict qualifier for given C API with C++ implementation where
+ * restrict is not the thing. */
+#define clspv_restrict
+
+#endif
+
```
rjodinchr, line 77: I don't like that optional qualifier on a public interface. Either we want it or not, but I don't think making it optional is a good idea.
As that interface already exists it better not to change it anymore.
Please remove

lib/Compiler.cpp
```diff
@@ -1352,7 +1351,7 @@ int CompileFromSourceString(const std::string &program,
 // C API
 ClspvError clspvCompileFromSourcesString(
     const size_t program_count, const size_t *program_sizes,
-    const char **programs, const char *options, char **output_binary,
+    const char *const *programs, const char *options, char **output_binary,
```
rjodinchr, line 1354: It is not good practice to change a public interface, it might break projects already relying on it.
Pleas remove

Dalmurii: How I understand about this interface is that, argument `programs` of `clspvCompileFromSourcesString` is an input value and the original value will not change via this operation. 

Adding the qualifier `const` is necessary to guarantee this since with `const char**` as a list of pointer, the elements as pointers could be mutable, making callers feel not safety.

If you do not like changing the interface since callers of `clspvCompileFromSourcesString`, I think you do not need to worry about them since the change of this would cover the original usage. (castring `const char**` to `const char* const *` is a safe behaviour, other way is dangerous.)

I would ask do you dislike this change since declaration crash from ones who copies the API?

include/clspv/Compiler.h
Dalmurii: This seems much better and make sense. Many thanks for pointing me this.

include/clspv/Compiler.h
Dalmurii: This is my absolute abomination. I would revert this.

lib/Compiler.cpp
Dalmurii: I would open an issue about this and revert this change.
This might be too far from the original purpose of this pull request.

lib/Compiler.cpp
```diff
@@ -1394,6 +1398,12 @@ ClspvError clspvCompileFromSourcesString(
 
   // Alloc and copy backing mem for spv output
   size_t spv_bytes = binary.size() * sizeof(uint32_t);
+  if (!spv_bytes) {
+    /** Early return: when allocation is not needed. */
+    *spv_bytes = static_cast<void *>(NULL);
+    return CLSPV_SUCCESS;
+  }
```
rjodinchr, line 1405:
```suggestion
  if (spv_bytes == 0) {
    /** Early return: when allocation is not needed. */
    *output_binary = static_cast<char *>(NULL);
    return CLSPV_SUCCESS;
  }
```
`spv_bytes` is a `size_t` not a boolean, let's keep thing as readable as possible by avoiding boolean operators on `size_t` (even so I understand that it produced the same outcome here).

`*spv_bytes` feels like a bug as `spv_bytes` is a `size_t` not a pointer.

lib/Compiler.cpp
```diff
@@ -1378,14 +1378,18 @@ ClspvError clspvCompileFromSourcesString(
   err =
       clspv::CompileFromSourcesString(vPrograms, sOptions, &binary, &buildLog);
 
-  if (!buildLog.empty()) {
-    // Alloc and copy backing mem for build log
-    *output_log = static_cast<char *>(std::malloc(buildLog.size() + 1));
-    if (*output_log == NULL) {
-      return CLSPV_OUT_OF_HOST_MEM;
+  if (output_log) {
```
rjodinchr, line 1381:
```suggestion
  if (output_log != NULL) {
```
`output_log` is a pointer, not a boolean, let's keep thing as readable as possible by avoiding boolean operators on pointers

lib/Compiler.cpp
Dalmurii: `spv_bytes` is a bug. changed to `output_binary`.

lib/Compiler.cpp
```diff
@@ -1378,14 +1378,18 @@ ClspvError clspvCompileFromSourcesString(
...
+    if (!buildLog.empty()) {
+      // Alloc and copy backing mem for build log
+      *output_log = static_cast<char *>(std::malloc(buildLog.size() + 1));
+      if (*output_log == NULL) {
+        return CLSPV_OUT_OF_HOST_MEM;
+      }
+      std::memcpy(static_cast<void *>(*output_log), buildLog.c_str(),
+                  buildLog.size() + 1);
+    } else {
+      *(output_log) = static_cast<void *>(NULL);
```
rjodinchr, line 1391: types do not match. `char *` on one side `void *` on the other

lib/Compiler.cpp
```diff
@@ -1394,6 +1398,12 @@ ClspvError clspvCompileFromSourcesString(
 
   // Alloc and copy backing mem for spv output
   size_t spv_bytes = binary.size() * sizeof(uint32_t);
+  if (spv_bytes == 0) {
+    /** Early return: when allocation is not needed. */
+    *output_binary = static_cast<void *>(NULL);
```
rjodinchr, line 1403: types do not match. `char *` on one side `void *` on the other

lib/Compiler.cpp
Dalmurii: Changed to `output_log != NULL`

rjodinchr (changes requested):

lib/Compiler.cpp
Dalmurii: Both expected `char*` but `void*` problem has been resolved.

lib/Compiler.cpp
rjodinchr, line 1209: No reason to remove that line, that PR is not touching anything around it.

Dalmurii: Thought that was from my modification. Sorry.

rjodinchr (approved):

Dalmurii:
CI checks seem to success.

rjodinchr:
@alan-baker PTAL

alan-baker (approved):

PR #1518, opened 2025-08-22
Update deps

alan-baker:
* Fix SPIR-V validation errors related to memory semantics


rjodinchr (approved):

rjodinchr:
Do you really need to update LLVM here?
I think you are hitting https://github.com/llvm/llvm-project/pull/141574#discussion_r2260773717

Issue #1519, opened 2025-08-28
Build libclc with LLVM_ENABLE_RUNTIMES instead of LLVM_ENABLE_PROJECTS


rjodinchr:
libclc has to be built with `LLVM_ENABLE_RUNTIMES`. It cannot be built with `LLVM_ENABLE_PROJECTS` with the latest LLVM.

rjodinchr:
We need fix in LLVM to be able to enable the build of libclc:
- https://github.com/llvm/llvm-project/pull/155836
- https://github.com/llvm/llvm-project/pull/155844

With those and https://github.com/rjodinchr/clspv/commit/a3721b0309e90314278aea90175b5196a4cc790a, I am able to build clspv.

rjodinchr:
cc @frasercrmck

PR #1526, opened 2025-09-29
patches: clean patch to get something we can merge upstream

rjodinchr:
Ref #1519


rjodinchr:
let's wait to see if we can land https://github.com/llvm/llvm-project/pull/161220 to merge that PR.

PR #1527, opened 2025-09-30
kokoro: windows: use Ninja instead of Visual Studio

rjodinchr:
Visual Studio is not capable of building libclc as a runtime. Let's switch to Ninja.

Ref #1519

### DIFF
--- a/kokoro/scripts/windows/build.bat
+++ b/kokoro/scripts/windows/build.bat
@@ -28,14 +28,6 @@ set PATH=c:\Python312;%PATH%
 cd %SRC%
 python utils/fetch_sources.py --ci
 
-:: #########################################
-:: set up msvc build env
-:: #########################################
-if %VS_VERSION% == 2022 (
-  set GENERATOR="Visual Studio 17 2022"
-  echo "Using VS 2022..."
-)
-
 cd %SRC%
 mkdir build
 cd build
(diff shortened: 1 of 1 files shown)


kokoro/scripts/windows/build.bat
```diff
@@ -28,14 +28,6 @@ set PATH=c:\Python312;%PATH%
 cd %SRC%
 python utils/fetch_sources.py --ci
 
-:: #########################################
-:: set up msvc build env
-:: #########################################
-if %VS_VERSION% == 2022 (
-  set GENERATOR="Visual Studio 17 2022"
-  echo "Using VS 2022..."
-)
```
dneto0, line 37: This doesn't work.

Errors out like this:
```
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:36 (project):
  The CMAKE_C_COMPILER:

    cl.exe

  is not a full path and was not found in the PATH.  Perhaps the extension is
  missing?

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
```

I think this needs to invoke the environment setup path. e.g. see SPIRV-Tools script here:

https://github.com/KhronosGroup/SPIRV-Tools/blob/main/kokoro/scripts/windows/build.bat#L30


```
if %VS_VERSION% == 2022 (
  call "C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Auxiliary\Build\vcvarsall.bat" x64
  echo "Using VS 2022..."
)
```

dneto0 (changes requested):
I'm fine with changing it, but the path setup script needs to be invoked.

rjodinchr:
closed as we fixed windows with: https://github.com/google/clspv/pull/1528

PR #1531, opened 2025-10-08
Run global var normalization early

### LINKED ISSUE #1524: LLVM update causes failures in work item tests
Cross-reference crbug.com/445660270

AI summary "The bug report details failures in OpenCL CTS tests due to incorrect values returned by `get_work_dim()` within the work_item_functions test. The root cause is an update to LLVM affecting address calculations, specifically in the `SimplifyPointerBitcastPass` and interactions with `instcombine`. The transformation of GEPs is leading to incorrect indexing in the generated SPIR-V code. This results in array indexing instead of correctly selecting struct elements."

Causes failures of work item CTS tests.

Also causes failures in Clspv lit-based tests

alan-baker:
Fixes #1524

* Run global variable normalization as a pass early on before it can get too confused by other optimizations


lib/PassRegistry.def
```diff
@@ -43,6 +43,7 @@ MODULE_PASS("lower-addrspacecast", clspv::LowerAddrSpaceCastPass)
 MODULE_PASS("lower-private-pointer-phi", clspv::LowerPrivatePointerPHIPass)
 MODULE_PASS("multi-version-ubo-functions", clspv::MultiVersionUBOFunctionsPass)
 MODULE_PASS("native-math", clspv::NativeMathPass)
+MODULE_PASS("noralize-global-variables", clspv::NormalizeGlobalVariablesPass)
```
rjodinchr, line 46:
```suggestion
MODULE_PASS("normalize-global-variables", clspv::NormalizeGlobalVariablesPass)
```

dneto0 (commented):
LGTM

PR #1540, opened 2025-10-31
Option.cpp: add missing options

rjodinchr:


rjodinchr:
https://github.com/KhronosGroup/OpenCL-CTS/issues/2563

PR #1545, opened 2025-11-17
LogicalPtrToInt: remove ptrtoint(addrspacecast(null))

rjodinchr:
clang is now emitting such pattern that are not handle by LogicalPtrToInt.
They are handled by AddrSpaceCastPass, but that pass might not be run.


dneto0 (approved):
Thanks!

PR #1555, opened 2026-01-09
Update LLVM

rjodinchr:
Since https://github.com/llvm/llvm-project/commit/f07988ff3ec8b8b838f5c6ce9e1ec519e16432a9 which removed a tool that used to set the linkage to linkonce_odr on the clc builtins. Thus we now need to make sure to link only what we needed to avoid bring the whole libclc (leading to a huge increase in build time, and various other bugs).


dneto0 (approved):
Thanks!

PR #1587, opened 2026-04-13
kokoro: Use Debug for windows-vs2022-amd64-debug

### LINKED ISSUE #1437: kokoro windows "debug" builds actually build RelWithDebInfo intentional?
The kokoro build config for windows debug builds use RelWithDebInfo.
Is that intentional?  Shouldn't it be Debug? Or does that take too long?

rjodinchr:
Fix #1437

### DIFF
--- a/kokoro/windows-vs2022-amd64-debug/build.bat
+++ b/kokoro/windows-vs2022-amd64-debug/build.bat
@@ -18,4 +18,4 @@
 set SCRIPT_DIR=%~dp0
 
 :: Call with correct parameter
-call %SCRIPT_DIR%\..\scripts\windows\build.bat RelWithDebInfo 2022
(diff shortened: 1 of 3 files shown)


kokoro/windows-vs2022-amd64-debug/build.bat
```diff
@@ -18,4 +18,4 @@
 set SCRIPT_DIR=%~dp0
 
 :: Call with correct parameter
-call %SCRIPT_DIR%\..\scripts\windows\build.bat RelWithDebInfo 2022
```
dneto0, line 21: oh no.  I'm sorry.  :-)

dneto0 (approved):
Huh, it wasn't using the version parameter.

I was worried about what happens when we upgrade to a new version. But at that point we should fix the scripts.

PR #1588, opened 2026-04-14
Add support for SPV_KHR_fma

### LINKED ISSUE #1533: Support SPV_KHR_fma
[SPV_KHR_fma](https://github.com/KhronosGroup/SPIRV-Registry/pull/361) adds a new `OpFmaKHR` instruction for correctly rounded fma. There is a corresponding Vulkan extension that allows implementations to advertise a correctly rounded fma. This should be preferred over alternative implementations.

rjodinchr:
Fix #1533

### DIFF
--- a/lib/Option.cpp
+++ b/lib/Option.cpp
@@ -453,6 +453,17 @@ static llvm::cl::opt<bool>
     untyped_pointers("untyped-pointers", llvm::cl::init(false),
                      llvm::cl::desc("Enable SPV_KHR_untyped_pointers"));
 
+static llvm::cl::list<clspv::Option::SpvKhrFma> spv_khr_fma(
+    "spv-khr-fma",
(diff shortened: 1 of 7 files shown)


lib/Option.cpp
```diff
@@ -453,6 +453,9 @@ static llvm::cl::opt<bool>
     untyped_pointers("untyped-pointers", llvm::cl::init(false),
                      llvm::cl::desc("Enable SPV_KHR_untyped_pointers"));
 
+static llvm::cl::opt<bool> spv_khr_fma("spv-khr-fma", llvm::cl::init(false),
```
alan-baker: The vulkan feature has separate bits for 16-, 32-, and 64-bit variants. Should this option reflect that?

rjodinchr: Done

alan-baker (approved):

PR #1589, opened 2026-04-15
Implement missing Subgroup Builtins for CTS

rjodinchr:
Those functions were missing to pass CTS, especially:
- test_subgroup functions_ballot
- test_subgroup functions_non_uniform_vote

Those functions started to fail since we added incomplete support in clspv.

Mapped bounded properties (`get_sub_group_size()`) securely over modulo fragments via scalar extraction matching Vulkan expectations.


lib/SPIRVProducerPass.cpp
```diff
@@ -4300,8 +4300,57 @@ SPIRVProducerPassImpl::GenerateSubgroupInstruction(
 
   spv::Op op = spv::OpNop;
   switch (FuncInfo.getType()) {
-  case Builtins::kGetSubGroupSize:
-    return loadBuiltin(spv::BuiltInSubgroupSize);
+  case Builtins::kGetSubGroupSize: {
```
alan-baker, line 4303: An alternative:
1. Create a private variable
2. At the start of the entry point store `ballotBitCount(ballot(true))`
3. Load the var here.

alan-baker (approved):
Optional alternative implementation.

lib/SPIRVProducerPass.cpp
rjodinchr: I guess that would add a requirement on `GroupNonUniformBallot` for that builtin.

Also the current approach relies on specialization constants so the Vulkan driver should be able to constant-fold the calculation I believe.

I think I prefer to keep the current implementation

Issue #1607, opened 2026-05-27
Supported llvm IR math instrinsics


EwanC:
When using the IR input to clspv I'm seeing inconsistencies around what math intrinsics are accepted. For example `@llvm.exp.f32` and  `@llvm.log.f32` intrinsic calls are accepted but  `@llvm.exp2.f32`, `@llvm.asin.f32`, and `@llvm.tan.f32` calls result in

```
Unsupported llvm intrinsic
UNREACHABLE executed at $HOME/clspv/lib/SPIRVProducerPass.cpp:2575!
```

rjodinchr:
Expect `clspv` to handle any LLVM IR it originally generated from OpenCL-C source, but anything beyond that is unpredictable.

I’m not suggesting it shouldn't be an objective, only that you’ll encounter plenty of unsupported patterns.

I plan to address some of these as time permits, but you’ll be on your own for others. If you can help out and submit PRs for what you discover, it would be greatly appreciated.

Regarding this specific issue, I suspect the accepted intrinsics are being optimized out by LLVM earlier in the process. Essentially, any math intrinsics left for `clspv` to handle will likely trigger that error. It might be possible to add support within `SPIRVProducerPassImpl::GenerateInstructionFromCall`, though I fear that list will need to be explicitly defined.

EwanC:
Thanks for the clarification, I'll try to find some time to put together some PRs. I've also run into a similar issue with `llvm.memmove` which I can generate in IR but unable to create an OpenCL-C reproducer for.

Closing this issue since it's was more of a question about support status than a bug report than needs fixed.

PR #1609, opened 2026-06-03
Add tests for IR input math intrinsics

EwanC:
There are currently no IR input tests for stressing LLVM math intrinsics are consumed without an error, and then mapped to GLSL extended instructions in SPIR-V output. These llvm intrinsics

This patch adds those tests as well as includes some missing mappings.

Related to https://github.com/google/clspv/issues/1607


rjodinchr (approved):

rjodinchr:
Thank you for the contribution 🙏

PR #1613, opened 2026-06-19
Add LLVM memmove intrinsic support

EwanC:
The [llvm.memmove](https://llvm.org/docs/LangRef.html#llvm-memmove-intrinsic) intrinsic can exist in input LLVM IR generated from the following OpenCL-C code, but is not currently handled by clspv.

```c
$ cat /tmp/memmove.cl
__kernel void foo(__global int* a, int size) {
  for (int i=0; i < size-1; i++) {
    a[i] = a[i+1];
  }
}

$ clang-19 -S -emit-llvm /tmp/memmove.cl -o /tmp/memmove.ll && cat /tmp/memmove.ll
; ModuleID = '/tmp/memmove.cl'
source_filename = "/tmp/memmove.cl"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

; Function Attrs: mustprogress nofree norecurse nosync nounwind willreturn memory(argmem: readwrite) uwtable
define dso_local spir_kernel void @foo(ptr nocapture noundef align 4 %0, i32 noundef %1) local_unnamed_addr #0 !kernel_arg_addr_space !7 !kernel_arg_access_qual !8 !kernel_arg_type !9 !kernel_arg_base_type !9 !kernel_arg_type_qual !10 {
  %3 = icmp sgt i32 %1, 1
  br i1 %3, label %4, label %9

4:                                                ; preds = %2
  %5 = add nsw i32 %1, -1
  %6 = getelementptr i8, ptr %0, i64 4
  %7 = zext nneg i32 %5 to i64
  %8 = shl nuw nsw i64 %7, 2
  tail call void @llvm.memmove.p0.p0.i64(ptr align 4 %0, ptr align 4 %6, i64 %8, i1 false), !tbaa !11
  br label %9

9:                                                ; preds = %4, %2
  ret void
}

; Function Attrs: nocallback nofree nounwind willreturn memory(argmem: readwrite)
declare void @llvm.memmove.p0.p0.i64(ptr nocapture writeonly, ptr nocapture readonly, i64, i1 immarg) #1
```

This patch expands the memmove intrinsic in `ReplaceLLVMIntrinsicPass` by copying to an intermediate alloca byte array when the number of bytes is known at compile time. When the number of bytes to copy isn't known at compile time, then we fallback to LLVMs utility for creating loops to do the copy. The same strategy as [SPIRVLowerMemmove.cpp](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVLowerMemmove.cpp) in the SPIRV-LLVM-Translator.

### DIFF
--- a/lib/ReplaceLLVMIntrinsicsPass.cpp
+++ b/lib/ReplaceLLVMIntrinsicsPass.cpp
@@ -553,37 +553,109 @@ static bool replaceMemmoveDynamic(Module &M, MemMoveInst *MMI) {
     return false;
   }
 
-  if (clspv::Option::PhysicalStorageBuffers()) {
-    // expandMemMoveAsLoop creates a 'icmp ptr' instruction that is converted
-    // into an int comparison to avoid lowering the icmp to OpPtrDiff which
-    // isn't supported for PhysicalStorageBuffer
-    ICmpInst *ICmpToProcess = nullptr;
-
-    for (auto &I : *BB) {
-      if (auto ICmp = dyn_cast<ICmpInst>(&I)) {
-        if (ICmp->getOperand(0) == MMI->getSource() &&
-            ICmp->getOperand(1) == MMI->getDest()) {
-          ICmpToProcess = ICmp;
-          break;
-        }
+  // expandMemMoveAsLoop creates a 'icmp ptr' instruction that we try to
+  // transform to avoid lowering to OpPtrDiff which isn't supported for
+  // PhysicalStorageBuffer or Function storage classes.
+  ICmpInst *ICmpToProcess = nullptr;
+  for (auto &I : *BB) {
+    if (auto ICmp = dyn_cast<ICmpInst>(&I)) {
+      if (ICmp->getOperand(0) == MMI->getSource() &&
+          ICmp->getOperand(1) == MMI->getDest()) {
+        ICmpToProcess = ICmp;
+        break;
       }
     }
+  }
 
-    if (ICmpToProcess) {
+  bool TransformedICMP = false;
+  if (ICmpToProcess) {
+    // Try to constant fold the icmp if the operands point to the same alloca
+    // in the private address space at compile time known offsets
+    Value *LHS = ICmpToProcess->getOperand(0);
+    Value *RHS = ICmpToProcess->getOperand(1);
+    auto Pred = ICmpToProcess->getPredicate();
+
+    // APInts to accumulate base offsets (relative to the alloca pointer in
+    // bits/bytes)
+    APInt LHSOffset(64, 0);
+    APInt RHSOffset(64, 0);
+
+    // Strip away GEPs/Bitcasts and extract the root pointer base allocation
+    auto DL = M.getDataLayout();
+    Value *LHSBase = LHS->stripAndAccumulateConstantOffsets(
+        DL, LHSOffset, /*AllowNonInbounds=*/true);
+    Value *RHSBase = RHS->stripAndAccumulateConstantOffsets(
+        DL, RHSOffset, /*AllowNonInbounds=*/true);
+
+    // Check if both pointers safely root back to the exact same alloca
+    // instruction
+    if (LHSBase && RHSBase && isa<AllocaInst>(LHSBase) && LHSBase == RHSBase) {
+      bool Result = false;
+
+      // Evaluate the ICmp predicate natively at compile time
+      switch (Pred) {
+      case ICmpInst::ICMP_EQ:
+        Result = (LHSOffset == RHSOffset);
+        break;
+      case ICmpInst::ICMP_NE:
+        Result = (LHSOffset != RHSOffset);
+        break;
+      case ICmpInst::ICMP_SLT:
+        Result = LHSOffset.slt(RHSOffset);
+        break;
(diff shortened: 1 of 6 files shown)


lib/ReplaceLLVMIntrinsicsPass.cpp
```diff
@@ -542,6 +545,70 @@ static void replaceMemsetDynamic(CallInst *CI, Module &M, Value *Dst,
...
+      }
+
+      for (auto MMI : CallsToReplace) {
+        if (!isa<ConstantInt>(MMI->getLength())) {
+          // Use LLVM utility to expand memmove intrinsic as a loop of byte
+          // sized copies, requiring the Int8 capability. This expansion also
+          // requires SPIRV 1.4 and later due to use of pointer comparison
+          // semantics in implementation to decide whether to copy front to
+          // back, or back to front.
+          expandMemMoveAsLoop(MMI, TargetTransformInfo(M.getDataLayout()));
```
rjodinchr: This function returns a boolean, I think we should check it

lib/ReplaceLLVMIntrinsicsPass.cpp
```diff
@@ -542,6 +545,70 @@ static void replaceMemsetDynamic(CallInst *CI, Module &M, Value *Dst,
...
+        if (auto MMI = dyn_cast<MemMoveInst>(U)) {
+          CallsToReplace.push_back(MMI);
+        }
+      }
+
+      for (auto MMI : CallsToReplace) {
+        if (!isa<ConstantInt>(MMI->getLength())) {
+          // Use LLVM utility to expand memmove intrinsic as a loop of byte
+          // sized copies, requiring the Int8 capability. This expansion also
+          // requires SPIRV 1.4 and later due to use of pointer comparison
```
rjodinchr: Let's check for spv1.4 then:
```
clspv::Option::SpvVersion() < clspv::Option::SPIRVVersion::SPIRV_1_4
```

lib/ReplaceLLVMIntrinsicsPass.cpp
```diff
@@ -542,6 +545,70 @@ static void replaceMemsetDynamic(CallInst *CI, Module &M, Value *Dst,
...
+          // semantics in implementation to decide whether to copy front to
+          // back, or back to front.
+          expandMemMoveAsLoop(MMI, TargetTransformInfo(M.getDataLayout()));
+          MMI->eraseFromParent();
+        } else {
+          // Creates an intermediate alloca byte array of a known size, then
+          // memcpy the src operand ptr into the alloca, followed by a memcpy of
+          // the alloca to the dst operand ptr. Requires Int8 capability.
```
rjodinchr: Check for `int8` support `clspv::Option::Int8Support()`

lib/ReplaceLLVMIntrinsicsPass.cpp
```diff
@@ -542,6 +545,70 @@ static void replaceMemsetDynamic(CallInst *CI, Module &M, Value *Dst,
   CI->eraseFromParent();
 }
 
+static void replaceMemmoveStatic(Module &M, MemMoveInst *MMI) {
+  auto *Length = cast<ConstantInt>(MMI->getLength());
+  auto NumBytes = Length->getZExtValue();
+  auto *AllocaTy =
+      ArrayType::get(IntegerType::getInt8Ty(M.getContext()), NumBytes);
```
rjodinchr, line 679: That alloca could be huge, are we sure we want it whatever the size?

rjodinchr (changes requested):

EwanC:
Thanks for rerunning the CI @rjodinchr I can see the clang-format fail log but not the `amber-ubuntu-gcc` build log, could paste the relevant compiler error so I can address it?

lib/ReplaceLLVMIntrinsicsPass.cpp
EwanC: That's a great point, I've added a command-line option that controls the limit and if that's exceeded then we fallback to the dynamic implemented. I've defaulted that to 128 but that was an arbitrary choice and happy to change that (or this solution if you don't think a command-line option is appropriate)

rjodinchr:
It seems to be an infra error, but somehow I'm not able to re-trigger it at the moment.
We will see when the CI will run with the fixed formatting.

test/LLVMIntrinsics/memmove.ll
```diff
@@ -0,0 +1,55 @@
+; RUN: clspv --physical-storage-buffers --arch=spir64 -x=ir %s -o %t.spv
```
rjodinchr: I think we should also test without `--physical-storage-buffers`

lib/ReplaceLLVMIntrinsicsPass.cpp
rjodinchr: Looking good, thank you.
For the default value, I guess whatever works for you right now would be the best start.

rjodinchr (approved):

PR #1616, opened 2026-07-01
Add options to control denormal modes

rjodinchr:
Introduce `-denorm-preserve` and `-denorm-flush-to-zero` options to clspv.
These options allow specifying which floating-point types (fp16, fp32, fp64)
should preserve denormals or flush them to zero.

Refactor existing options to use a unified `FloatingPointType` enum instead of
separate enums for FMA and RoundingModeRTE.

Introduce `DenormMode` enum to represent the denormal handling state
(preserve, flush_to_zero, unspecified, error).

The new options are used to:
- Set Clang's `FP32DenormalMode` and `FPDenormalMode` accordingly.
- Emit `DenormPreserve` or `DenormFlushToZero` execution modes and
  capabilities in the SPIR-V producer.

Validation is added in `ParseOptions` to ensure:
- Preserve and FlushToZero are mutually exclusive for a given type.
- fp16 and fp64 share the same mode, as they are mapped to the single
  `FPDenormalMode` in Clang CodeGenOptions.

The unused `cl-denorms-are-zero` option is marked as having no effect.

### DIFF
--- a/include/clspv/Option.h
+++ b/include/clspv/Option.h
@@ -189,7 +189,7 @@ inline bool LanguageUsesGenericAddressSpace() {
               clspv::FeatureMacro::__opencl_c_generic_address_space) > 0);
 }
 
-enum class RoundingModeRTE : uint32_t {
+enum class FloatingPointType : uint32_t {
   fp16,
   fp32,
   fp64,
@@ -197,7 +197,17 @@ enum class RoundingModeRTE : uint32_t {
 
 // Returns true when the execution mode RoundingModeRTE should be set for a
 // floating point type.
-bool ExecutionModeRoundingModeRTE(RoundingModeRTE rm);
+bool ExecutionModeRoundingModeRTE(FloatingPointType rm);
+
+enum class DenormMode : uint32_t {
+  error,
+  preserve,
+  flush_to_zero,
+  unspecified,
+};
+
+// Returns the execution mode for a floating point type.
+DenormMode ExecutionModeDenorm(FloatingPointType fpty);
 
 // Return the SPIR-V binary version
 enum class SPIRVVersion : uint32_t {
@@ -319,11 +329,6 @@ bool UntypedPointers();
 // Returns true if untyped pointers are supported in aspace.
 bool UntypedPointerAddressSpace(unsigned aspace);
 
-enum class SpvKhrFma : uint32_t {
-  fp16,
-  fp32,
-  fp64,
-};
 // Returns true if OpFmaKHR is supported for the type size.
 bool SupportsFmaKHR(uint32_t scalarSizeInBits);
 
--- a/lib/Compiler.cpp
+++ b/lib/Compiler.cpp
@@ -93,6 +93,8 @@ static llvm::cl::opt<bool> cl_single_precision_constants(
     llvm::cl::desc("Treat double precision floating-point constant as single "
                    "precision constant."));
 
+// This option has not effect. Mainly because of lack of expresiveness. Instead
+// use `-denorm-preserve` and `-denorm-flush-to-zero`.
 static llvm::cl::opt<bool> cl_denorms_are_zero(
     "cl-denorms-are-zero", llvm::cl::init(false),
     llvm::cl::desc("If specified, denormalized floating point numbers may be "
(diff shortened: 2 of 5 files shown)


lib/Compiler.cpp
```diff
@@ -93,10 +93,6 @@ static llvm::cl::opt<bool> cl_single_precision_constants(
     llvm::cl::desc("Treat double precision floating-point constant as single "
                    "precision constant."));
 
-static llvm::cl::opt<bool> cl_denorms_are_zero(
-    "cl-denorms-are-zero", llvm::cl::init(false),
```
dneto0, line 97: This is part of OpenCL 3.0.

Instead of removing it, make it do nothing. (Or make it emit a warning that it does nothing)

https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#math-intrinsics-options

include/clspv/Option.h
```diff
@@ -199,6 +199,20 @@ enum class RoundingModeRTE : uint32_t {
 // floating point type.
 bool ExecutionModeRoundingModeRTE(RoundingModeRTE rm);
 
+enum class DenormMode : uint32_t {
```
dneto0, line 202: The name is confusing: I was expecting modes like "preserve" or "flushToZero".

I'd prefer to name this "DenormPreserve"

lib/Compiler.cpp
```diff
@@ -317,8 +313,35 @@ int SetCompilerInstanceOptions(
     instance.getLangOpts().NativeHalfType = true;
     instance.getLangOpts().NativeHalfArgsAndReturns = true;
   }
-  instance.getCodeGenOpts().FPDenormalMode = llvm::DenormalMode::getDynamic();
-  instance.getCodeGenOpts().FP32DenormalMode = llvm::DenormalMode::getDynamic();
+  // FP32 denormal mode
+  if (clspv::Option::ExecutionModeDenormPreserve(
+          clspv::Option::DenormMode::fp32)) {
+    instance.getCodeGenOpts().FP32DenormalMode = llvm::DenormalMode::getIEEE();
+  } else if (clspv::Option::ExecutionModeDenormFlushToZero(
```
dneto0: This if-else structure means that if denorm-preserve is set then the denorm-flush-to-zero option is ignored.  That's confusing.  I would structure it so that at most one of them can be used, so it's three states:   default, preserve, flush.

dneto0 (changes requested):

lib/Compiler.cpp
rjodinchr: Keeping it doesn't feel right because it lacks expressiveness. Simply emitting a warning is highly error-prone for those using clspv for their runtime. Additionally, the warning would expose a clspv-specific flag, potentially to users who shouldn't have to deal with it.

While I understand clspv is intended to compile OpenCL-C, I would argue it doesn't need to expose every OpenCL compile flag directly. It should, however, provide those writing runtimes the ability to implement every OpenCL flag.

I understand your point too. Please let me know if you truly prefer to keep it, while including a warning to indicate that it serves no functional purpose.

include/clspv/Option.h
rjodinchr: I agree. I was trying to avoid duplicating it for both `Preserve` and `FlushToZero`. Maybe it would be a good call to just merge it with `RoundingModeRTE` and name it based on what it really is... `FloatingPointType`, for example?

lib/Compiler.cpp
rjodinchr: I agree, I'll change that.

lib/Compiler.cpp
dneto0: I think the best option is to keep it, update its description to say it does nothing, why it's being kept, and avoid emitting a warning.

include/clspv/Option.h
dneto0: oh, I didn't see the pattern of RoundingModeRTE just above it (which has the example of 3 fp bit widths), and I did not think much about the use for both the preserve and flush-to-zero cases.

Merging into a single `FloatingPointType` makes sense.  If we can use a name that indicates that the enum is used for selecting modes or behaviours then that would be great, if it's not too long.  I can't think of a good short name other then FloatingPointType.

dneto0 (approved):
Nice.
thank you!

PR #1622, opened 2026-07-25
Prefer OptionalPassInfoMixin

boomanaiden154:
PassInfoMixin in upstream LLVM will soon be moved to a detail namespace with users expected to use RequiredPassInfoMixin or OptionalPassInfoMixin depending upon the behavior they want. Given none of these passes explicitly set isRequired to true, move them all to OptionalPassInfoMixin.

### DIFF
--- a/lib/AddFunctionAttributesPass.h
+++ b/lib/AddFunctionAttributesPass.h
@@ -20,7 +20,7 @@
 
 namespace clspv {
 struct AddFunctionAttributesPass
-    : llvm::OptionalOptionalPassInfoMixin<AddFunctionAttributesPass> {
+    : llvm::OptionalPassInfoMixin<AddFunctionAttributesPass> {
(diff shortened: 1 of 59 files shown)


lib/AddFunctionAttributesPass.h
```diff
@@ -20,7 +20,7 @@
 
 namespace clspv {
 struct AddFunctionAttributesPass
-    : llvm::PassInfoMixin<AddFunctionAttributesPass> {
+    : llvm::OptionalOptionalPassInfoMixin<AddFunctionAttributesPass> {
```
dneto0: this is a typo.

dneto0 (changes requested):

lib/AddFunctionAttributesPass.h
dneto0: I've fixed it in the commit.

dneto0 (approved):

PR #1623, opened 2026-08-10
Support physical storage buffers with non-clustered ubo pod args

EwanC:
This patch allows a user to compile a kernel with the following combination of arguments, which forces every POD argument into its own uniform buffer.

```sh
$ clspv -pod-ubo -cluster-pod-kernel-args=0 -physical-storage-buffers --arch=spir64
```

I found this arrangement useful when running on the Qualcomm Adreno Vulkan driver on Android. This specific driver does not like handling the ulong physical buffer argument pointers in clustered pod/ubo structs, despite the SPIRV passing spirv-val and being run correctly by the several other Vulkan drivers I've tried. Using individual single element ubo arguments works around this driver issue, so I can enable these flags when that specific driver is detected.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -1301,6 +1301,7 @@ void SPIRVProducerPassImpl::FindTypesForResourceVars() {
       break;
     case clspv::ArgKind::Pod:
     case clspv::ArgKind::PodUBO:
+    case clspv::ArgKind::PointerUBO:
     case clspv::ArgKind::PodPushConstant:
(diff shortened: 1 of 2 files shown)


dneto0:
>  This specific driver does like handling the ulong physical buffer argument pointers in clustered pod/ubo structs,

I assume you meant  "does _not_ like handling" ?

EwanC:
> > This specific driver does like handling the ulong physical buffer argument pointers in clustered pod/ubo structs,
> 
> I assume you meant "does _not_ like handling" ?

indeed :facepalm: Fixed that in the description

alan-baker (approved):

PR #1627, opened 2026-08-19
fix replaceMul when apply on libclc functions

rjodinchr:


alan-baker (approved):

rjodinchr:
`CI-ubuntu-clang-debug` has passed some time ago, it has just not been reported somehow.

PR #1631, opened 2026-08-21
Canonicalize operands of FMin/FMax/NMin/NMax when FTZ is enabled

rjodinchr:
In OpenCL, math builtins such as fmax, fmin, maxmag, and minmag require correct handling of subnormal values and signed zeros. When Flush-To-Zero (FTZ) mode is active (default for single/double precision unless preserved, and optional for half precision), subnormals must be treated as signed zeros (+0.0 / -0.0) before min/max comparisons and magnitude equality checks.

This change canonicalizes floating-point inputs to:
- llvm.minimumnum and llvm.maximumnum intrinsics (mapped to FMin/FMax)
- GLSL extended instructions FMin, FMax, NMin, and NMax

when denormals should be flushed to zero according to the configured DenormMode.

### DIFF
--- a/lib/SPIRVProducerPass.cpp
+++ b/lib/SPIRVProducerPass.cpp
@@ -4927,9 +4927,8 @@ SPIRVID SPIRVProducerPassImpl::GenerateCanonicalize(Value *val) {
   };
   Type *IntegerTy;
   if (auto VecTy = dyn_cast<FixedVectorType>(InputTy)) {
-    IntegerTy =
-        FixedVectorType::get(getSameSizeIntegerTy(VecTy->getElementType()),
-                             VecTy->getNumElements());
+    IntegerTy = FixedVectorType::get(
+        getSameSizeIntegerTy(VecTy->getElementType()), VecTy->getNumElements());
   } else {
     IntegerTy = getSameSizeIntegerTy(InputTy);
   }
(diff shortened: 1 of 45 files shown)


lib/SPIRVProducerPass.cpp
```diff
@@ -4900,6 +4902,83 @@ SPIRVID SPIRVProducerPassImpl::GenerateFabs(Value *Input) {
   return addSPIRVInst(spv::OpBitcast, Ops);
 }
 
+bool SPIRVProducerPassImpl::ShouldFlushDenorms(Type *Ty) const {
+  Type *ScalarTy = Ty->getScalarType();
+  if (ScalarTy->isHalfTy()) {
+    return clspv::Option::ExecutionModeDenorm(
+               clspv::Option::FloatingPointType::fp16) ==
+           clspv::Option::DenormMode::flush_to_zero;
```
alan-baker, line 4910: Why does f16 differ from the other widths?

rjodinchr: This is because of fp16 libclc implementations which assume standard IEEE half arithmetic unless explicitly running in DAZ/FTZ mode

rjodinchr: The issue lies with drivers that are unable to either force flushing-to-zero or preserve-denorm. In those cases, we need to try to stay compliant; for that scenario, I am getting the correct result with libclc by flushing fp32 while preserving fp16.

alan-baker (approved):

Issue #1596, opened 2026-04-28
Request for Feedback: OpenCL C cooperative matrix in CLSPV


lpavank:
We are starting to look at implementing support for OpenCL C Cooperative matrix extension in CLSPV.
And, wanted to gather feedback from CLSPV maintainers (@rjodinchr , @alan-baker) about our flow/ideas here.

Here are following assumptions/what we have:
1. Reference implementation for clang/llvm (contributed by Imagination) is what we used which is here
https://gitlab.khronos.org/opencl/llvm-project/-/merge_requests/5/diffs 
2. RFC from OpenCL community (mainly driven by QCOM) is here https://discourse.llvm.org/t/rfc-clang-frontend-changes-for-opencl-c-cooperative-matrix-extension/90148
3. SPIRV-COOPERATIVE-MATRIX https://github.khronos.org/SPIRV-Registry/extensions/KHR/[SPV_KHR_cooperative_matrix](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_cooperative_matrix.html).html is already spec-ed out and defines all operations that are needed.

With the above, here is what we were thinking.
1. ReplaceBuiltPass in CLSPV need to be adopted as needed.
2. SPIRVProducerPass in CLSPV will generate the SPIRV based on SPIRV spec above with mapping/information as needed.

PS: We do have first hello world running on a device with layered implementation (on a device where VK/SPIRV exists for the extension) and are willing to share the lower-level details/code as needed here, but want to share the above first to see if there is anything more, we should consider here.
We do understand there are things in FLUX here on assumptions, but that's what we have to start with.

CC @gowtham-sarc , shyam Manohar (contributors)

alan-baker:
This looks relatively straightforward. The target EXT type should make this reasonable track. I expect most of the work should just be in the producer.

lpavank:
Yupp thats what we had "most of work in producer", other than some mangling fix up in "replacebuiltins" (given the clang implementation we have).

rjodinchr:
LGTM

lpavank:
Thanks @alan-baker and @rjodinchr. We will come back with implementation once all moving parts from WG are settled and/or when ready. Closing the issue, since we are aligned on feedback.

PR #965, opened 2022-11-28
Add vulkanMemoryModel

omarahmed1111:
# PR Changes
* Add `vulkan-memory-model` option to clspv
* Generate VulkanMemoryModel shaders from clspv
* Change memorySemantics for atomics on VulkanMemoryModel
* Remove Coherent decoration on VulkanMemoryModel as it is not supported and replace it with relevant `Available` and `Visible` on read and write instructions

### [Related issue](https://github.com/google/clspv/issues/268)

**This contribution is being made by Codeplay on behalf of Samsung.**

PR #1645, opened 2026-09-04
Add support for cl_khr_cooperative_matrix

shyammanohar:
Depends on the Clang frontend changes in llvm/llvm-project#213986 (WIP). 
NOTE: CI tests will fail until the Clang patch is merged and the LLVM submodule is updated.
           The logic has been verified locally with a patched Clang build and spirv-val.

This patch lowers the OpenCL C cooperative matrix builtins to a spirv.CooperativeMatrixKHR target extension type and a set of __spirv_CooperativeMatrix* calls, consuming the IR generated by the Clang frontend.

Implementation Details:
- Add helpers to identify the cooperative matrix type and read its element type, scope, rows, columns and use.
- Map the builtin names, stripping the type suffix clang appends before the existing lookup.
- Generate OpTypeCooperativeMatrixKHR and the load, store, mulAdd, binary (add/sub/mul/div), scalar (mul/neg), length and init instructions.
- Infer the pointee type for cooperative matrix load/store pointer operands.
- Select memory operands by storage class.
- Emit the accumulator component type as a signed integer so the matrix type matches VkCooperativeMatrixPropertiesKHR, which is validated against the type itself before any mulAdd operands are available.
- Add the new opcodes to the SPIR-V binary writer.
- Both logical and physical addressing are supported.

Testing:
Tests are added under test/CooperativeMatrix covering fp16 and int8/uint8. Each test checks the generated matrix types and the signedness operands on OpCooperativeMatrixMulAddKHR, which differ across the three variants.

Verification:
Verified on Linux with an AMD RDNA GPU using 16^3 kernels in basic, tiled, and shmem OpenCL kernels. Testing included variants of the reference shaders from: https://github.com/jeffbolznv/vk_cooperative_matrix_perf/tree/master/shaders

Part of https://github.com/google/clspv/issues/1596


lpavank:
tagging @rjodinchr and @alan-baker for review/inputs

rjodinchr:
Making this a draft as long as the LLVM change are not in clspv

lib/SPIRVProducerPass.cpp
```diff
@@ -3571,6 +3592,17 @@ void SPIRVProducerPassImpl::GenerateModuleInfo() {
     addSPIRVInst<kExtensions>(spv::OpExtension, "SPV_KHR_integer_dot_product");
   }
 
+  // Add extension for VulkanMemoryModel
+  if (CapabilitySet.count(spv::CapabilityVulkanMemoryModel)) {
```
alan-baker, line 3596: I forgot about this dependence. I think more work might be needed to do this properly. A couple examples: device atomics should be moved to queue family scope and coherent decorations need to be changed to use av/vis instead.

alan-baker: Most memory accesses need NonPrivatePointer set and image load/store need NonPrivateTexel (though maybe only if the image fence is used).

alan-baker (changes requested):
The coop matrix specific seems ok, but I don't think we can just enable the vulkan memory model. There was an unfinished PR (#965) which maybe could be revived.

lib/SPIRVProducerPass.cpp
```diff
@@ -7450,7 +7489,9 @@ void SPIRVProducerPassImpl::WriteSPIRVBinary(
     case spv::OpGroupNonUniformFMax:
     case spv::OpGroupNonUniformShuffle:
     case spv::OpGroupNonUniformShuffleXor:
-    case spv::OpGroupNonUniformRotateKHR: {
+    case spv::OpGroupNonUniformRotateKHR:
+    case spv::OpCooperativeMatrixLoadKHR:
+    case spv::OpCooperativeMatrixMulAddKHR: {
```
rjodinchr, line 7494: Missing `spv::OpCooperativeMatrixLengthKHR`?

test/CooperativeMatrix/coop_mat_int8_signed.cl
```diff
@@ -0,0 +1,45 @@
...
+// RUN: FileCheck %s < %t2.spvasm
+
+// CHECK-DAG: OpCapability CooperativeMatrixKHR{{$}}
+// CHECK-DAG: OpExtension "SPV_KHR_cooperative_matrix"
+// CHECK-DAG: [[i8:%[^ ]+]] = OpTypeInt 8 0
+// CHECK-DAG: [[i32:%[^ ]+]] = OpTypeInt 32 1
+// CHECK-DAG: OpTypeCooperativeMatrixKHR [[i8]]
+// CHECK-DAG: [[acc_ty:%[^ ]+]] = OpTypeCooperativeMatrixKHR [[i32]]
+// CHECK: OpCooperativeMatrixMulAddKHR [[acc_ty]] {{%[^ ]+}} {{%[^ ]+}} {{%[^ ]+}}
+// MatrixASignedComponentsKHR|MatrixBSignedComponentsKHR|MatrixCSignedComponentsKHR|MatrixResultSignedComponentsKHR
```
rjodinchr, line 13: It feels like something is missing here. `CHECK_SAME` maybe? I'm not sure

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  Value *MatC = Call->getArgOperand(2);
+  SPIRVOperandVec Ops;
+  Ops << ResultType << MatA << MatB << MatC;
+
+  // Helper to check if a Value's type is an integer cooperative matrix
+  auto isIntMat = [&](Value *V) {
+    auto *T = dyn_cast<TargetExtType>(V->getType());
+    return T && T->getTypeParameter(0)->isIntegerTy();
+  };
+  auto *SignedFlag = dyn_cast<ConstantInt>(Call->getArgOperand(3));
```
rjodinchr, line 8771: Are we sure that the operand is always there? If it is not, we need to guard it somehow.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  const bool DataSigned = SignedFlag && SignedFlag->isOne();
```
rjodinchr, line 8772: My understanding is that operand 3 is a bitmask of `CooperativeMatrixOperands` (`MatrixASignedComponentsKHR = 0x1`, `MatrixBSignedComponentsKHR = 0x2`), allowing independant signedness for `A` and `B`. Testing `isOne` couples `A` and `B` together and ignores explicit masks.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+      operands |= spv::CooperativeMatrixOperandsMatrixBSignedComponentsKHRMask;
+  }
+  if (isIntMat(MatC))
+    operands |= spv::CooperativeMatrixOperandsMatrixCSignedComponentsKHRMask;
+
+  // Use RetTy directly since we already know it's a TargetExtType
+  if (RetTy->getTypeParameter(0)->isIntegerTy()) {
+    operands |=
+        spv::CooperativeMatrixOperandsMatrixResultSignedComponentsKHRMask;
+  }
```
rjodinchr, line 8788: I don't think `isIntegerTy` means that this is a signed matrix, right?

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  Ops << GetCoopMatPointerOperand(Ptr) // Pointer
+      << Obj                           // Object (the matrix)
+      << Call->getArgOperand(2)        // Memory Layout <id>
+      << Call->getArgOperand(3);       // Stride <id>, in ELEMENTS
+
+  AppendCoopMatMemoryOperand(Ops, Ptr, Obj->getType(), /*IsStore=*/true);
+  return addSPIRVInst(spv::OpCooperativeMatrixStoreKHR, Ops);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatMulAdd(CallInst *Call) {
```
rjodinchr, line 8752: It feels like saturation is ignored here.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatUnaryNeg(CallInst *Call) {
+  // coop_mat_scalar_neg(matrix, is_signed) -> OpSNegate/OpFNegate
+  auto *RetTy = dyn_cast<TargetExtType>(Call->getType());
+  SPIRVID ResultType = getSPIRVType(RetTy);
+  Value *Matrix = Call->getArgOperand(0);
+  llvm::Type *ElemTy = RetTy->getTypeParameter(0);
+  bool IsFloat = ElemTy->isFloatingPointTy();
+  SPIRVOperandVec Ops;
+  Ops << ResultType << Matrix;
+  return addSPIRVInst(IsFloat ? spv::OpFNegate : spv::OpSNegate, Ops);
```
rjodinchr, line 8860: `OpSNegate` requires a operand type with `Signedness = 1`, which might now be the case here I believe.

lib/SPIRVProducerPass.cpp
rjodinchr: Also the extension is needed only for `SpvVersion < 1.3` I think

lib/Types.cpp
```diff
@@ -164,6 +162,28 @@ Type *InferUserType(User *user, bool &isPointerTy, unsigned operand,
...
+          return ElemTy;
+        }
+        return nullptr;
+      }
+      break;
+    }
+    case clspv::Builtins::kCoopMatLength: {
+      // coop_mat_length(matrix) -> uint32_t
+      return call->getType();
+    }
```
rjodinchr, line 186: `InferUserType` infers the pointee type for a pointer argument used by an instruction. `coop_mat_length` takes a matrix value, not a pointer, and has no pointer operands. This case does not make sense and should be removed.

lib/BuiltinsMap.inc
```diff
@@ -1138,6 +1141,22 @@ static std::unordered_map<const char *, Builtins::BuiltinType, cstr_hash,
...
+        // ..._global_f16_sc3_16x16_u0),
+        // LookupBuiltinType() in Builtins.cpp truncates recover the base name.
+        {"__spirv_CooperativeMatrixLoadKHR", Builtins::kCoopMatLoad},
+        {"__spirv_CooperativeMatrixStoreKHR", Builtins::kCoopMatStore},
+        {"__spirv_CooperativeMatrixMulAddKHR", Builtins::kCoopMatMulAdd},
+        {"__spirv_CompositeConstruct", Builtins::kCoopMatInit},
+        {"__spirv_CooperativeMatrixFAdd", Builtins::kCoopMatBinaryAdd},
+        {"__spirv_CooperativeMatrixFSub", Builtins::kCoopMatBinarySub},
+        {"__spirv_CooperativeMatrixFMul", Builtins::kCoopMatBinaryMul},
+        {"__spirv_CooperativeMatrixFDiv", Builtins::kCoopMatBinaryDiv},
```
rjodinchr, line 1155: No integer operation, is that normal?

lib/ReplaceOpenCLBuiltinPass.cpp
```diff
@@ -757,6 +757,21 @@ bool ReplaceOpenCLBuiltinPass::runOnFunction(Function &F) {
...
+  case Builtins::kCoopMatBinaryAdd:
+  case Builtins::kCoopMatBinarySub:
+  case Builtins::kCoopMatBinaryMul:
+  case Builtins::kCoopMatBinaryDiv:
+  case Builtins::kCoopMatScalarMul:
+  case Builtins::kCoopMatScalarNeg:
+  case Builtins::kCoopMatInit:
+  case Builtins::kCoopMatLength:
+    return false;
+
```
rjodinchr, line 774: Default is already to return `false`, I don't think we want to keep that.

lib/BuiltinsMap.inc
```diff
@@ -1153,7 +1172,8 @@ static std::unordered_map<const char *, Builtins::BuiltinType, cstr_hash,
         {"clspv.sampler_var_literal", Builtins::kClspvSamplerVarLiteral},
         {"clspv.composite_construct", Builtins::kClspvCompositeConstruct},
 
-        {"clspv.get_normalized_sampler_mask", Builtins::kClspvGetNormalizedSamplerMask},
+        {"clspv.get_normalized_sampler_mask",
```
rjodinchr, line 1175: No reason to format a line that is not getting changed.
Also I would argue that all the other line added above should not be formatted in 2 lines, 1 is fine here.

lib/SPIRVProducerPass.cpp
```diff
@@ -699,6 +712,9 @@ struct SPIRVProducerPassImpl {
   TypeMapType ImageTypeMap;
   // A unique-vector of LLVM types that map to a SPIR-V type.
   TypeList Types;
+  // Maps for caching cooperative matrix types.
+  std::map<llvm::TargetExtType *, SPIRVID> CoopMatTypeMap;
```
rjodinchr, line 716: Those are the same types as the ones cached by `TypeMap`, I don't see why we want to use a different map. Also, the whole file is using `DenseMap`.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+               spv::MemoryAccessNonPrivatePointerMask |
+               (IsStore ? spv::MemoryAccessMakePointerAvailableMask
+                        : spv::MemoryAccessMakePointerVisibleMask))
+        << getSPIRVInt32Constant(spv::ScopeWorkgroup);
+    return;
+  }
+
+  if (sc == spv::StorageClassStorageBuffer ||
+      sc == spv::StorageClassPhysicalStorageBuffer ||
+      sc == spv::StorageClassCrossWorkgroup) {
```
rjodinchr, line 8708: `GetStorageClass` never returns `spv::StorageClassCrossWorkGroup`.

lib/Types.cpp
```diff
@@ -704,6 +721,29 @@ bool clspv::IsWriteOnlyImageType(llvm::Type *type) {
   return false;
 }
 
+bool clspv::IsCooperativeMatrixType(llvm::Type *type) {
+  if (!type)
+    return false;
+
+  // Check for spirv.CooperativeMatrixKHR TargetExtType.
+  if (auto *ext_ty = dyn_cast<TargetExtType>(type)) {
```
rjodinchr, line 729: `dyn_cast` already handles the `!type` case, so we can remove the first check.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  Ops << Call->getArgOperand(0);
+  return addSPIRVInst(spv::OpCompositeConstruct, Ops);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatLength(CallInst *Call) {
+  // coop_mat_length(matrix) -> OpCooperativeMatrixLengthKHR, Returns Rows *
+  // Cols
+  SPIRVID ResultType = getSPIRVType(Call->getType());
+  Value *Matrix = Call->getArgOperand(0);
+  auto *RetTy = dyn_cast<TargetExtType>(Matrix->getType());
```
rjodinchr, line 8885: This is not the return type, the name is confusing.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+    return addSPIRVInst<kConstants>(spv::OpConstantComposite, Ops);
+  }
+
+  Ops << Call->getArgOperand(0);
+  return addSPIRVInst(spv::OpCompositeConstruct, Ops);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatLength(CallInst *Call) {
+  // coop_mat_length(matrix) -> OpCooperativeMatrixLengthKHR, Returns Rows *
+  // Cols
```
rjodinchr, line 8882: I don't think that's really the case. I think it returns the implementation-dependent number of components accessible to each invocation in the subgroup.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  SPIRVOperandVec ConstructOps;
+  ConstructOps << ResultType << Scalar;
+  SPIRVID ScalarMatrix = addSPIRVInst(spv::OpCompositeConstruct, ConstructOps);
+  SPIRVOperandVec MulOps;
+  MulOps << ResultType << Matrix << ScalarMatrix;
+  return addSPIRVInst(IsFloat ? spv::OpFMul : spv::OpIMul, MulOps);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatUnaryNeg(CallInst *Call) {
+  // coop_mat_scalar_neg(matrix, is_signed) -> OpSNegate/OpFNegate
```
rjodinchr, line 8852: `coop_mat_scalar_neg`? This is a matrix negation, not a scalar, right?

lib/Builtins.cpp
```diff
@@ -45,6 +45,20 @@ Builtins::LookupBuiltinType(const std::string &builtin_name) {
   if (ii != s_func_map.end()) {
     return (*ii).second;
   }
+
+  // Cooperative matrix builtins carry a typed suffix appended by the Clang
+  // frontend, e.g.:__spirv_CooperativeMatrixLoadKHR_global_f16_sc3_16x16_u0
+  if (name.rfind("__spirv_", 0) == 0) {
+    size_t pos = name.find('_', strlen("__spirv_"));
```
rjodinchr, line 52: It feels like we could do something cleaner here.

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+    llvm_unreachable("Invalid binary operation type");
+  }
+  SPIRVOperandVec Ops;
+  Ops << ResultType << MatA << MatB;
+  return addSPIRVInst(Opcode, Ops);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatScalarMul(CallInst *Call) {
+  // coop_mat_scalar_mul(matrix, scalar, is_signed) -> OpIMul/OpFMul
+  auto *RetTy = dyn_cast<TargetExtType>(Call->getType());
```
rjodinchr, line 8837: We should check `dyn_cast` does not return null I think

lib/SPIRVProducerPass.cpp
```diff
@@ -8543,3 +8584,312 @@ void SPIRVProducerPassImpl::GeneratePrintfReflection() {
...
+  ConstructOps << ResultType << Scalar;
+  SPIRVID ScalarMatrix = addSPIRVInst(spv::OpCompositeConstruct, ConstructOps);
+  SPIRVOperandVec MulOps;
+  MulOps << ResultType << Matrix << ScalarMatrix;
+  return addSPIRVInst(IsFloat ? spv::OpFMul : spv::OpIMul, MulOps);
+}
+
+SPIRVID SPIRVProducerPassImpl::GenerateCoopMatUnaryNeg(CallInst *Call) {
+  // coop_mat_scalar_neg(matrix, is_signed) -> OpSNegate/OpFNegate
+  auto *RetTy = dyn_cast<TargetExtType>(Call->getType());
```
rjodinchr, line 8853: We should check `dyn_cast` does not return null I think

lib/Types.cpp
```diff
@@ -704,6 +721,29 @@ bool clspv::IsWriteOnlyImageType(llvm::Type *type) {
...
+    if (ext_ty->getName() == "spirv.CooperativeMatrixKHR")
+      return true;
+  }
+  return false;
+}
+
+llvm::Type *clspv::GetCoopMatrixElementType(llvm::Type *type) {
+  // Return the first type parameter (element type) of a cooperative matrix.
+  if (auto *targetExtTy = llvm::dyn_cast<llvm::TargetExtType>(type)) {
+    if (targetExtTy->getName() == "spirv.CooperativeMatrixKHR" &&
```
rjodinchr, line 739: Should we use `IsCooperativeMatrixType` here to avoid the string duplication.

test/CooperativeMatrix/coop_mat_fp16.cl
```diff
@@ -0,0 +1,50 @@
...
+// CHECK-DAG: OpExtension "SPV_KHR_cooperative_matrix"
+// CHECK-DAG: [[uint:%[^ ]+]] = OpTypeInt 32 0
+// CHECK-DAG: [[half:%[^ ]+]] = OpTypeFloat 16
+// CHECK-DAG: [[float:%[^ ]+]] = OpTypeFloat 32
+// CHECK-DAG: [[subgroup:%[^ ]+]] = OpConstant [[uint]] 3
+// CHECK-DAG: [[dim:%[^ ]+]] = OpConstant [[uint]] 16
+// CHECK-DAG: [[mat_a:%[^ ]+]] = OpTypeCooperativeMatrixKHR [[half]] [[subgroup]] [[dim]] [[dim]] {{%[^ ]+}}
+// CHECK-DAG: [[mat_c:%[^ ]+]] = OpTypeCooperativeMatrixKHR [[float]] [[subgroup]] [[dim]] [[dim]] {{%[^ ]+}}
+// CHECK: OpCooperativeMatrixLoadKHR [[mat_a]]
+// CHECK: OpCooperativeMatrixMulAddKHR [[mat_c]] {{%[^ ]+}} {{%[^ ]+}} {{%[^ ]+}}{{$|.*}}
```
rjodinchr, line 17: trailing whitespace

rjodinchr (changes requested):
Missing test coverage for:
- `coop_mat_length`
- `+`, `-`, `*` & `/`
- scalar multiplication & unary negation
- non-zero / dynamic  `coop_mat_init`
- `__local_` (workgroup) memory load/store

PR #1649, opened 2026-09-10
unwrap AddrSpaceCastOperator in atomic_flag builtins

rjodinchr:
In OpenCL 2.0 / generic address space mode, Clang emits an AddrSpaceCast to generic address space on pointer arguments passed to atomic_flag builtins. In ReplaceOpenCLBuiltinPass, replaceAtomicFlagTestAndSet and replaceAtomicFlagClear checked pointer->getType()->getAddressSpace() directly to determine whether the pointer is in global or local memory. Because the pointer had generic address space (addrspace(4)), is_global evaluated to false, erroneously falling back to WorkgroupMemory semantics (0x100) instead of UniformMemory semantics (0x40). On multi-SM GPUs, this broke memory visibility and synchronization across workgroups for device-scope atomic_flag operations.

Unwrap the pointer operand if it is an AddrSpaceCastOperator so that is_global correctly identifies global memory pointers and generates the appropriate UniformMemory semantics. Also add OpenCL 2.0 test coverage to atomic_flag.cl.


dneto0 (approved):
Nice. Must have been tricky.

PR #1650, opened 2026-09-10
Do not allocate descriptors for metadata-only image arguments

rjodinchr:
When an image argument is only used for metadata queries (get_image_channel_order and get_image_channel_data_type) and is never sampled, read, or written, AllocateDescriptorsPass previously still allocated a descriptor set and binding for it.

Because no read or write operations exist to infer the image type, clspv defaults the descriptor variable to a float sampled image (OpTypeImage %float). In runtimes such as clvk, binding a non-float image view (such as an integer image) to this unused descriptor triggers Vulkan validation errors (VUID-vkCmdDispatch-format-07753).

This patch:
- Checks in AllocateDescriptorsPass if an image kernel argument's only uses are image metadata queries (looking through bitcasts and addrspacecasts), and skips descriptor allocation for it if so.
- Updates SetImageMetadataPass::getOrdinal to accept an llvm::Argument directly (stripping casts), so that push constant reflection metadata for channel order and channel data type is still generated for unallocated image arguments.
- Adds a test verifying that no image variable or sampled image reflection is generated while the push constant reflection is preserved.


lib/AllocateDescriptorsPass.cpp
```diff
@@ -53,6 +54,44 @@ const uint32_t kMemorySemanticsUniformMemory = 0x40;
...
+bool IsImageMetadataQuery(const CallInst *Call) {
+  if (auto *Callee = Call->getCalledFunction()) {
+    auto builtin_type = clspv::Builtins::Lookup(Callee).getType();
+    return builtin_type == clspv::Builtins::kGetImageChannelOrder ||
+           builtin_type == clspv::Builtins::kGetImageChannelDataType;
+  }
+  return false;
+}
+
+bool HasOnlyImageMetadataQueries(const Argument &Arg) {
```
alan-baker, line 66: Should this traverse user calls? If the kernel argument is passed to helpers it, but only used for metadata this wouldn't catch it I think.

rjodinchr: We don't need to traverse user calls here because `InlineFuncWithImageMetadataGetterPass` runs earlier in the pipeline and already inlines any helpers calling `get_image_channel_order` or `get_image_channel_data_type` into the kernel (`SetImageMetadataPass` actually relies on this too, since it only inspects kernels).

If the helper does anything else with the image, DRA will need a descriptor anyway, so bailing out and allocating one is the right thing to do.

I've added a test case with a helper to `test/ImageBuiltins/get_image_channel_no_image_variable.cl` to make sure it's covered.

alan-baker (approved):

PR #1651, opened 2026-09-10
Check atomics when determining coherence

### LINKED ISSUE #1638: Coherent decoration and OpenCL 2.0 acquire/release atomics
Follow-up to [#126](https://github.com/google/clspv/issues/126), which was fixed by [#272](https://github.com/google/clspv/pull/272).

**1. Does the OpenCL 1.2 relaxed-atomics assumption still hold?**

[docs/OpenCLCOnVulkan.md](https://github.com/google/clspv/blob/main/docs/OpenCLCOnVulkan.md) describes the input language as OpenCL C 1.2, but the same document also lists support for the OpenCL 2.0 atomic functions, and notes that `memory_order_seq_cst` is weakened to acquire for loads and release for stores. That reads as though non-relaxed orders are expected to reach the generated SPIR-V. Given both statements, is it still the case that only relaxed atomics need to be considered?

**2. If not, should such atomics trigger the `Coherent` decoration?**

`AllocateDescriptorsPass` decides whether a buffer argument needs `Coherent` by asking whether the kernel's call tree contains a barrier or fence with `UniformMemory` or `ImageMemory` semantics. Atomic instructions are not examined, per [this review comment](https://github.com/google/clspv/pull/272#discussion_r263893045), that followed from OpenCL 1.2 atomics always being Relaxed. So a kernel that synchronizes only through acquire/release atomics gets no decoration:

```c
__kernel void mp(__global int *data, __global atomic_int *flag, __global int *out) {
    size_t lid = get_local_id(0);
    if (lid == 0) {
        data[0] = 42;
        atomic_store_explicit(flag, 1, memory_order_release, memory_scope_work_group);
    } else if (lid == 1) {
        if (atomic_load_explicit(flag, memory_order_acquire, memory_scope_work_group) == 1)
            out[0] = data[0];
    }
}
```

Both invocations are in the same work-group, so this doesn't rely on inter-work-group forward progress, and `data` is both written and read, but `uses_barriers` is false.

This matters when the module is consumed under the Vulkan memory model: `spirv-opt --upgrade-memory-model` from [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools) adds `NonPrivatePointer` + `MakePointerAvailable`/`MakePointerVisible` to `Workgroup` accesses, and to `StorageBuffer` accesses only when the base variable carries a `Coherent` or `Volatile` decoration. Without it, the payload `OpStore`/`OpLoad` on `data` remain private, so the `release`/`acquire` pair on `flag` still orders the two atomic accesses but does not make the write to `data` visible to the second invocation.

Is the intent that programs like this are out of scope, or should the decoration cover them?"

alan-baker:
Fixes: #1638

* Check atomic functions as synchronization primitives when determining whether to make a descriptor coherent


test/Coherent/atomic_store.ll
```diff
@@ -0,0 +1,52 @@
...
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64-G1"
+target triple = "spirv32-unknown-vulkan"
+
+@__spirv_WorkgroupSize = local_unnamed_addr addrspace(8) global <3 x i32> zeroinitializer
+
+define spir_kernel void @test(ptr addrspace(1) nofree align 4 %data, ptr addrspace(1) align 4 %a) local_unnamed_addr #0 !kernel_arg_addr_space !0 !kernel_arg_access_qual !1 !kernel_arg_type !2 !kernel_arg_base_type !3 !kernel_arg_type_qual !4 !kernel_arg_name !77 !clspv.pod_args_impl !79 {
+entry:
+  %0 = load i32, ptr addrspace(1) %data, align 4
+  tail call void @_Z8spirv.op.228.PU3AS1jjj(i32 228, ptr addrspace(1) %a, i32 1, i32 68, i32 11) #2
```
dneto0, line 15: 228 is OpAtomicStore.

Testing is a little thin: only one opcode is covered in this PR.   This is ok as a pipecleaner test, but isn't full coverage.

dneto0 (approved):

lib/AllocateDescriptorsPass.h
```diff
@@ -63,7 +63,7 @@ struct AllocateDescriptorsPass
   // The compiler targets OpenCL 1.2, which only provides support for relaxed
   // atomics which means they cannot be used as synchronization primitives.
   // That is why the pass does not consider them for the addition of coherence.
```
rjodinchr: We should update that comment I think

lib/AllocateDescriptorsPass.cpp
```diff
@@ -54,6 +54,8 @@ const uint32_t kMemorySemanticsUniformMemory = 0x40;
 // SPIR-V. Used to test barrier semantics.
 const uint32_t kMemorySemanticsImageMemory = 0x800;
 
+const uint32_t kMemorySemanticsNonRelaxed = 0x2 | 0x4 | 0x8 | 0x10;
```
rjodinchr:
```suggestion
const uint32_t kMemorySemanticsNonRelaxed = spv::MemorySemanticsAcquireMask |  spv::MemorySemanticsReleaseMask |
    spv::MemorySemanticsAcquireReleaseMask |
    spv::MemorySemanticsSequentiallyConsistentMask;
```

lib/AllocateDescriptorsPass.cpp
```diff
@@ -909,7 +916,10 @@ bool clspv::AllocateDescriptorsPass::CallTreeContainsGlobalBarrier(
   bool uses_barrier = false;
   for (auto &BB : *F) {
     for (auto &I : BB) {
-      if (auto *call = dyn_cast<CallInst>(&I)) {
+      if (auto *atomicrmw = dyn_cast<AtomicRMWInst>(&I)) {
+        uses_barrier =
+            isStrongerThan(atomicrmw->getOrdering(), AtomicOrdering::Monotonic);
```
rjodinchr, line 923: I think we should also check the addrspace of the pointer to avoid using a barrier on a __local pointer:
```
atomicrmw->getPointerAddressSpace() == clspv::AddressSpace::Global
```

test/Coherent/atomic_store.ll
```diff
@@ -0,0 +1,52 @@
+; RUN: clspv-opt %s -o %t.ll --passes=allocate-descriptors
```
rjodinchr, line 1: Maybe rename `atomic_store.ll`?

rjodinchr (approved):

